PR #55, opened 2023-12-05
show Nixpkgs issue in Admin view

fricklerhandwerk:


src/website/shared/migrations/0003_nixpkgsissue_id.py
```diff
@@ -0,0 +1,57 @@
+# Generated by Django 4.2.6 on 2023-12-05 01:13
+
+from django.db import migrations, models
+import django.db.models.deletion
+import django.utils.timezone
+
+
+class Migration(migrations.Migration):
+    replaces = [
+        ("shared", "0003_nixpkgsissue_rename_nixadvisory_nixpkgsadvisory_and_more"),
```
RaitoBezarius, line 10: this is weird, there's no such thing here, just drop them from the migration as you never committed those.

src/website/tracker/settings.py
```diff
@@ -170,3 +172,5 @@ def get_secret(name: str, encoding: str = "utf-8") -> str:
 # https://docs.djangoproject.com/en/4.2/ref/settings/#default-auto-field
 
 DEFAULT_AUTO_FIELD = "django.db.models.BigAutoField"
+
+INTERNAL_IPS = ["127.0.0.1"]
```
RaitoBezarius, line 176: add `[::1]`

src/website/shared/models.py
```diff
@@ -255,14 +257,29 @@ class Type(models.TextChoices):
...
 ##
 
 
-class NixIssue(models.Model):
+class NixpkgsIssue(models.Model):
     """The Nixpkgs version of a cve."""
 
+    @property
+    def issue_id(self):
+        return f"NIXPKGS-{str(self.created.year)}-{str(self.number).zfill(5)}"
```
RaitoBezarius, line 270: todo: the short code is a settings of the deployment.

src/website/shared/migrations/0003_nixpkgsissue_id.py
fricklerhandwerk: Done

src/website/shared/models.py
```diff
@@ -255,14 +257,29 @@ class Type(models.TextChoices):
...
+
+    created = models.DateField(auto_now_add=True)
+    number = models.IntegerField(default=0)
+
+    def save(self, *args, **kwargs):
+        if not self.pk:
+            max_number = NixpkgsIssue.objects.filter(
```
RaitoBezarius, line 277: either we make `number` an `models.AutoField()` and we lose the incremental ID **per year** but this is more reliable
either we have to make PostgresSQL use a subselect for the query during update because this thing is vulnerable to TOCTOU.

also please add a integrity constraint on unique_together = ("year", "number") so that the invariant is preserved

src/website/tracker/settings.py
fricklerhandwerk: Done

src/website/shared/models.py
RaitoBezarius: either case, add a TODO for the first item, add the integrity check please

src/website/webview/urls.py
```diff
@@ -1,8 +1,14 @@
...
+from webview.views import HomeView, NixpkgsIssueView, NixpkgsIssueListView
+from webview.converters import NixpkgsIssueConverter
 
 app_name = "webview"
+
+register_converter(NixpkgsIssueConverter, "nixpkgs")
+
 urlpatterns = [
     path("", HomeView.as_view(), name="home"),
+    path("issue/<nixpkgs:id>", NixpkgsIssueView.as_view(), name="issue_detail"),
```
RaitoBezarius, line 12: ideally, it should be `/<nixpkgs:id>` directly, no need for an additional subpath, leave it for the rest.
the central feature is here

fricklerhandwerk: No, this pollutes the namespace. Leave me that piece of the bikeshed.

src/website/shared/models.py
fricklerhandwerk: Broke out an issue

src/website/shared/models.py
fricklerhandwerk: Done

fricklerhandwerk:
Addressed comments and resolved conflicts.

src/website/webview/urls.py
```diff
@@ -1,8 +1,16 @@
...
 
-from webview.views import HomeView
+from webview.views import HomeView, NixpkgsIssueView, NixpkgsIssueListView
 
 app_name = "webview"
+
+
 urlpatterns = [
     path("", HomeView.as_view(), name="home"),
+    path("issues/", NixpkgsIssueListView.as_view(), name="issue_list"),
```
RaitoBezarius, line 10: Sure, but that's pretty much a standard https://security.archlinux.org/AVG-2596 FWIW.

PR #68, opened 2023-12-07
feat: Add triaged info on cve records and keep track of ingestions

Tom-Hubrecht:
Also split github API authentication into `shared.utils` and add a unit to rune the delta ingestion each day at 03h


nix/web-security-tracker.nix
```diff
@@ -113,42 +113,81 @@ in {
...
+          daphne ${networking} \
+            tracker.asgi:application
+        '';
       };
-      environment = {
-        DATABASE_URL = databaseUrl;
-        USER_SETTINGS_FILE = "${configFile}";
+
+      web-security-tracker-delta = {
+        description = "A web security tracker ASGI server";
```
RaitoBezarius, line 156:
```suggestion
        description = "Web security tracker catch up with CVEs";
```

nix/web-security-tracker.nix
```diff
@@ -113,42 +113,81 @@ in {
...
+          StateDirectory = "web-security-tracker";
+          RuntimeDirectory = "web-security-tracker";
+          LoadCredential = credentials;
+          Type = "oneshot";
+        };
+        environment = {
+          DATABASE_URL = databaseUrl;
+          USER_SETTINGS_FILE = "${configFile}";
+        };
+        preStart = ''
```
RaitoBezarius, line 174: Order it after the server script IMHO otherwise we may run into concurrent migrations…

RaitoBezarius: (or before the load, however you want)

src/website/shared/fetchers.py
```diff
@@ -177,20 +177,30 @@ def mkAffectedProduct(data: Dict[str, Any]) -> models.AffectedProduct:
...
+        uuid=data["assignerOrgId"], short_name=data.get("assignerShortName")
     )
-    ctx["requester"] = mkOrganization(uuid=data.get("requesterUserId"))
-    ctx["date_reserved"] = mkDate(data.get("dateReserved"))
-    ctx["date_updated"] = mkDate(data.get("dateUpdated"))
-    ctx["date_published"] = mkDate(data.get("datePublished"))
-    ctx["serial"] = data.get("serial", 1)
 
-    return models.CveRecord.objects.create(**ctx)
+    assert org is not None
```
RaitoBezarius, line 193:
```suggestion
    assert org is not None, "Organisation cannot be empty"
```

RaitoBezarius (approved):

PR #69, opened 2023-12-07
fix: m2m relationships on insertion

alejandrosame:
On insertion, the function `set_m2m` was getting the relationiships for the same class collapsed into a `None` key. This caused m2m relationships to get lost during the processing.

With this fix, the logic can again retrieve the proper foreign keys from the `self.m2m_models` list.

For example, now each `nixderivationmeta` gets correctly linked to its corresponding `nixderivationmeta_maintainers`. This can be checked after ingestion by running the query `SELECT * FROM shared_nixderivationmeta_maintainers;`.

### DIFF
--- a/src/website/shared/management/commands/ingest_manual_evaluation.py
+++ b/src/website/shared/management/commands/ingest_manual_evaluation.py
@@ -193,7 +193,7 @@ class BulkSave(object):
             if "op_pos" not in model.__dict__.keys():
                 raise RuntimeError(
                     "Temporary model id cannot be set. "
-                    + "Check that all `save` operations set a temporary if "
+                    + "Check that all `save` operations set a temporary id "
                     + "inside `model.op_pos`."
                 )
             mid = model.op_pos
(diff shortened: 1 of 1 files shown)


RaitoBezarius (approved):
This looks to be the right solution.

src/website/shared/management/commands/ingest_manual_evaluation.py
```diff
@@ -180,6 +184,19 @@ def set_m2m(self, model, attname, objects):
...
+        # Set up a temporary id from the operation position;
+        # otherwise all the m2m operations get collapsed into
+        # a None key and relationships get lost when calling
+        # save_m2m (that is, the m2m will only be set for the
+        # last entry found).
+        if model.pk is None:
+            if "op_pos" not in model.__dict__.keys():
+                raise RuntimeError(
+                    "Temporary model id cannot be set. "
+                    + "Check that all `save` operations set a temporary if "
```
fricklerhandwerk, line 196:
```suggestion
                    + "Check that all `save` operations set a temporary id "
```

Issue #43, opened 2023-11-26
False positive: CVE-2023-32611 in glib


raboof:
Running the local scanner on the testcase at https://github.com/Nix-Security-WG/nix-security-tracker/tree/e3e8d9a880e10e07a6942ee00e86294b5eb548fa, it reports [CVE-2023-32611](https://www.cve.org/CVERecord?id=CVE-2023-32611) in glib.

The problem is the version range in the CVE is wrong: it says all versions are affected, while in fact this issue was fixed in 2.74.2.

The best way to solve this issue would probably be to ingest the NVD feed, as https://nvd.nist.gov/vuln/detail/CVE-2023-32611 correctly has the version range as 'affected up to 2.74.2' #35.

GHSA does not have this improved data yet at the time of writing this issue (https://github.com/advisories/GHSA-qjp4-g2mq-4r5w)

raboof:
confirmed that this advisory is now filtered out based on the version range

PR #70, opened 2023-12-07
Add unit test

raboof:
As the first false positive I wanted to verify was gone now that we switched to NVD version matching, https://github.com/Nix-Security-WG/nix-security-tracker/issues/43, was actually just gone because the CVE couldn't be parsed in the first place.

Not working yet, struggling with hspec/LogT/ReaderT syntax


raboof:
thanks for the assist! looks like this confirms the parsing works as expected after all, so I'll be hunting for what seemed to be going wrong later in the process tomorrow.

PR #72, opened 2023-12-08
feat: add knownVulnerabilities to derivation metadata

fricklerhandwerk:
- use a JSON field, because it's a list of strings in Nixpkgs
  - m2m imposes additional burden on ingestion logic


RaitoBezarius:
no, please just use an arrayfield from postgresql: https://docs.djangoproject.com/en/4.2/ref/contrib/postgres/fields/

fricklerhandwerk:
Oh, are we hardcoding Postgres? Okay, makes things easier.

RaitoBezarius:
Now that we are leaning towards full text search and various things, yes. SQLite was just for bootstrapping. :)

fricklerhandwerk:
Rebased

src/website/shared/models/cve.py
```diff
@@ -293,13 +295,13 @@ class NixpkgsIssue(models.Model):
     created = models.DateField(auto_now_add=True)
     code = models.CharField(max_length=len("NIXPKGS-YYYY-") + 19)
 
-    cve = models.ManyToManyField(CveRecord)
```
RaitoBezarius, line 296: I don't understand what is the value of this change, also it seems unrelated to that PR

RaitoBezarius:
Also, we need the counterpart for the ingestion

src/website/shared/models/cve.py
```diff
@@ -300,6 +302,7 @@ class NixpkgsIssue(models.Model):
         choices=IssueStatus.choices,
         default=IssueStatus.UNKNOWN,
     )
+    derivations = models.ManyToManyField(NixDerivation)
```
RaitoBezarius, line 305: what does it serve as a purpose?

RaitoBezarius (approved):

PR #86, opened 2023-12-09
fix: make overlay work as intended

fricklerhandwerk:
the overlay packages did not end up in the final package set. now they are.

@alejandrosame


fricklerhandwerk:
Reviewed and fixed up with @RaitoBezarius

PR #87, opened 2023-12-09
style: set up `nixfmt` according to RFC 166

fricklerhandwerk:


fricklerhandwerk:
Discussed with @RaitoBezarius

PR #92, opened 2023-12-10
feat: keep CVE data in a local cache directory for efficient reimports

alejandrosame:
Most of the time spent on CVE import is fetching and uncompressing the CVE bundle from Github. This penalizes development work on the database, where multiple reimports might be needed.

Now the CVE data gets stored in a local cache directory after the first fetch. It's still possible to force the deletion of the CVE cache to download the bundle again with the `--force-download` option.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ __pycache__
 .credentials
 *.jsonl # Evaluation results
 CACHE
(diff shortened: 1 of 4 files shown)


fricklerhandwerk:
Nice. Tested, works as promised.

PR #93, opened 2023-12-10
feat: store CVE per derivation

fricklerhandwerk:
@raitobezarius


RaitoBezarius (changes requested):
Rationale? This makes it impossible to group CVEs to one Nix issue and we seemingly completely lose the interest of the security tracker here?

fricklerhandwerk:
I was under the impression we'd want to have that fine-grained information. Apparently not.

PR #99, opened 2023-12-11
feat: display knownVulnerabilities per derivation

fricklerhandwerk:
depends on # 66


fricklerhandwerk:
Reviewed with @RaitoBezarius

PR #100, opened 2023-12-11
fix: merge migrations

fricklerhandwerk:
merged #95 without noticing that it skipped a migration


alejandrosame:
I just noticed this error, merging to fix.

PR #115, opened 2023-12-11
style: apply nixfmt globally

fricklerhandwerk:
before introducing noise in every commit, let's settle this once.
this was made necessary by #87, sorry.


alejandrosame (approved):

alejandrosame:
#87 without formatting the tree files is indeed making the diffs noisy

PR #119, opened 2023-12-11
feat: test nginx reverse proxy as well

fricklerhandwerk:
now we can also test static files being served

### DIFF
--- a/nix/tests/utils.nix
+++ b/nix/tests/utils.nix
@@ -13,7 +13,7 @@ in
         machine.wait_for_unit("multi-user.target")
         machine.wait_for_unit("web-security-tracker-server.service")
         machine.wait_for_open_port(8000)
-        machine.succeed("curl http://127.0.0.1:8000")
+        machine.succeed("curl --fail -H 'Host: example.org' http://127.0.0.1:80/static/style.css")
         machine.succeed("wst-manage test")
       '';
       defaults = {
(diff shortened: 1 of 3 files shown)


nix/tests/utils.nix
```diff
@@ -19,20 +19,31 @@
         machine.wait_for_unit("multi-user.target")
         machine.wait_for_unit("web-security-tracker-server.service")
         machine.wait_for_open_port(8000)
-        machine.succeed("curl http://127.0.0.1:8000")
+        machine.succeed("curl --fail -H 'Host: example.org' http://127.0.0.1:80/static/style.css")
```
RaitoBezarius, line 22:
```suggestion
        machine.succeed("curl --fail -H 'Host: example.org' http://127.0.0.1/static/style.css")
```

nix/tests/utils.nix
```diff
@@ -19,20 +19,31 @@
...
+        let
+          cfg = config.services.web-security-tracker;
+        in
         {
           imports = [
             wstModule
             machine
           ];
 
+          services.nginx.virtualHosts.${cfg.domain}.listen = [
```
RaitoBezarius, line 38: This should be a default, why are we doing that in the test?

RaitoBezarius (dismissed):
This looks good to me modulo redundant defaults.

PR #122, opened 2023-12-12
fix: default values for maintainer's github values

alejandrosame:
Commit `b57ab937463868f95556ad8ec5b577700df79897` was missing a migration and a model update to make all the migrations work seamlessly to make `manage makemigrations` an idempotent operation after `mange migrate`.


RaitoBezarius:
NACK, we should have no integrity bullshit in the database, please just drop anything that doesn't have a proper github like I did it in the prior migrations.

alejandrosame:
@raitobezarius My database is empty. This is a blocker introduced by the database changes done in the commit I mention (that commit leaves the repo in a broken state where no further migrations can be done). In practice, at most one entry would have those default values given the uniqueness of those fields.

In any case, after a break I checked that this doesn't fix the problem either :( so I'll just make another PR.

I'd suggest squashing migrations so from the POV of django we don't have to manually ignore the change from nullable to non-nullable fields. We know these values are dropped on insertion, but the migration logic doesn't.

RaitoBezarius:
@alejandrosame I agree that we should squash the migrations, I have a fix for that somewhere in my #123 PR.

RaitoBezarius:
You can cherry-pick it.

PR #125, opened 2023-12-12
Invoking the ws

raboof:
Seems to invoke the webservice fine.

TODO:
* add a flag to specify the online tracker URI, or to skip the network call entirely
* test with actual data that the result is correct
* by default don't output advisories that have a nixpkgs issue in `notforus` state or similar


raboof:
Tested, made some final tweaks, confirmed works!

PR #126, opened 2023-12-14
feat: nixpkgsissue activity log view

alejandrosame:
feat: nixpkgsissue activity log view
    
Improve the messy representation of the different `pghistory` event tables:
    
- Create a view `log_nixpkgsissue` in the PostgreSQL backend and introduce an unmanaged model `NixpkgsIssueLogView` to display a list of changes per timestamp and user.
- Register `NixpkgsIssueLogView` in the `pghistory` app to declutter the log reporting from the already crowded `shared` section.
- Delete the previous activitiy log models and views in favour of the unified view model.
    
If performance starts being problematic for the `log_nixpkgsissue` view, we can
    make it a materialized view and add a trigger to update its contents on `shared.nixpkgsissue`
    and it's m2m relationships updates.

**NOTE:** This PR includes #101


src/website/shared/models/cve.py
```diff
@@ -301,6 +304,11 @@ class IssueStatus(models.TextChoices):
     WONTFIX = "W", _("wontfix")
 
 
+# `cve` and `derivations` fields have to be tracked with via their `trough` models
```
RaitoBezarius, line 307:
```suggestion
# `cve` and `derivations` fields have to be tracked with via their `through` models
```

src/website/shared/models/cve.py
```diff
@@ -344,6 +352,49 @@ def generate_code(
...
+    model_name="NixpkgsIssueDerivationsLog",
+)
+class NixpkgsIssueDerivation(NixpkgsIssue.derivations.through):
+    class Meta:
+        proxy = True
+
+
+class NixpkgsIssueLogView(models.Model):
+    class Meta:
+        managed = False
```
RaitoBezarius, line 379: Can we add a comment on why this is not managed by the ORM?

RaitoBezarius (approved):
Looks mostly good to me, I am just thinking about the GDPR considerations with the activity log, it's a good idea to make the cascade do nothing in case of deletion, but I wonder if we ought to implement a soft delete operation and bake in GDPR stuff in there directly.

src/website/shared/models/cve.py
alejandrosame: I pasted the wrong commit message in the PR description. I'll add similar comments in the model + the fact that Django doesn't seem to support multi table proxy models (which would be the equivalent of VIEW tables in ORM).

alejandrosame:
@RaitoBezarius the cascade does nothing because it prevents Django from trying to delete content from a VIEW, which would just yield an error (it's read-only after all).

The data is collected from other `auth` and `pghistory` tables so we have to double check there what exactly we want to do on user deletion. There's also the default Django `admin` table, which registers less finegrained changes made by users.

I'll make an issue to not to forget about looking into this.

src/website/shared/models/cve.py
alejandrosame: fixed

src/website/shared/models/cve.py
alejandrosame: comment added

alejandrosame:
Please don't merge yet. Working on some other functionality I noticed a mistake on the way I set the migration file in this PR.

I'll take this out of draft as soon as I push the fix.

alejandrosame:
The PR is in a mergeable state again.

alejandrosame:
rebased with latest changes

alejandrosame:
@RaitoBezarius if you can reconfirm that you are ok with this approach after the rebases, this is ready to be merged.

fricklerhandwerk:
Towards https://github.com/Nix-Security-WG/nix-security-tracker/issues/169

alejandrosame:
We have activity log for suggestions with #366. If we were to reintroduce activity logs for issues, it should be based on top of what was implemented there.

PR #140, opened 2023-12-18
feat: add view for affected packages per channel

alejandrosame:
Create a simple tabbed view for those derivations that appear in an issue marked as 'affected'.

The styles broke in weird ways when moved to the main `style.css`, so I leave them in the template document for the time being.

### DIFF
--- a/src/website/webview/urls.py
+++ b/src/website/webview/urls.py
@@ -28,6 +28,7 @@ urlpatterns = [
         NixderivationPerChannelView.as_view(),
         name="affected_list_per_channel",
     ),
+    # TODO: We may want to put an overview page here
     path(
         "affected/",
         RedirectView.as_view(url="nixos-unstable", permanent=True),
--- a/src/website/webview/urls.py
+++ b/src/website/webview/urls.py
@@ -1,7 +1,14 @@
 from django.urls import path, re_path
+from django.views.generic.base import RedirectView
 from shared.auth.github_webhook import handle_github_hook
 
-from webview.views import HomeView, NixpkgsIssueListView, NixpkgsIssueView, TriageView
+from webview.views import (
+    HomeView,
+    NixderivationPerChannelView,
+    NixpkgsIssueListView,
+    NixpkgsIssueView,
+    TriageView,
+)
 
 app_name = "webview"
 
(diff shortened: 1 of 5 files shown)


src/website/shared/models/cve.py
```diff
@@ -368,3 +369,41 @@ class AdvisorySeverity(models.TextChoices):
         CRITICAL = "CRITICAL", _("critical")
 
     issues = models.ManyToManyField(NixpkgsIssue)
+
+
+class NixDerivationAffectedView(models.Model):
```
RaitoBezarius, line 374: Not sure I understand here, shouldn't the list of derivations affected derive from an NixpkgsIssue?

alejandrosame: that's how the function view `affected_derivation_per_channel_view` is defined.

In any case, I'll push a change soon where I'll get rid of the view table and achieve the same with a queryset.

alejandrosame:
Rebased and updated code to bring this PR to a mergeable state.

src/website/webview/views.py
```diff
@@ -401,3 +403,44 @@ class NixpkgsIssueListView(ListView):
...
+                drv_system=F("derivations__system"),
+                drv_path=F("derivations__derivation_path"),
+                channel_id=F("derivations__parent_evaluation__channel_id"),
+            )
+            .filter(channel_id=channel.channel_branch)
+        )
+
+    def get_context_data(self, **kwargs: Any) -> Any:
+        context = super().get_context_data(**kwargs)
+        context["channels"] = ["NIXOS-UNSTABLE", "NIXOS-24.05", "NIXOS-23.11"]
```
fricklerhandwerk, line 438: Why are existing channels hard coded? I didn't take time to fully penetrate the code, but what speaks against taking the list of channels from the evaluation data?

alejandrosame: Implemented in the last commit. The tabs look like this now:
image

src/website/webview/urls.py
```diff
@@ -16,4 +23,14 @@
...
+    re_path(
+        "^affected/(?P<channel>(nixos|nixpkgs)-.*)$",
+        NixderivationPerChannelView.as_view(),
+        name="affected_list_per_channel",
+    ),
+    path(
+        "affected/",
+        RedirectView.as_view(url="nixos-unstable", permanent=True),
+        name="affected_list",
+    ),
```
fricklerhandwerk, line 35: Maybe we should reserve this for an overview page, e.g. listing all stored channels? But this would be for later, so no real change required now.

```suggestion
    # XXX: We may want to put an overview page here
    path(
        "affected/",
        RedirectView.as_view(url="nixos-unstable", permanent=True),
        name="affected_list",
    ),
```

fricklerhandwerk (approved):
Looks pretty good now

src/website/webview/urls.py
alejandrosame: Pushed comment. Squashing and merging

PR #143, opened 2023-12-18
Fix manual ingestion, add manage to the nix shell

Yasuke:


RaitoBezarius (approved):

default.nix
```diff
@@ -13,6 +13,10 @@ let
     package = pkgs.web-security-tracker;
     module = import ./nix/web-security-tracker.nix;
 
+    manage = pkgs.writeScriptBin "manage" ''
+      ${python3}/bin/python ./src/website/manage.py $@
```
fricklerhandwerk, line 17:
```suggestion
      ${python3}/bin/python ${toString ./src/website/manage.py} $@
```

This makes it independent of CWD

PR #123, opened 2023-12-12
Holy grail: end-to-end pipeline for evaluation of Nixpkgs

RaitoBezarius:
Fixes #8.

PR #145, opened 2023-12-19
add code for managing a Nixpkgs checkout

fricklerhandwerk:
breakdown of #123

Does a few renamings and fixes missing type annotations to appease the linter


fricklerhandwerk:
Tested the hell out of since I had to fix the build. Merging, also because #123 needs more cleanup work that's otherwise stalled.

PR #148, opened 2023-12-19
feat(models): improve the Nix evaluation data models

fricklerhandwerk:
We now have more fields for the evaluation, and timestamping in preparation
for managing the Nixpkgs repository.

as part of breaking down #123


RaitoBezarius:
Please hold off with this PR, I am not totally done with the fields yet.

fricklerhandwerk:
Yep, there are a few future commits which still change them. I'd like to have as many changes at once as possible, otherwise we'll just be piling up migrations.

RaitoBezarius:
> Yep, there are a few future commits which still change them. I'd like to have as many changes at once as possible, otherwise we'll just be piling up migrations.

The goal is to avoid having 50 migrations, so while I am developing in the branch, if you chip away individual migrations, I cannot squash them anymore.

fricklerhandwerk:
Generally, can we squash away all migrations before we deploy to production?

RaitoBezarius:
We can but I don't recommend squashing migrations more than necessary, they are valuable development knowledge. I recommend squashing migrations done too hastily in development.

RaitoBezarius:
#123 is ready, I propose to hop on this one rather and merge it directly.

PR #152, opened 2023-12-19
Add basic issue linking (and unlinking)

Yasuke:


src/website/webview/urls.py
```diff
@@ -13,4 +19,6 @@
         NixpkgsIssueView.as_view(),
         name="issue_detail",
     ),
+    path("derviations/<id>", NixDerivationsView.as_view(), name="derivation_detail"),
```
RaitoBezarius, line 22:
```suggestion
    path("derivations/<id>", NixDerivationsView.as_view(), name="derivation_detail"),
```

src/website/webview/urls.py
```diff
@@ -13,4 +19,6 @@
...
+    path("link_issues/<id>", LinkIssuesView.as_view(), name="link_issues_detail"),
```
RaitoBezarius, line 23:
```suggestion
    path("link-issues/<id>", LinkIssuesView.as_view(), name="link_issues_detail"),
```

src/website/shared/templates/base.html
```diff
@@ -1,61 +1,66 @@
...
   <head>
     <title>{% block title %}Nixpkgs Security Tracker{% endblock %}</title>
     <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
     <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
-    <link rel="stylesheet" type="text/css" href="/static/style.css" />
+    <script src="https://cdn.tailwindcss.com"></script>
+    <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Space+Mono">
```
fricklerhandwerk, line 13: I'm not a fan of those, as it pollutes the templates with non-semantic noise that's hard to read (and make any sense of to begin with if you don't happen to know all of the framework). I won't block on it today because we have to get shit done, but really this is a maintenance liability.

src/website/shared/templates/base.html
```diff
@@ -1,61 +1,66 @@
...
+        </div>
         {% endif %}
         {% endblock %}
-      </ul>
+      </div>
     </nav>
 
     {% block layout %}
-    <article>
+    <article class="flex-grow bg-blue-100 overflow-y-scroll">
```
fricklerhandwerk, line 55: This can easily be expressed in the regular style sheet, where it belongs. Unless there's something important for displaying the new information at all I'd rather not have any presentational aspect in this PR.

src/website/webview/views.py
```diff
@@ -1,15 +1,67 @@
...
+    NixpkgsIssue,
+)
 
 
-class HomeView(TemplateView):
+class HomeView(ListView):
     template_name = "home_view.html"
 
+    def get_queryset(self) -> BaseManager[NixDerivationMeta]:
+        return NixDerivationMeta.objects.all()
```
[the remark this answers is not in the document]
RaitoBezarius: Correct, there's like 1.8M of them IIRC.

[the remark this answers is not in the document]
RaitoBezarius: But this queryset should and can be filtered further (and paginated) to avoid this.

src/website/shared/templates/base.html
Yasuke: @fricklerhandwerk as progress gets made the eventual result is simply a tailwind config and generated css file that applies the style of the site based on much simpler rules and much less inline css. 

As a WIP however, to be able to make adjustments on the fly and discover out various styles that make sense, we start from the CDN + exploratory styles.

Otherwise one has to explore out the html, and manage the complexity of the CSS at the same time and refactoring CSS is futile endeavor.

Put simply, this wouldn't remain this way as the style and widgets present on the site stabilize, but as an early PR that is the first in many, complexity will go up before it gets managed back down during this discovery phase.

src/website/shared/templates/base.html
Yasuke: see [here](https://github.com/Nix-Security-WG/nix-security-tracker/pull/152#discussion_r1432941422)

PR #157, opened 2024-06-10
docs: make the database setup instructions to work immediately

fricklerhandwerk:
@alejandrosame tried to set it up from scratch and ran into this


alejandrosame (approved):
I thought this was the case already, so +1.

PR #158, opened 2024-06-11
fix: update reset script commands

alejandrosame:
I decided to reset the dev environment and ran into issues with the default setup. This PR solves the inconsistencies found so that `./contrib/reset.sh` works again.

Ideally, we should converge the behaviour of the ingestion through the listener and the manual ingestion so it's clear how to do it as a one-shot operation.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -86,6 +86,9 @@ let
       pkgs.mkShell {
         DATA_CACHE_DIRECTORY = toString ./. + "/.data_cache";
         REDIS_SOCKET_URL = "unix:///run/redis/redis.sock";
+        # `./src/website/tracker/settings.py` by default looks for LOCAL_NIXPKGS_CHECKOUT
+        # in the root of the repo. Make it the default here for local development.
+        LOCAL_NIXPKGS_CHECKOUT = toString ./. + "/nixpkgs";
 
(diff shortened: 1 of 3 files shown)


default.nix
```diff
@@ -86,6 +86,9 @@ let
       pkgs.mkShell {
         DATA_CACHE_DIRECTORY = toString ./. + "/.data_cache";
         REDIS_SOCKET_URL = "unix:///run/redis/redis.sock";
+        # `./src/website/tracker/settings.py` by default looks for LOCAL_NIXPKGS_CHECKOUT
+        # in the root of the repo. Make it the default here for local development.
+        LOCAL_NIXPKGS_CHECKOUT = toString ./. + "/nixpkgs";
```
RaitoBezarius, line 91: Hm, this seems a bit weird, by default, we should probably reuse global defaults or something.

RaitoBezarius (approved):

default.nix
RaitoBezarius: Or maybe it should end up in some `.state` directory, not sure.

RaitoBezarius: Or a cache directory like `DATA_CACHE_DIRECTORY`

PR #159, opened 2024-06-18
feat: WIP Github Oauth integration

alejandrosame:


src/website/shared/templatetags/gh_tags.py
```diff
@@ -0,0 +1,101 @@
...
+from github.Team import Team
+from shared.models import SocialUser
+from shared.utils import get_gh
+
+register = template.Library()
+github: Github = get_gh()
+logger = logging.getLogger(__name__)
+
+
+def get_gh_username(user: User) -> str | None:
```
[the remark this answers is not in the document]
alejandrosame: Right now there's no caching, and that's a problem of course. The tags are going away, they just allowed me to quickly work with the Github calls.

My plan is to move the logic to map Github memberships to Django's group membership and do authorization from there. These mappings would act as a cache and then the question will be how often do we want to refresh them.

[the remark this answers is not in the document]
RaitoBezarius: In that case, it must be very explicit what is the upperbound on the refresh for the authorizations and I suggest to have an systemd timer to refresh them every X and let user figure out what they want.

src/website/shared/management/commands/update_groups.py
```diff
@@ -0,0 +1,77 @@
...
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from shared.templatetags.gh_tags import get_gh_organization, get_gh_team
+
+# TODO: move to shared.auth or shared.auth_utils
+
+logger = logging.getLogger(__name__)
+
+
+def get_team_member_ids(orgname: str, teamname: str) -> set[int]:
```
RaitoBezarius: Any reason not to use `@functools.lru_cache` with an appropriate size?

src/website/shared/management/commands/update_groups.py
```diff
@@ -0,0 +1,77 @@
...
+            group_objects[groupname] = Group.objects.get(name=groupname)
+
+        logger.info("Using Github ID cache to update database groups...")
+
+        users = User.objects.all()
+        for user in users:
+            social = user.socialaccount_set.filter(provider="github").first()  # type: ignore
+            if social:
+                # Open a single transaction for the db
+                with transaction.atomic():
```
RaitoBezarius, line 38: What does it serve to be atomic over each user instead of being atomic over all the update?

src/website/shared/templatetags/gh_tags.py
```diff
@@ -20,8 +19,10 @@ def get_gh_username(user: User) -> str | None:
     """
     Return the Github username of a given Auth.User.
     """
-    social_user = SocialUser.objects.get(id=user.id)  # type: ignore
-    social_account: SocialAccount | None = social_user.github_account
+    social_user = User.objects.get(id=user.id)  # type: ignore
+    social_account: SocialAccount | None = (
+        social_user.socialaccount_set.filter(provider="github").first()  # type: ignore
```
RaitoBezarius: Can a user have multiple socialaccounts of type GitHub, does it make sense? It look like an integrity issue to me.

src/website/shared/apps.py
```diff
@@ -7,3 +8,10 @@ class SharedConfig(AppConfig):
 
     def ready(self) -> None:
         import shared.listeners  # noqa
+        from shared.auth import reset_group_permissions
+
+        # Configuration of signals
+
+        # Reset general group perimissions after migrations to
+        # include potential new permissions added.
+        post_migrate.connect(reset_group_permissions, sender=self)
```
RaitoBezarius, line 18: Can you provide a concrete usecase for this? I'm not exactly sure it makes sense.

src/website/shared/auth.py
```diff
@@ -0,0 +1,143 @@
...
+    Return the Github username of a given Auth.User.
+    """
+    social_user = User.objects.get(id=user.id)  # type: ignore
+    social_account: SocialAccount | None = (
+        social_user.socialaccount_set.filter(provider="github").first()  # type: ignore
+    )
+    if social_account:
+        return social_account.extra_data.get("login")  # type: ignore
+
+    logger.warning(f"Failed to get GitHub username for user {user}.")
```
RaitoBezarius, line 29: Please use `logger`'s interpolation as much as possible and do not resort to external formatting, otherwise the logger lose the opportunity to tap into the object you pass.

src/website/shared/auth.py
```diff
@@ -0,0 +1,143 @@
...
+def get_gh_username(user: User) -> str | None:
+    """
+    Return the Github username of a given Auth.User.
+    """
+    social_user = User.objects.get(id=user.id)  # type: ignore
+    social_account: SocialAccount | None = (
+        social_user.socialaccount_set.filter(provider="github").first()  # type: ignore
+    )
+    if social_account:
+        return social_account.extra_data.get("login")  # type: ignore
```
RaitoBezarius, line 27: Sounds like what you care about is `.values_list('extra_data__login')` or something like that, no? You could get rid of the many attributes that could get passed via the SQL query.

src/website/shared/auth.py
```diff
@@ -0,0 +1,143 @@
...
+
+
+def get_gh_organization(orgname: str) -> Organization | None:
+    """
+    Return the Github Organization instance given an organization name.
+    """
+    try:
+        return github.get_organization(login=orgname)
+    except Exception as e:
+        logger.warning(f"Failed to get organization {orgname}: {e}")
```
RaitoBezarius, line 40: `logger.exception` should probably be used here.

src/website/shared/management/commands/update_groups.py
alejandrosame: Hmm, I remember thinking on the spot about the inconvenience of a partial update of permissions for any given user. However, with the caching being eventually set up, there's less of a reason to not make the whole update atomic.

I'll change it.

src/website/shared/auth.py
alejandrosame: Doing
```
def get_gh_username(user: User) -> str | None:
    """
    Return the Github username of a given Auth.User.
    """
    social_user = User.objects.get(id=user.id)  # type: ignore
    try: 
        extra_login: str = (
            social_user.socialaccount_set.filter(provider="github")
                .values_list('extra_data__login', flat=True)[0]
        )
        return extra_login
    except Exception:
        logger.exception("Failed to get GitHub username for user %s.", user)
        return None
```
Gives me 
```
django.core.exceptions.FieldError: Unsupported lookup 'login' for JSONField or join on the field not permitted.
```

For now I'll keep the `extra_data.get("login")` and give this another spin later.

src/website/shared/templatetags/gh_tags.py
alejandrosame: No, you are right, They would be two different accounts. 

I was doing the `filter`...`first` pattern to deal superusers that don't necessarily need to log in via github. I'm changing those patterns to the following with try except to make this intention more clear:

```
def get_gh_username(user: User) -> str | None:
    """
    Return the Github username of a given Auth.User.
    """
    social_user = User.objects.get(id=user.id)  # type: ignore
    try: 
        extra_data_login: str = (
            social_user.socialaccount_set.get(provider="github")
                .extra_data.get("login")
        )
        return extra_data_login
    except Exception:
        logger.exception("Failed to get GitHub username for user %s.", user)
        return None
```

src/website/shared/auth.py
alejandrosame: Done [c698428](https://github.com/Nix-Security-WG/nix-security-tracker/pull/159/commits/c69842824f83be36e467f0d997f0afead9e0e01f)

src/website/shared/auth.py
alejandrosame: Done [c698428](https://github.com/Nix-Security-WG/nix-security-tracker/pull/159/commits/c69842824f83be36e467f0d997f0afead9e0e01f)

src/website/shared/management/commands/update_groups.py
alejandrosame: Changed in [aa398c9](https://github.com/Nix-Security-WG/nix-security-tracker/pull/159/commits/aa398c923360d17d9ef2bf4f39af74f51212289a)

src/website/shared/templatetags/gh_tags.py
alejandrosame: For example: [6602bc8](https://github.com/Nix-Security-WG/nix-security-tracker/pull/159/commits/6602bc892ab83140b894233d73f1f1b42616b2e4)

src/website/shared/apps.py
alejandrosame: When new tables are created, new permission values get created. These will have to be assigned to admins, at least (superusers bypass permissions and get access to everything).

The alternative is getting rid of this signal and add the function as part of a relevant migration (or as a subsequent migration) when we notice some permissions being missed by admins.

The part of that function that relates to readers could be left apart as more  tables prefixed with nix won't probably be added.

src/website/shared/management/commands/update_groups.py
alejandrosame: When I was mentioning caching I was thinking about database caching and getting the info from there, but I see that you meant memoization.

I'll implement it with functools decorators. I think to make it work the memoized function will have to be triggered by a signal so it's always executed by the running webserver process, otherwise the command will not profit from the memoized cache.

Then the systemd timer can call this command by invalidating first the cache.

As an alternative to the systemd timer, I saw that cachedtools has timed versions of functools caches with `TTLCache`. I haven't read the implementation yet to see the tradeoffs, and it would also mean configuring the TTL at settings level, which seems less flexible that changing the refresh timer.

src/website/shared/apps.py
alejandrosame: Permissions are now based on custom view mixins.

src/website/shared/tests/test_auth.py
```diff
@@ -0,0 +1,199 @@
...
+        # Make sure that only one issue was created
+        self.assertEqual(models.NixpkgsIssue.objects.count(), 1)
+
+        # And that the changelist view only shows one issue
+        redirect_response = self.client.get(
+            reverse("admin:shared_nixpkgsissue_changelist")
+        )
+        self.assertContains(redirect_response, "1 nixpkgs issue", status_code=200)
+
+    def test_committer_cannot_add_non_related_issue_from_admin_site(self) -> None:
```
fricklerhandwerk, line 136:
```suggestion
    def test_maintainer_cannot_add_non_related_issue_from_admin_site(self) -> None:
```

This only tests what *maintainers* can do

fricklerhandwerk: Also there seems to be no test for actual committers.

src/website/shared/auth.py
```diff
@@ -0,0 +1,116 @@
...
+@lru_cache(maxsize=1)
+def get_gh_organization(orgname: str) -> Organization | None:
+    """
+    Return the Github Organization instance given an organization name.
+    """
+    try:
+        return github.get_organization(login=orgname)
+    except Exception:
+        logger.exception("Failed to get organization %s", orgname)
+        return None
```
fricklerhandwerk, line 29: This is not needed as we're only caching two teams, so it's at most two calls to `get_organization` for an application lifetime.

src/website/shared/auth.py
```diff
@@ -0,0 +1,116 @@
...
+        # The iterator will make the extra page API calls for us.
+        return {member.id for member in members}
+    return set()
+
+
+def is_team_member(username: str, orgname: str, teamname: str) -> bool:
+    """
+    Return whether a given username is a member of a Github team
+    """
+    gh_named_user: NamedUser = cast(NamedUser, github.get_user(login=username))
```
fricklerhandwerk, line 74: We can use [`get_user_by_id`](https://pygithub.readthedocs.io/en/stable/github.html?highlight=get_user#github.MainClass.Github.get_user_by_id) here, which returns a `NamedUser`

src/website/shared/management/commands/update_groups.py
```diff
@@ -0,0 +1,62 @@
...
+from django.conf import settings
+from django.contrib.auth.models import Group, User
+from django.core.management.base import BaseCommand
+from django.db import transaction
+from shared.auth import get_team_member_ids
+
+logger = logging.getLogger(__name__)
+
+
+class Command(BaseCommand):
```
fricklerhandwerk, line 14: We probably rather want to set up a webhook on [team membership events](https://docs.github.com/en/webhooks/webhook-events-and-payloads#membership)

alejandrosame:
Closed in favor of #165

PR #185, opened 2024-09-24
chore: update npins

erictapen:
I'm putting this out for now, even though the updated `pyright` linter throws a lot of errors now. Run `pyright` in the nix-shell to reproduce. Would be happy if somebody could take a look at these, as I'm not savvy yet in Python type annotations.

Should contain only cosmetic changes that are related to updating all the npins (specifically Nixpkgs) to the most recent versions.

- Use nixfmt-rfc-style from Nixpkgs
- remove a few python packages that landed in Nixpkgs
- format Nix code with new nixfmt
- format/lint Python code with ruff, prettier, pyright


erictapen:
Now all the pre-commit checks work out, but the diff is quite considerable. Especially the formatting by `prettier` is really not looking great, but I guess it just changed/got stricter with the new version?

src/website/shared/templates/socialaccount/login.html
```diff
@@ -1,23 +1,23 @@
-{% extends "base.html" %}
-
-{% block title %}Sign In{% endblock %}
-
-{% block auth %}{% endblock %}
-
-{% block content %}
-{% if process == "connect" %}
+{% extends "base.html" %} {% block title %}Sign In{% endblock %} {% block auth
```
fricklerhandwerk, line 1: Most of the templates got a lot harder to read... Can we exclude that?

PR #191, opened 2024-09-25
fix: run wst-manage in service directory

fricklerhandwerk:
for everything else it will error on CHDIR, even if $PWD has the right
owner and permissions.


proofconstruction (approved):
*sigh* why is it always like this

PR #192, opened 2024-09-25
feat: allow deployment on local network

fricklerhandwerk:
removing constraints on the Host header in development mode allows
serving requests from the local network


proofconstruction (approved):
Yeah we need this

PR #193, opened 2024-09-25
fix: add `git` to service environment

fricklerhandwerk:
`fetch-all-channels` uses Git underneath, make sure it's available.


proofconstruction (approved):
Good catch

PR #194, opened 2024-09-25
refactor: single source of truth for GIT_CLONE_URL

fricklerhandwerk:


proofconstruction (approved):
Good, we should remove magic strings whenever possible

PR #195, opened 2024-09-25
fix: get periodic evaluation to run

fricklerhandwerk:


proofconstruction (approved):
:+1: but this also makes me wonder whether it would ever make sense to just pull all the rest of these settings out into the nix module like [we do here](https://github.com/Nix-Security-WG/nix-security-tracker/blob/8c115c4b51987f334dc6020b2a62b9c76e06c302/nix/web-security-tracker.nix#L108-L121). It's a slippery slope to just codegenning all the python though.

fricklerhandwerk:
That piece is not pulling it out, it's overriding the settings. And we need defaults to run the front end in the shell for local testing.

PR #230, opened 2024-09-27
Documentation updates

proofconstruction:
A couple commits to start making development a bit less painful, and onboarding a bit more... possible.


proofconstruction:
Some of the new `README` files could use some more detail.

I mostly wanted to start documenting everything in the Django project, and thought it made sense to start on the "outside" of the project before doing so.

npins/README.md
fricklerhandwerk, line 1: This makes little sense to have

contrib/README.md
fricklerhandwerk, line 1: The files should have comments explaining what the scripts are about. The directory itself shouldn't need more than a short blurb telling what it's for

CONTRIBUTING.md
```diff
@@ -0,0 +1,236 @@
...
+## Overview
+
+This file contains general contribution information, but the other directories in this repository have additional `README.md` files with more specific information relevant to their sibling files:
+
+- [`src/website/`](./src/website/README.md): Sources and documentation for the web tracker.
+- [`nix`](./nix/README.md): Nix expressions for the service, tests, and the pkgs overlay.
+- [`pkgs`](./pkgs/README.md): Nix derivations for vendored Python dependencies.
+- [`contrib`](./contrib/README.md): Useful scripts for developing the project.
+- [`npins`](./npins/README.md): Pinned sources.
+- [`staging`](./staging/README.md): System configurations and service definitions for deployment.
```
fricklerhandwerk, line 14: Tending to duplicate information is error-prone and increases friction when moving things around. Ideally we'd generate the table of contents automatically on each commit, but it's not wirth doing it now. I suggest that instead just say that each directory has a README explaining what it's about, and both reduce the number of directories and improve their legibility by organizing the code in a more obvious fashion.

For instance, IMO all the Nix-related support stuff, including `npins`, should just go into `nix`; `staging` should be called `deployment` -- if we need it at all; why would you store any machine-specifics in this repo? all you need are good defaults and module documentation -- and we probably don't need 3 Django applications either.

fricklerhandwerk (commented):
Moving the contributor docs to a separate file is good, then we have more space in the README to explain what the project is about. But we need to keep the contributor docs scalable; left some comments on that.

The next steps after this would be updating the actual contribution docs and figuring out a way to unify the workflows for local development and deployment, as currently the development shell works very different from a NixOS deployment, which eats into the time to get used to working with the system.

CONTRIBUTING.md
```diff
@@ -0,0 +1,285 @@
...
+> GitHub credentials are required to use the OAuth login feature, but fake secrets can be used for local development.
+> To get going quickly, run the following script to generate secrets required by the server:
+>
+> ```console
+> ./contrib/mkcredentials.sh
+> ```
+
+1. Create a new or select an existing GitHub organisation to associate with the application
+
+   In the organisation settings, under **Personal access tokens** go through the questionnaire to allow personal access tokens.
```
fricklerhandwerk, line 97: @Erictapen merging this will require conflict resolution on your PR updating these lines. Feel free to either merge it and fix up yours, or the other way round. Or @proofconstruction, whoever comes first.

fricklerhandwerk (approved):

CONTRIBUTING.md
erictapen: We can merge this first, I'll fixup

PR #232, opened 2024-09-29
feat: export packages

fricklerhandwerk:
this way one can deploy the service as a container and rest assured that
it will work. otherwise there will almost certainly be conflicts with
the host system's Nixpkgs version.

Example:

```nix
{ config, ... }:
let
  sectracker = import ~/src/nix-security-tracker { };
in
{
  containers.nix-security-tracker = {
    # ...
    config = { ... }: {
      nixpkgs = {
        inherit (sectracker) pkgs overlays;
      };
      imports = [
        sectracker.module
      ];
      # ...
    };
  };
}
```

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -5,6 +5,7 @@
 }:
 let
   self = rec {
+    inherit pkgs;
     inherit (pkgs) python3;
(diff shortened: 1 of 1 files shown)


erictapen (approved):
Makes sense for this specific use case

PR #233, opened 2024-09-30
feat: add container configuration for local development

fricklerhandwerk:


erictapen:
This doesn't work yet on my flakes based setup (e.g. as `.credentials/` isn't tracked by git. I was working on something similar and I think it would be great to have dev module for NixOS users. For a pure version (without builtins.currentSystem and untracked directories) it might require a thin options interface to use.

fricklerhandwerk:
@erictapen let's fix foward, glad to review and merge PRs making the setup more pure.

PR #234, opened 2024-10-01
refactor: simplify default.nix attrset

fricklerhandwerk:
the diff looks huge, but semantically this is a tiny change:
add `lib` to the top-level arguments and don't merge tests separately
since they're already part of the `rec` attrset.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -7,127 +7,125 @@
     overlays = [ overlay ];
     inherit system;
   },
+  lib ? import "${sources.nixpkgs}/lib",
 }:
-let
-  self = rec {
-    inherit pkgs;
-    inherit (pkgs) python3;
-    localPythonPackages = import ./pkgs { inherit pkgs python3; };
+rec {
+  inherit pkgs;
+  inherit (pkgs) python3;
+  localPythonPackages = import ./pkgs { inherit pkgs python3; };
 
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -7,127 +7,125 @@
     overlays = [ overlay ];
     inherit system;
   },
+  lib ? import "${sources.nixpkgs}/lib",
```
erictapen, line 10: What's the benefit over using `lib ? pkgs.lib` here? This way a custom pkgs's lib would automatically used along.

erictapen (approved):

default.nix
fricklerhandwerk: This may be a bit of an idiosyncratic style thing of mine, but doing it this way gives another layer of confidence that `lib` is pure. [The Nixpkgs entry point is impure by default](https://github.com/NixOS/nixpkgs/issues/339635), and lumping together `pkgs` and `lib` is kind of dirty. That's one thing the Nixpkgs' flake outputs do better, and I'm mimicking that here.

It doesn't matter in practice, it's more of a heads-up to readers. (And seemed to do the trick since it raised your eyebrows.)

erictapen: Okay, understood!

PR #248, opened 2024-10-02
fix(build): drop postgis as it seems unneeded in the codebase

RaitoBezarius:
It doesn't seem that PostGIS is required for this project (yet)?

### DIFF
--- a/nix/dev-setup.nix
+++ b/nix/dev-setup.nix
@@ -5,7 +5,6 @@
 {
   config,
   lib,
-  pkgs,
   ...
 }:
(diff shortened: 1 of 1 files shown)


proofconstruction:
Can’t think of any reason we would ever need GIS capability in the CVE tracker

RaitoBezarius:
I will let @erictapen approve this to let them know if we missed something.

erictapen (approved):

erictapen:
Yes that was indeed copied wrongly

PR #249, opened 2024-10-02
feat(module): fix wst-manage script

RaitoBezarius:
Make some of the required programs be available in the environment.


fricklerhandwerk (approved):
The [git for every service](https://github.com/Nix-Security-WG/nix-security-tracker/pull/249/files#diff-67d6d7f181f969b45f396a547025c9c6a0972b505c844f72fb5f7497845dac70L172) needs to go away then. I had added it to the wrong place...

PR #258, opened 2024-10-04
fix(cd): install a `nixpkgs` channel

RaitoBezarius:
Otherwise, it's hard to deploy anything.


fricklerhandwerk:
But this is because we're using `nixos-rebuild` instead of calling into a pinned `evalModules` right?

PR #261, opened 2024-10-04
feat(ops/staging): proper secret handling for our dev application

### LINKED ISSUE #227: Staging deployment
As a security tracker developer, I want to see the current state of the `main` branch on some public domain and interacti with the system.

There should always be *some* data to work with, but it's okay if there are no guarantees.

RaitoBezarius:
This introduces all the machinery for our secrets.

This includes a new feature, to do `GH_TOKEN` for our app, we just need to use the private key of our app to emit valid JWT.

Fixes #227.


RaitoBezarius:
This is ready and has been deployed actually.

staging/deploy.sh
```diff
@@ -6,11 +6,21 @@ set -eo pipefail
...
 
-# Perform a dry-activation first.
-echo "dry-activating the configuration first..."
-nixos-rebuild dry-activate -I nixos-config=$DIR/staging/configuration.nix --target-host root@sectracker.nixpkgs.lahfa.xyz
+if [[ "$VERB" != "build" ]]; then
+  # Perform a dry-activation first.
+  echo "dry-activating the configuration first..."
+  nixos-rebuild dry-activate -I nixos-config=$DIR/staging/configuration.nix --target-host root@sectracker.nixpkgs.lahfa.xyz
+else
+  echo "skipping the dry-activation as we are using an offline verb."
```
fricklerhandwerk, line 14: This doesn't make full sense to me... if the default is `switch`, how is this in any way offline? Why wouldn't we dry-activate in that case?

staging/deploy.sh
```diff
@@ -6,11 +6,21 @@ set -eo pipefail
...
+fi
 
-# This requires IPv6 to work as SSH is only IPv6-only.
-# Sorry, not sorry.
-echo "$VERB-ing the configuration now."
-nixos-rebuild $VERB -I nixos-config=$DIR/staging/configuration.nix --target-host root@sectracker.nixpkgs.lahfa.xyz
+
+if [[ "$VERB" != "build" ]]; then
```
fricklerhandwerk, line 18: We're checking this twice now. Typo or why doesn't the work happen in the same conditional?

src/website/shared/utils.py
```diff
@@ -16,16 +19,29 @@ def get_gh(per_page: int = 30) -> Github:
...
+        logger.info(f"Using credentials directory: {credentials_dir}")
 
         try:
             with open(f"{credentials_dir}/GH_TOKEN", encoding="utf-8") as f:
                 gh_auth = Auth.Token(f.read().rstrip("\n"))
-                logger.warn("Using GitHub Token to connect to the API.")
+                logger.info("Using GitHub Token to connect to the API.")
         except FileNotFoundError:
-            logger.warn(
+            logger.debug(
```
fricklerhandwerk, line 31: Why only debug and not warning or at least info?

src/website/shared/utils.py
```diff
@@ -16,16 +19,29 @@ def get_gh(per_page: int = 30) -> Github:
...
+            )
+
+        try:
+            with open(f"{credentials_dir}/GH_APP_PRIVATE_KEY", encoding="utf-8") as f:
+                payload = {
+                    "iat": int(time.time()),
+                    "exp": int(time.time()) + 600,
+                    "iss": get_secret("GH_CLIENT_ID"),
+                }
+                gh_auth = Auth.Token(jwt.encode(payload, f.read(), algorithm="RS256"))
```
fricklerhandwerk, line 42: Nice!

staging/inputs.nix
```diff
@@ -0,0 +1,14 @@
...
+  sources = import ./npins;
+in
+{
+  imports = [
+    "${sources.agenix}/modules/age.nix"
+  ];
+
+  _module.args.inputs = {
+    inherit sources;
+  };
```
fricklerhandwerk, line 13: I think this should just go into `configuration.nix`, I found it needlessly hard to follow.

staging/npins/default.nix
fricklerhandwerk, line 1: I don't quite understand why we have separate pins for the deployment. Is it because we're using `nixos-rebuild` that sources some ambient Nixpkgs? If so, alright, I guess I'll clean it up one of these days. All this here really shouldn't be that convoluted.

staging/inputs.nix
```diff
@@ -0,0 +1,14 @@
+# This gives Flakes vibes for everyone who are missing them.
+{ ... }:
+let
+  sources = import ./npins;
+in
+{
+  imports = [
+    "${sources.agenix}/modules/age.nix"
```
fricklerhandwerk, line 8: I guess we also need age in the dev environment? Or at least proper instructions how to set up your system such that it works. Almost sure I won't be able to work with this without painstakingly figuring it out. Not a blocker for this PR, but at least some lines of documentation with pointers will help me making the cleanup a lot faster.

fricklerhandwerk (approved):
@RaitoBezarius makes roughly sense overall; a bit more clarity in the call graph would be great, but no need to overdo it at this point (overdoing it is *my* job...)

staging/deploy.sh
RaitoBezarius: `build` is an offline verb, we are just building the derivation locally, not sending any closure to the target host.
dry-activate performs network operations.

staging/deploy.sh
RaitoBezarius: Well these are two different phases, the reason for the second conditional is actually a bug in `nixos-rebuild` which performs a non-necessary copy-closure to the target host in presence of that flag.

We could log it as a bug but I really don't want to interact with the nonexistent upstream of these tools as you may know :).

src/website/shared/utils.py
RaitoBezarius: warning is spammy, it's completely valid not to have that file.
It's only when both options are exhausted this becomes a warning.

staging/inputs.nix
RaitoBezarius: That's fine to me.

staging/npins/default.nix
RaitoBezarius: Yes, we are using `nixos-rebuild` and we have separate pins for deployments because it doesn't make sense to use the same pins for development and deployments, BTW.

Staging can go at its pace.

Honestly, it doesn't seem convoluted, it's just combining a bunch of very low level primitives because they're still not first class citizens in Nix default tools.

We can fix that just by moving to colmena as well.

staging/inputs.nix
RaitoBezarius: I can add agenix in the parent development shell or introduce a deployment specific development shell, your pick.

fricklerhandwerk: Let's have it all in one devshell

staging/deploy.sh
fricklerhandwerk: I mean why don't we

```bash
if $VERB == build then
  echo "skipping dry-activate for local build"
  nixos-rebuild build ...
else
  nixos-rebuild dry-activate ...
  nixos-rebuild $VERB --target-host ...
fi
```

staging/npins/default.nix
fricklerhandwerk: > Yes, we are using nixos-rebuild and we have separate pins for deployments because it doesn't make sense to use the same pins for development and deployments, BTW.

I don't understand... shouldn't we have just the same Nixpkgs we work with locally *on the deployed machine*?

staging/deploy.sh
RaitoBezarius: because theoretically we only need to exclude dry-activate, but this is a bug in presence of --target-host
I can implement what you just wrote, but this is just semantically incorrect

staging/npins/default.nix
RaitoBezarius: > > Yes, we are using nixos-rebuild and we have separate pins for deployments because it doesn't make sense to use the same pins for development and deployments, BTW.

> I don't understand... shouldn't we have the same Nixpkgs we work with locally on the deployed machine? Right now we're not even using Nixpkgs in the deployment config AFAICT, and yet we have at least 3 different versions (devshell, channel in the CI action, deployment pin). This seems like asking for trouble, and I have a hunch this will fall apart the next time something shifts around.

> Again, I can clean this up after you, just trying to make sure this is deliberate and trying to understand what the trade-offs are except "let's just bang it out".

Yeah, let me expand. Most people usually do Nix because they can pin things up. We don't care about this.

Actually, we would like to decorrelate the build environment for staging from the build environment for development, the reason for that is that we have different operational constraints.

The devshell needs to accommodate developer needs, we need its nixpkgs to follow not so far from the staging's nixpkgs. Ideally, we should be in the range of breaking changes, e.g. not use APIs that are unavailable in staging → this is caught by the staging CI now.

The staging nixpkgs needs to accomodate operational needs, we need its nixpkgs to be consistently updated to the latest and we need… to track its own CVEs! (which is currently done by the security team efforts automatically, as it is with every NixOS system out there).

The production environment will have similar needs.

The main difference is that we can allow ourselves some customizations in the staging nixpkgs, to apply experimental patches on the top of our dependencies, if we need to, before we promote them.

As soon as we see that this model is not working out because, e.g. staging is on 22.05 and dev is on 25.05 and there's too many API breakages, we can change this and unify things. I see no reason to do it now and there's no cost to manage those different lockfiles as long as we introduce automation to bump them automatically.

But if we do so, I don't want to automatic bumps of the staging environment to interfere with the ongoing development work.

staging/inputs.nix
RaitoBezarius: OK.

staging/deploy.sh
fricklerhandwerk: Please do that or explain in a comment why the code looks differently from what one would expect; whatever you find easier :)

staging/npins/default.nix
fricklerhandwerk: Alright, got it. So the automatic updates go in through the channel set up in CI, and the pins will bring in staging dependencies? And the Nixpkgs for now seems to be a dummy, right?

staging/deploy.sh
RaitoBezarius: ack

staging/npins/default.nix
RaitoBezarius: Okie dokie

PR #264, opened 2024-10-04
feat: orchestrate workflows

RaitoBezarius:
This avoids double building needlessly.

### DIFF
--- a/.github/workflows/deploy.yaml
+++ b/.github/workflows/deploy.yaml
@@ -1,5 +1,11 @@
 name: Deployments
 on:
+  # Run the deployment only once staging has been prepared.
+  workflow_run:
+    workflows: ["Trusted checks"]
+    branches: [main]
+    types:
+      - completed
   push:
     branches: main
 jobs:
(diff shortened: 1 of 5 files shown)


.github/workflows/deploy.yaml
```diff
@@ -1,5 +1,11 @@
 name: Deployments
 on:
+  # Run the deployment only once staging has been prepared.
+  workflow_run:
+    workflows: ["Trusted checks"]
```
fricklerhandwerk, line 5: I got confused by those names, had to read everything again to make sense of it. Can we call them "build" and "dry-activate" so it's clear what they're doing? One check being more "trusted" than the other doesn't convey much meaning.

fricklerhandwerk (approved):
Logic looks good, only needs fixing [`Left Karlton`](https://martinfowler.com/bliki/TwoHardThings.html)

.github/workflows/deploy.yaml
RaitoBezarius: @fricklerhandwerk This is done.

fricklerhandwerk (approved):
Yes, nice, thanks for the extra explanatory comments

PR #265, opened 2024-10-04
feat(mgmt/delta-cve): ingestion of delta catch up based on the next ingestion day

### LINKED ISSUE #228: Catch up with failed ingestions
Currently, if ingesting a day worth of CVEs fails, there is no way to catch up with that.

RaitoBezarius:
Fixes #228.

This fetches everything from [last ingestion day + 1, date] (notice the bracket to mean included/excluded precisely).

In another way, fetches everything in ]last ingestion day, date].

### DIFF
--- a/src/website/shared/management/commands/ingest_delta_cve.py
+++ b/src/website/shared/management/commands/ingest_delta_cve.py
@@ -11,6 +11,7 @@ import requests
 from django.core.management.base import BaseCommand, CommandError
 from django.db import transaction
 from github import UnknownObjectException
+from github.Repository import Repository
 from shared import models
 from shared.fetchers import make_cve
 from shared.models import CveIngestion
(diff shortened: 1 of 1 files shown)


src/website/shared/management/commands/ingest_delta_cve.py
```diff
@@ -19,11 +20,72 @@
...
+    return CveIngestion.objects.create(valid_to=day, delta=True)
+
+
 class Command(BaseCommand):
     help = "Ingest CVEs in bulk using the Mitre CVE repo"
 
     def add_arguments(self, parser: argparse.ArgumentParser) -> None:
-        parser.add_argument("date", help="Date of the delta to download.")
+        parser.add_argument(
+            "date", help="End date until which we will download all the missing deltas."
```
fricklerhandwerk, line 87: Having a type check here would help a lot with both using and using and reading the code.

fricklerhandwerk (approved):
Looks good

src/website/shared/management/commands/ingest_delta_cve.py
RaitoBezarius: This is done.

PR #269, opened 2024-10-07
feat: make the delta code more resilient to upstream data issues

RaitoBezarius:
This fixes temporary issues by upstream.

### DIFF
--- a/src/website/shared/management/commands/ingest_delta_cve.py
+++ b/src/website/shared/management/commands/ingest_delta_cve.py
@@ -30,12 +30,13 @@ def ingest_day(repo: Repository, day: datetime.datetime) -> CveIngestion:
     logger.info(f"Fetched release: {release.title}")
 
     # Get the bulk cve list asset
+    if not release.assets:
+        raise CommandError(f"End of day release for {day} has no asset attached.")
+
     bundle = release.assets[0]
 
     if bundle.name != f"{day}_delta_CVEs_at_end_of_day.zip":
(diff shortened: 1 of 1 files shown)


src/website/shared/management/commands/ingest_delta_cve.py
```diff
@@ -126,4 +128,9 @@ def handle(self, *args: Any, **kwargs: Any) -> None:
             next_ingestion + datetime.timedelta(days=x)
             for x in range((date - next_ingestion).days + 1)
         ):
-            ingest_day(repo, day)
+            try:
+                ingest_day(repo, day)
+            except UnknownObjectException:
```
fricklerhandwerk, line 133: This is a thing that GitHub can throw, and we're catching that exception type in `ingest_day`. Shouldn't we instead catch *any* exception and print that rather than swallowing those details?

fricklerhandwerk (changes requested):
Yeah right, not barfing in the middle of the loop would be smart. Just not sure about the exact handling.

src/website/shared/management/commands/ingest_delta_cve.py
RaitoBezarius: If there's more fatal errors, there's no need to continue the loop, the script should terminate early and this is abnormal.

If there's a _known_ non-problematic error, we should catch it and decide whether to continue or not. Here, we decide that if GitHub has not a release on a single day, no big deal, we shall continue.

fricklerhandwerk (approved):

PR #272, opened 2024-10-08
feat: extract severity from metrics

### LINKED ISSUE #112: Extract severity from the `metrics` JSON object
Required for https://github.com/Nix-Security-WG/nix-local-security-scanner/issues/51

In CVEs the data could look something like this:
```json
            "metrics": [
                {
                    "other": {
                        "content": {
                            "text": "low"
                        },
                        "type": "Textual description of severity"
                    }
                },
                {
                    "cvssV3_1": {
                        "version": "3.1",
                        "vectorString": "CVSS:3.1/AV:N/AC:L/PR:L/UI:N/S:C/C:L/I:L/A:N",
                        "baseScore": 6.4,
                        "baseSeverity": "MEDIUM"
                    }
                }
            ],

```

but at the time of writing our data model for that is

```python
class Metric(models.Model):
    """Class representing an impact information related to a CVE record."""

    format = models.CharField(max_length=64)
    scenarios = models.ManyToManyField(Description)
    content = models.JSONField()


class Container(models.Model):
    # ...
    metrics = models.ManyToManyField(Metric)
```

RaitoBezarius:
Fixes #112 and some chores as well.


src/website/shared/models/cve.py
```diff
@@ -153,9 +153,96 @@ class Impact(models.Model):
...
+            _("ADJACENT_NETWORK"),
+        )
+        NETWORK = ("NETWORK", _("NETWORK"))
+
+    class Severity(models.TextChoices):
+        NONE = ("NONE", _("NONE"))
+        LOW = ("LOW", _("LOW"))
+        MEDIUM = ("MEDIUM", _("MEDIUM"))
+        HIGH = ("HIGH", _("HIGH"))
+        CRITICAL = ("CRITICAL", _("CRITICAL"))
```
fricklerhandwerk, line 183: This looks extremely tedious to me. Why can't we

```python
class ChoiceEnum(Enum):
    @classmethod
    def choices(cls):
        return [(choice.name, _(str(choice.value))) for choice in cls]

    def __str__(self):
        return self.name

class Scopes(ChoiceEnum):
    UNCHANGED = "UNCHANGED"
    CHANGED = "CHANGED"

class AttackVectors(ChoiceEnum):
    PHYSICAL = "PHYSICAL"
    LOCAL = "LOCAL"
    ADJACENT_NETWORK = "ADJACENT_NETWORK"
    NETWORK = "NETWORK"

class Severity(ChoiceEnum):
    NONE = "NONE"
    LOW = "LOW"
    MEDIUM = "MEDIUM"
    HIGH = "HIGH"
    CRITICAL = "CRITICAL"
```

And also I'm not sure I fully understand why we need a `TextChoice` and can't do `IntegerChoice` protobuf-style. I guess we'd have to do a mapping on ingestion, but that seems low-overhead.

src/website/shared/fetchers.py
```diff
@@ -87,10 +87,43 @@ def make_problem_type(data: dict[str, Any]) -> models.ProblemType:
...
 def make_metric(data: dict[str, Any]) -> models.Metric:
     ctx: dict[str, Any] = dict()
     ctx["format"] = "cvssV3_1"
-    ctx["content"] = data.get("cvssV3_1", {})
+    raw_cvss = data.get("cvssV3_1", {})
+    ctx["raw_cvss_json"] = raw_cvss
+
+    if raw_cvss:
+        ctx["scope"] = raw_cvss.get("scope")
+        ctx["vector_string"] = raw_cvss.get("vectorString")
```
fricklerhandwerk, line 109: Why do we handle these two separately? Currently they could be read in the loop with like the others.

src/website/shared/models/cve.py
RaitoBezarius: Your proposal does not support built-in gettext translation. And these strings _should_ be translated.

src/website/shared/fetchers.py
RaitoBezarius: `vectorString` actually serves to generate the next fields as well. `scope` could be threaded through in the same way but these fields have different "statuses" in the specification.

src/website/shared/models/cve.py
fricklerhandwerk: Doesn't `return [(choice.name, _(str(choice.value))) for choice in cls]` cut it?

src/website/shared/fetchers.py
fricklerhandwerk: `attackVector` also has different choices, this is why I'm asking

src/website/shared/models/cve.py
RaitoBezarius: This is not statically discoverable.

RaitoBezarius: Tools needs to mark the callsites for `_("")`

src/website/shared/fetchers.py
RaitoBezarius: `attackVector` is derivable from `vectorString` though. This is the common point of all `vector_fields`, they are *fields* from the `vectorString`, it's not really the case of `baseScore`, etc.

fricklerhandwerk: Ah, now I also understand what the TODO means. You want to check that the separately stored fields actually correspond what's in the `vectorString`.

fricklerhandwerk (approved):

Issue #274, opened 2024-10-08
Improve accuracy and offer rationale of matchings


RaitoBezarius:
We need to:

- improve matchings wrt to version constraint checks and more.

We also need to provide:

- rationale of matchings so that they can be highlighted for #190.

Issue #189, opened 2024-09-25
Compare CVE version constraints with Nixpkgs package versions


erictapen:
As a security team member, when reviewing the version constraints in a match suggestion, I want to be able to asses wether the constraints are indicating a security issue. The CVE can offer *version constraints*, while the selected Nixpkgs packages offer a set of versions, one for every channel and package.

- **No usable version constraints** should indicate a need for manual review
- An **existing version constraint** that **matches** with one of our versions needs high visibility
- An **existing version constraint** that **doesn't match** needs low visiblility

Specifically for Nixpkgs package versions,

- if **stable** and **unstable** version matches, we summarise
- if **stable** and **unstable** don't match, we list separately
- if the version in one channel differ (e.g. nixpkgs-unstable and nixos-unstable-small), we show these differences

Depends on:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/334

fricklerhandwerk:
Closed by https://github.com/Nix-Security-WG/nix-security-tracker/pull/365

Issue #203, opened 2024-09-26
View a list of pre-computed match suggestions


fricklerhandwerk:
As a security team member, I want to have an overview of untriaged CVEs, displayed as a priority list.

Each item in that list should only show the most relevant information:
  - CVE ID
  - CVE logline
  - Number of derivations affected
  - Number of channels affected
  
 The list should be sorted by some reasonable combination of
 - CVE age
 - CVE severity
 - Relevance for Nixpkgs (e.g. number of supposedly affected derivations)
 - Confidence in the matching of CVE and Nixpkgs metadata
 
 Depends on:
 - #221

erictapen:
@fricklerhandwerk @RaitoBezarius and me discussed, that we will keep the concept of a match suggestion separate from a Nix security issue. That applies both to the internal model and how it's communicated to the user.

erictapen:
![tmp yRLhZ2GsEM](https://github.com/user-attachments/assets/bc6fa432-de1f-4543-bc26-776bba181028)
![tmp 2jPHitPcr2](https://github.com/user-attachments/assets/118f79fe-d131-4e9b-9ab3-9c727fd44f09)

The wireframes we want to work with for now. From now I'll focus on implementation, to see how this works with real data.

erictapen:
Provenance of how the match was created could be represented in some background colors of the relevant fields
https://github.com/Nix-Security-WG/nix-security-tracker/pull/282

PR #282, opened 2024-10-14
feat: introduce provenance for derivation candidates

RaitoBezarius:
In the future, we will be able to bolt on more fields to take notes
which embedding has been used, which search term has been used, etc,
etc.

For now, we have a provenance flags we can use to hint the frontend what
has been used.

The dance for the M2M is required because it's not trivial to migrate
everything.

Progress towards:

- #189
- #274


erictapen:
Sounds promising to have. I'd think of ways how to visualise this in https://github.com/Nix-Security-WG/nix-security-tracker/issues/203

RaitoBezarius:
> Sounds promising to have. I'd think of ways how to visualise this in #203

Yeah, this is only progress towards that goal, there's more to come to help you with hints :)

erictapen (approved):
The content of ProvenanceFlags looks sensible to me, even though I'm not sure how I would use the hardware and kernel constraints.

RaitoBezarius:
> The content of ProvenanceFlags looks sensible to me, even though I'm not sure how I would use the hardware and kernel constraints.

Showing "Linux" / "macOS" / etc, I'd say?

src/website/shared/listeners/automatic_linkage.py
```diff
@@ -10,19 +11,36 @@
...
 
-def produce_linkage_candidates(container: Container) -> set[NixDerivation]:
+class ProvenanceFlags(IntFlag):
+    PACKAGE_NAME_MATCH = auto()
+    VERSION_CONSTRAINT_INRANGE = auto()
+    VERSION_CONSTRAINT_OUTOFRANGE = auto()
+    NO_SOURCE_VERSION_CONSTRAINT = auto()
+    # Whether the hardware constraint is matched for this derivation.
+    HARDWARE_CONSTRAINT_INRANGE = auto()
+    KERNEL_CONSTRAINT_INRANGE = auto()
```
fricklerhandwerk, line 21: Why do we have that type twice, and the other is strict?

RaitoBezarius: Fixed.

Issue #210, opened 2024-09-27
View a CVE description


fricklerhandwerk:
When triaging a suggestion or editing a security record draft, I want to inspect the complete CVE description to understand what the issue is about.

Depends on:
- #203
- dependency (and its direction) to discuss: https://github.com/Nix-Security-WG/nix-security-tracker/issues/201

PR #293, opened 2024-10-28
feat: Suggestion view pt.2

erictapen:
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/203
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/210

Some better rendering for package versions.

### DIFF
--- a/src/website/webview/views.py
+++ b/src/website/webview/views.py
@@ -47,6 +47,9 @@ from shared.models import (
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
 )
+from shared.models.nix_evaluation import (
+    channel_structure,
+)
 
 from webview.forms import NixpkgsIssueForm
 from webview.paginators import CustomCountPaginator
@@ -488,23 +491,8 @@ class SuggestionListView(ListView):
     def get_context_data(self, **kwargs: Any) -> Any:
         context = super().get_context_data(**kwargs)
 
-        # Major channels are the important channels that a user wants to keep an eye on.
-        # FIXME make it dynamic
-        major_channels = ["nixos-23.11", "nixos-24.05", "nixos-24.11", "nixos-unstable"]
-
         for obj in context["object_list"]:
-            obj.packages = dict()
-            for derivation in obj.derivations.all():
-                attribute = derivation.attribute.removesuffix(f".{derivation.system}")
-                channel = derivation.parent_evaluation.channel.channel_branch
-                # FIXME This is wrong. Replace with something like builtins.parseDrvName
-                version = derivation.name.split("-")[-1]
-                if attribute not in obj.packages:
-                    obj.packages[attribute] = dict()
-                obj.packages[attribute][channel] = {
-                    "version": version,
-                    "major": channel in major_channels,
-                }
+            obj.packages = channel_structure(obj.derivations.all())
         return context
 
     def get_queryset(self) -> Any:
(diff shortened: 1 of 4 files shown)


src/website/webview/views.py
```diff
@@ -490,22 +490,57 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
+        major_channels = ["23.11", "24.05", "24.11", "unstable"]
+
+        def get_major_channel(branch_name: str) -> str | None:
+            for mc in major_channels:
+                if mc in branch_name:
+                    return f"nixos-{mc}"
+            return None
 
         for obj in context["objects"]:
+            # TODO Validate assumption that there is only one container
```
RaitoBezarius, line 502: It's a wrong assumption, there are multiple containers, but we shall only select the most relevant one

src/website/webview/views.py
```diff
@@ -490,22 +490,57 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
+                major_channel = get_major_channel(branch_name)
                 if attribute not in obj.packages:
                     obj.packages[attribute] = dict()
-                obj.packages[attribute][channel] = {
-                    "version": version,
-                    "major": channel in major_channels,
-                }
+                # FIXME This quietly drops unfamiliar branch names
+                if major_channel:
+                    if major_channel not in obj.packages[attribute]:
```
RaitoBezarius, line 515: Can the channel analyzing logic be moved in its own functions and possibly a utility file for channel analysis?

erictapen: Definetly, this entire logic is way too noisy for the view.

src/website/webview/views.py
erictapen: Any idea how to figure out the most relevant one?

src/website/webview/views.py
```diff
@@ -490,22 +490,57 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
 
         for obj in context["objects"]:
+            # TODO Validate assumption that there is only one container
             obj.cve_container = obj.cve.container_set.all().first()
             obj.packages = dict()
             for derivation in obj.derivations.all():
                 attribute = derivation.attribute.removesuffix(f".{derivation.system}")
-                channel = derivation.parent_evaluation.channel.channel_branch
+                branch_name = derivation.parent_evaluation.channel.channel_branch
                 # FIXME This is wrong. Replace with something like builtins.parseDrvName
```
erictapen, line 508: https://github.com/Nix-Security-WG/nix-security-tracker/issues/294

src/website/webview/views.py
```diff
@@ -490,22 +490,57 @@ def get_context_data(self, **kwargs: Any) -> Any:
 
         # Major channels are the important channels that a user wants to keep an eye on.
         # FIXME make it dynamic
```
erictapen, line 492: https://github.com/Nix-Security-WG/nix-security-tracker/issues/295

src/website/webview/views.py
RaitoBezarius: Hmm… I need to think, but basically you need to select the first container containing a package name or containing a CVSS vector string, this sort of stuff.

Maybe, the backend should do more preprocessing to make this palatable.

src/website/webview/views.py
erictapen: This is probably wrong but I ended up moving this into the models for now? I thought it might make sense, as the function operates on the types in the `nix_evaluation` model.

src/website/webview/views.py
RaitoBezarius: correct!

fricklerhandwerk (approved):

PR #299, opened 2024-10-30
fix: increase timeout on the container

fricklerhandwerk:
since migrations may be quite expensive when the datase state diverges
enough, the container will get killed before it's ready even if
everything is alright. the logs are quite opaque in that case,
because, well, there's no failure.

### DIFF
--- a/nix/web-security-tracker.nix
+++ b/nix/web-security-tracker.nix
@@ -200,6 +200,7 @@ in
           requires = [ "postgresql.service" ];
           wantedBy = [ "multi-user.target" ];
           serviceConfig.Restart = cfg.restart;
+          serviceConfig.TimeoutStartSec = "10m";
           preStart = ''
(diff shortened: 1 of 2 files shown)


RaitoBezarius:
should be put in the general module, not only this one

nix/web-security-tracker.nix
```diff
@@ -200,6 +200,7 @@ in
           requires = [ "postgresql.service" ];
           wantedBy = [ "multi-user.target" ];
           serviceConfig.Restart = cfg.restart;
+          serviceConfig.TimeoutStartSec = "10m";
```
RaitoBezarius, line 203: mkDefault it

fricklerhandwerk:
```suggestion
          serviceConfig.TimeoutStartSec = lib.mkDefault "10m";
```

RaitoBezarius (approved):

Issue #279, opened 2024-10-11
Password policy for the staging deployment


fricklerhandwerk:
Since we're publishing the dumps, admin passwords will be present there in hashed form. Document that fact and show a warning on the relevant pages. Check if we can disable the overriding of Django password policy enforcement.

PR #308, opened 2024-10-31
feat: display a security disclaimer when being run in debug mode to warn staging users

erictapen:
Afair we don't really track wether the software runs as staging. I figured just determining wether it's running in debug mode would be enough as a heuristic.

Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/279

tmp 2KliF3G2EB


fricklerhandwerk (approved):
Good enough

Issue #206, opened 2024-09-26
Accept a match suggestion


fricklerhandwerk:
As a security team member, I want to be able to create a draft security issue from a suggested match.

Depends on:
- #203

Issue #207, opened 2024-09-26
Dismiss a match suggestion


fricklerhandwerk:
As a security team member, I want to be able to mark a suggested match as invalid.

Depends on:
- #203

PR #309, opened 2024-10-31
feat: add buttons to add another package, dismiss and create a draft

erictapen:
Very quick layout for the buttons themselfes, that doesn't do anything yet.

Addresses, but not closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/206 and https://github.com/Nix-Security-WG/nix-security-tracker/issues/207.

### DIFF
--- a/src/website/webview/static/style.css
+++ b/src/website/webview/static/style.css
@@ -380,7 +380,6 @@ article .nixpkgs-packages {
 }
 
 .nixpkgs-package h3 {
-  flex: 1;
   margin: unset;
 }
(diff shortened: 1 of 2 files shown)


fricklerhandwerk (approved):
I removed the 'add package' thing for now, because https://github.com/Nix-Security-WG/nix-security-tracker/issues/213 is a bit more of an effort to get right. I also made the dummy buttons show zalgo text: we shouldn't put things on there that don't work yet, as will only lead to disappointment.

image

PR #310, opened 2024-10-31
ci: pin Nixpkgs for deployments

fricklerhandwerk:
this doesn't rely on NIX_PATH and is therefore fully pure, and uses our
own, controlled dependencies.

the Nixpkgs for deployment currently not decoupled from the one for
development because the build is broken on the newest version.

we still need a policy for how to run updates, but ideally we'd use
exactly one version of Nixpkgs for everything.

### DIFF
--- a/.github/workflows/builds.yaml
+++ b/.github/workflows/builds.yaml
@@ -17,7 +17,5 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v29
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable-small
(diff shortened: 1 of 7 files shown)


RaitoBezarius (approved):

RaitoBezarius:
This doesn't pass CI though.

fricklerhandwerk:
Nothing passes CI at the moment because we can't build the server against unstable.

fricklerhandwerk:
The dry activations check is expected to fail because it runs in the base branch's context (it's has a `pull_request_target` trigger). It should unbreak once this change is merged.

PR #311, opened 2024-10-31
ci: auto-update pins for staging

fricklerhandwerk:
we can extend this easily to also update the development pins. I suggest we keep them in sync, so there's no room for divergence, because breaking builds will hold up everything anyway, so we may just fix them for all outputs at once.

### DIFF
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -7,6 +7,10 @@ on:
   # Allow manual trigger
   workflow_dispatch:
 
+env:
+  ACTIONS_RUNNER_DEBUG: true
+  ACTIONS_STEP_DEBUG: true
+
 jobs:
   update-npins:
     runs-on: ubuntu-latest
--- a/.github/workflows/bump.yaml
+++ b/.github/workflows/bump.yaml
@@ -13,12 +13,9 @@ jobs:
     steps:
       - uses: actions/checkout@v4
       - uses: cachix/install-nix-action@v29
-        with:
-          nix_path: nixpkgs=channel:nixos-unstable-small
       - uses: DeterminateSystems/magic-nix-cache-action@main
       - run: |
-          cd ./staging
-          nix-shell -p npins --run "npins update"
+          nix-shell --run "npins -d ./staging update"
       - uses: peter-evans/create-pull-request@v7
         with:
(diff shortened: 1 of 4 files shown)


RaitoBezarius (approved):
LGTM.

fricklerhandwerk:
Sorry for the horrendous amount of notification spam. This should work now, with all the org and repo permission set correctly. The last test run of the trigger only stopped at the missing `workflow_dispatch` on `builds.yaml`, which needs to be in the PR branch itself: https://github.com/Nix-Security-WG/nix-security-tracker/pull/316#issuecomment-2451127815

.github/workflows/bump.yaml
```diff
@@ -0,0 +1,32 @@
+name: Weekly npins update
+
+on:
+  schedule:
+    # run at 00:00 UTC every Monday
+    - cron: "0 0 * * 1"
+  # allow manual trigger
+  workflow_dispatch:
+
+permissions:
```
fricklerhandwerk, line 10:
```suggestion
# NOTE: for this to work you need to set "Workflow permissions" to "Read and write permissions"
# and check "Allow GitHub Actions to create and approve pull requests" in
# https://github.com/organizations/<owner>/settings/actions
# in order enable the checkbox "Allow GitHub Actions to create and approve pull requests" in
#  https://github.com/<owner>/<repo>/settings/actions
# where you only need to set "Read repository contents and packages permissions"
#
# otherwise you'll get an opaque 403 on the pull-request workflow
permissions:
```

.github/workflows/bump.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - run: |
+          nix-shell default.nix -A ci --run "npins -d ./staging/npins update"
+      - uses: peter-evans/create-pull-request@v7
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "chore: npins update"
+          title: "chore: weekly npins update"
+          body: |
+            Automatic npins update performed by GitHub Actions
+          branch: npins-auto-update
```
fricklerhandwerk, line 41:
```suggestion
          # NOTE: in this repo we disabled creating additional branches.
          # in order for this to work, the rules must have an exception for this special branch.
          branch: npins-auto-update
```

PR #318, opened 2024-11-01
feat: provide basic pagination in suggestions view

erictapen:
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/291


fricklerhandwerk (approved):
It broke the header (fixed it), otherwise great!

PR #319, opened 2024-11-01
feat: actually change suggestion state in suggestions view

erictapen:
very hastily done

Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/206
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/207

### DIFF
--- a/src/website/webview/views.py
+++ b/src/website/webview/views.py
@@ -31,7 +31,7 @@ from django.db.models.functions import Coalesce
 from django.db.models.manager import BaseManager
 from django.db.models.query import QuerySet
 from django.http import HttpRequest, HttpResponse, HttpResponseRedirect
-from django.shortcuts import get_object_or_404
+from django.shortcuts import get_object_or_404, redirect
 from django.utils.decorators import method_decorator
 from django.views.generic import DetailView, ListView, TemplateView
 from shared.models import (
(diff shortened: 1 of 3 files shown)


src/website/webview/views.py
```diff
@@ -546,3 +547,14 @@ class CastToText(Func):
             description=F("cve__container__descriptions__value"),
         )
         return queryset
+
+    def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+        suggestion_id = request.POST.get("suggestion_id")
+        new_status = request.POST.get("new_status")
+        suggestion = get_object_or_404(CVEDerivationClusterProposal, id=suggestion_id)
+        if new_status == "REJECTED":
```
fricklerhandwerk, line 555: We may want to automatically map it to the enum and catch the exception.

@RaitoBezarius we'd soon also need the permission mapping to take effect here. Shall I open an issue for that?

erictapen: Is there a way to cast a string to an enum? I was under the impression I'd need to write that myself anyways.

fricklerhandwerk (approved):
Nice, this works. Did a few cosmetic fixups.

src/website/webview/views.py
fricklerhandwerk: @RaitoBezarius:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/324
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/325

@erictapen:
```python
from enum import Enum

class Color(Enum):
    RED = 1
    GREEN = 2
    BLUE = 3

print(Color["RED"])
print(Color["SOMETHING"])
 ```
 
 ```console
 > python3 test.py
Color.RED
Traceback (most recent call last):
  File "/home/vg/src/nix-security-tracker/test.py", line 9, in <module>
    print(Color["SOMETHING"])
          ~~~~~^^^^^^^^^^^^^
  File "/nix/store/h3i0acpmr8mrjx07519xxmidv8mpax4y-python3-3.12.5/lib/python3.12/enum.py", line 814, in __getitem__
    return cls._member_map_[name]
           ~~~~~~~~~~~~~~~~^^^^^^
KeyError: 'SOMETHING'
```

Issue #204, opened 2024-09-26
Display previously dismissed suggestions


fricklerhandwerk:
As a security team member, I want to revisit suggestions that have been dismissed in the past.

Depends on:
- if we want to arrive there by filtering: https://github.com/nix-security-wg/nix-security-tracker/issues/200
- https://github.com/nix-security-wg/nix-security-tracker/issues/207

Issue #208, opened 2024-09-26
View a list of already triaged suggestions


fricklerhandwerk:
As a security team member, I want to have an overview of all pre-triaged issues, so I can pick one of them for refinement.

Depends on:
- #206

fricklerhandwerk:
Closed by https://github.com/Nix-Security-WG/nix-security-tracker/pull/328

PR #328, opened 2024-11-06
feat: show dismissed suggestions and draft issues

fricklerhandwerk:
closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/204 #208

### DIFF
--- a/src/website/webview/templates/draft_list.html
+++ b/src/website/webview/templates/draft_list.html
@@ -35,19 +35,3 @@
   {% endfor %}
 </div>
 {% endblock content %}
-
-{% block pagination %}
-  <ul>
-    {% for page_number in adjusted_elided_page_range %}
-      <li class="index">
-        {% if page_number == paginator.ELLIPSIS %}
-        {{page_number}}
-        {% else %}
-        <a href="{{request.path}}?page={{page_number}}" class="{% if page_number == page_obj.number %}current{% endif %}">
-          {{page_number}}
-        </a>
-        {% endif %}
-      </li>
-    {% endfor %}
-  </ul>
-{% endblock %}
--- a/src/website/webview/templates/draft_list.html
+++ b/src/website/webview/templates/draft_list.html
@@ -1,7 +1,13 @@
 {% extends "base.html" %}
 
 {% block content %}
-<h1>Drafts</h1>
+<h1>Selected suggestions</h1>
+<p>
+<span class="hint green-button">Create draft</span> to convert the suggestion into a draft security issue that can be edited before publishing. <s><a href="/drafts">View all drafts</a></s>
+</p>
+<p>
+<span class="hint red-button">Dismiss</span> to remove a suggestion from the queue. <a href="/dismissed">View all dismissed suggestions</a>
+</p>
 <div id="suggestion-list">
   {% for object in object_list %}
     <article class="suggestion">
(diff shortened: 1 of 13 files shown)


src/website/webview/templates/draft_list.html
```diff
@@ -0,0 +1,102 @@
...
+          {% elif object.base_severity == "NONE" %}
+            No severity data
+          {% else %}
+            Unknown Severity
+          {% endif %}
+        </div>
+      </div>
+      <h2>{{ object.package_name }}</h2>
+      <!-- XXX: just another idea how to do it -->
+      <details class="description">
```
erictapen, line 30: Yeah I don't really agree with this. Semantically I don't think that the title is just a summary of the description and the unfolding element makes the layout more janky and wastes space on mobile.

tmp Y9Ji6RTXal

src/website/webview/views.py
```diff
@@ -561,3 +531,108 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
...
+    # def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
+    # suggestion_id = request.POST.get("suggestion_id")
+    # new_status = request.POST.get("new_status")
+    # suggestion = get_object_or_404(CVEDerivationClusterProposal, id=suggestion_id)
+    # if new_status == "REJECTED":
+    # suggestion.status = CVEDerivationClusterProposal.Status.REJECTED
+    # elif new_status == "ACCEPTED":
+    # suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
+    # suggestion.save()
+    # return redirect("/suggestions")
```
erictapen, line 638: Something needs to be done with this.

erictapen (commented):
This introduces a lot of duplicated template code. Maybe it's time to move stuff into individual files for easier reusability?

src/website/webview/templates/draft_list.html
fricklerhandwerk: Reading a bunch of them, it does look like the title is a reasonable summary of the more detailed description, this is why I tried this out in the first place. We can figure out the word wrapping. The details tag has two advantages:
- no JS needed
- doesn't jump around when toggling state. the `show more` thing appears somewhere else depending on collapsed state

fricklerhandwerk:
Yes, we should definitely extract the template bits for a suggestion singleton and a suggested derivation.

src/website/webview/templates/draft_list.html
fricklerhandwerk: Here's an alternative rendering:

image

fricklerhandwerk: What actually can't be done without JS is moving the "Show less..." after the contents and still keep it interactive. But again, that would mean things would jump around.

I really don't care much whether it's the classic triangle or a word, but I slightly prefer the triangle because it has well-established semantics that don't require the user to parse text.

fricklerhandwerk: Here's what it's like with a triange:

image

erictapen: deal!

fricklerhandwerk:
image

erictapen:
After a series of refactoring PRs, I just merged main into this PR again. This should have now less duplicated code, even though the `views.py` file still looks pretty messy...

As discussed with @fricklerhandwerk we may eventually move towards one view for suggestions, dismissed and selected anyways, so maybe this is fine for now?

PR #340, opened 2024-11-11
fix: link derivation metadata during ingestion

fricklerhandwerk:
didn't test it yet but it should be right. someone needs to run an evaluation


erictapen:
From first glance, just running an evaluation didn't produce the wanted result:

```console
# systemctl start web-security-tracker-fetch-all-channels.service

# sudo -su web-security-tracker psql -c "select metadata_id from shared_nixderivation where metadata_id != null;"
 metadata_id 
-------------
(0 rows)
```

Issue #338, opened 2024-11-11
NixDerivation has `metadata` always set to `null` currently


erictapen:
We have lots of available metadata in `shared_nixderivationmeta` but none of it seems to be linked to a derivation:

```console
$ psql
psql (15.8)
Type "help" for help.

web-security-tracker=> select metadata_id from shared_nixderivation where metadata_id != null;
 metadata_id 
-------------
(0 rows)
```

Issue #333, opened 2024-11-11
View derivation description in suggestion view


fricklerhandwerk:
As a security team member, while triaging I want to know what a derivation is about, and for that I need to at least see the short description string right in the suggestion.

Blocked by:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/359 (we'd need a useful data dump to iterate on the UI)

erictapen:
Blocked by https://github.com/Nix-Security-WG/nix-security-tracker/issues/338

fricklerhandwerk:
#388 is closed now, unblocked.

fricklerhandwerk:
@RaitoBezarius @Erethon Blocked by https://github.com/Nix-Security-WG/nix-security-tracker/issues/359 and https://github.com/Nix-Security-WG/nix-security-tracker/issues/198 since we also don't have a working eval availabe on staging, and we're using that to pull fresh DB dumps. @erictapen  and I can work around this by spending time to set up a local eval, but it would be a better use of everyone's time if we didn't and could focus on UX tasks.

PR #352, opened 2024-11-14
feat: display meta.description for package in suggestion list

erictapen:
Old code that I should have pushed long ago.

Addresses https://github.com/Nix-Security-WG/nix-security-tracker/issues/333 but doesn't close it yet, as it is blocked by #338.

I needed to introduce the flexbox again, but I think it is fine, as the versions block has a fixed width? Shouldn't be too hard to make it look right.

tmp 9UPmnPjbS4

This change is invisible actually, as there is no data to display yet.

### DIFF
--- a/src/website/shared/models/nix_evaluation.py
+++ b/src/website/shared/models/nix_evaluation.py
@@ -341,36 +341,46 @@ def channel_structure(derivations: list[NixDerivation]) -> dict:
         # FIXME This is wrong. Replace with something like builtins.parseDrvName
         version = derivation.name.split("-")[-1]
         if attribute not in packages:
-            packages[attribute] = dict()
+            packages[attribute] = {
+                "versions": {},
+            }
+            if derivation.metadata and derivation.metadata.description:
+                packages[attribute]["description"] = derivation.metadata.description
         branch_name = derivation.parent_evaluation.channel.channel_branch
         major_channel = get_major_channel(branch_name)
         # FIXME This quietly drops unfamiliar branch names
         if major_channel:
-            if major_channel not in packages[attribute]:
-                packages[attribute][major_channel] = {
+            if major_channel not in packages[attribute]["versions"]:
+                packages[attribute]["versions"][major_channel] = {
                     "major_version": None,
                     "uniform_versions": None,
                     "sub_branches": dict(),
                 }
             if not branch_name == major_channel:
-                packages[attribute][major_channel]["sub_branches"][branch_name] = (
+                packages[attribute]["versions"][major_channel]["sub_branches"][
+                    branch_name
+                ] = version
+            else:
+                packages[attribute]["versions"][major_channel]["major_version"] = (
                     version
                 )
-            else:
-                packages[attribute][major_channel]["major_version"] = version
-    for package in packages:
-        for mc in packages[package].keys():
+    for package_name in packages:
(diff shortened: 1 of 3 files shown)


src/website/shared/models/nix_evaluation.py
```diff
@@ -341,36 +341,46 @@ def channel_structure(derivations: list[NixDerivation]) -> dict:
         # FIXME This is wrong. Replace with something like builtins.parseDrvName
         version = derivation.name.split("-")[-1]
         if attribute not in packages:
-            packages[attribute] = dict()
+            packages[attribute] = {
+                "versions": {},
```
fricklerhandwerk, line 345: Nit, maybe it should be `nixpkgs-releases` instead of versions, for clarity?

erictapen: When I hear nixpkgs release I think of 24.04 or unstable, but this is naming the data structure that contains package versions, grouped by nixpkgs releases.

src/website/webview/templates/components/nixpkgs_package_list.html
```diff
@@ -1,29 +1,34 @@
...
-              {% endfor %}
-              </ul>
-            </details>
-          </li>
-        {% endfor %}
-      </ul>
+      <div class="package-meta">
+        <div class="package-description">
+          {{ pdata.description }}
+        </div>
```
fricklerhandwerk, line 8:
```suggestion
        <p class="package-description">
          {{ pdata.description }}
        </p>
```

No reason to make it a generic container, this is clearly phrasing content.

src/website/webview/templates/components/nixpkgs_package_list.html
```diff
@@ -1,29 +1,34 @@
...
+        <ul class="channel-list">
+          {% for major_channel, val in pdata.versions %}
+            <li>
+              <details {% if val.uniform_versions %}open{% endif %}>
+                <summary>
+                  <span class="branch-major">
+                    <span class="branch-name">{{ major_channel }}</span>
+                    <span class="version">{% if val.major_version %}{{ val.major_version }}{% else %}???{% endif %}</span>
+                  </span>
+                </summary>
```
fricklerhandwerk, line 18:
```suggestion
                <summary class="branch-major">
                    <span class="branch-name">{{ major_channel }}</span>
                    <span class="version">{% if val.major_version %}{{ val.major_version }}{% else %}???{% endif %}</span>
                </summary>
```

May want to reduce the number of moving parts here. Since we keep piling them up, we need to keep an eye on noise growing beyond being manageable.

fricklerhandwerk (approved):
LGTM, some superficial notes

src/website/shared/models/nix_evaluation.py
fricklerhandwerk: Ah then I misread, sorry.

src/website/webview/templates/components/nixpkgs_package_list.html
erictapen: Yeah this was necessary, as otherwise the black triangle disappears for some reason.

erictapen: I reverted for now.

fricklerhandwerk: Ah, could be because of the `display: flex`.

Issue #323, opened 2024-11-01
Remove derivations from an issue suggestion


fricklerhandwerk:
As a security team member, already while triaging I want to remove irrelevant packages from a suggestion.

PR #353, opened 2024-11-14
feat: allow for package deselection on the suggestions view

erictapen:
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/323

Needed some refactoring, so we don't show the checkbox in the other views

There is lots more to do, but this should provide basic functionality.

tmp 0MsFuAB1xS

### DIFF
--- a/src/website/shared/models/nix_evaluation.py
+++ b/src/website/shared/models/nix_evaluation.py
@@ -343,9 +343,11 @@ def channel_structure(derivations: list[NixDerivation]) -> dict:
         if attribute not in packages:
             packages[attribute] = {
                 "versions": {},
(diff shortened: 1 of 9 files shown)


fricklerhandwerk:
We discussed collapsing the items with CSS and having a custom checkbox label, right?

Also we could actually show the same UI in the other views but, there, add a button to update an entry instead of just changing state. That would be a cheap stand-in for dynamic htmx stuff.

fricklerhandwerk:
image
image

fricklerhandwerk (approved):

erictapen:
Turns out there are duplicated ids for the input elements, I'm investigating

PR #354, opened 2024-11-14
fix(evaluation): link metadata to derivations

### LINKED ISSUE #338: NixDerivation has `metadata` always set to `null` currently

RaitoBezarius:
This was disabled for performance reasons (probably?!), re-enabled.

This was tested to work locally.

Fixes #338.

### DIFF
--- a/src/website/shared/evaluation.py
+++ b/src/website/shared/evaluation.py
@@ -426,7 +426,7 @@ class SyncBatchAttributeIngester:
                 metadatas.append(metadata)
 
             derivations[evaluation.drv_path] = self.ingest_derivation_shell(
-                evaluation, parent_evaluation, None
+                evaluation, parent_evaluation, metadata
             )
         logger.debug(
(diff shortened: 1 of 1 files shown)


src/website/shared/evaluation.py
```diff
@@ -426,7 +426,7 @@ def ingest(self, parent_evaluation: NixEvaluation) -> list[NixDerivation]:
                 metadatas.append(metadata)
 
             derivations[evaluation.drv_path] = self.ingest_derivation_shell(
-                evaluation, parent_evaluation, None
+                evaluation, parent_evaluation, metadata
```
fricklerhandwerk, line 429: Oof, slightly hidden knob there

fricklerhandwerk (approved):

PR #364, opened 2024-11-15
fix: proper ellipsis in suggestions

erictapen:
I just couldn't not do it :see_no_evil:


fricklerhandwerk (approved):
You did the right thing.

PR #365, opened 2024-11-15
feat: display CPE identifier, especially affected package name and version constraints

### LINKED ISSUE #334: Display version constraints for a CVE in the suggestion view
As a security team member I want to see at a glance which versions of suggested packages are affected (assuming the suggestions are relevant). Currently we'd have to click through the upstream CVE description.

Version constraints are part of an CPE identifier, which should be shown in full. That means the same workflow would allow for seeing which affected software information (e.g. package name) the CVE provides.

erictapen:
Closes #334 #189 

As discussed with @fricklerhandwerk, this also moves the affected package name into the bottom and makes the CVE title the primary heading.

For this we can also drop the arbitrary requirement, that all displayed suggestions need to have an affected package. The reason behind this filter was, to show more interesting data for demo purposes.

Turns out, there are *a lot* of suggestions without an affected package. Maybe deduplication helps with this, but otherwise we have to think of something to properly demo this and maybe even filter out the noise.

tmp H3toigalpg


src/website/webview/templates/suggestion_list.html
```diff
@@ -24,11 +24,15 @@ <h1>Automatically generated suggestions</h1>
           {% severity_badge object.base_severity %}
         </div>
 
-        <h2>{{ object.package_name }}</h2>
-
         <details class="description">
           <summary>
-              {% if object.title %}{{ object.title }}{% else %}{{ object.description|truncatewords:10 }}{% endif %}
+            <h2>
+              {% if object.title %}
```
fricklerhandwerk, line 30: Is the title really optional? I left the condition while refactoring because it was there originally, but I don't actually know the data quality here.

src/website/webview/templates/suggestion_list.html
```diff
@@ -24,15 +24,21 @@ <h1>Automatically generated suggestions</h1>
           {% severity_badge object.base_severity %}
         </div>
 
-        <h2>{{ object.package_name }}</h2>
-
         <details class="description">
           <summary>
-              {% if object.title %}{{ object.title }}{% else %}{{ object.description|truncatewords:10 }}{% endif %}
+            <h2>
```
fricklerhandwerk, line 29: No need to make it a section or wrap it at all
```suggestion
```

src/website/webview/views.py
```diff
@@ -492,8 +492,18 @@ class SuggestionListView(ListView):
     def get_context_data(self, **kwargs: Any) -> Any:
         context = super().get_context_data(**kwargs)
 
+        affected_pk = list()
```
fricklerhandwerk, line 495: With https://github.com/Nix-Security-WG/nix-security-tracker/pull/369 you can now do

```suggestion
        affected_pk = list()
        metrics_pk = list()
        
        for obj in context["object_list"]:
            metrics_pk.extend(obj.cve.container.values_list("metrics", flat=True))
        metrics = Metric.objects.in_bulk(metrics_pk)
```

The only difference we have to figure out is whether we want *all* the `baseSeverity`s from the list of `Metric`s or e.g. a "max" over the enum.

src/website/webview/templates/suggestion_list.html
erictapen: Currently we need this to distinguish from the "Show more" styling. And it *is* the title, so might have some benefit for accessibility tools?

fricklerhandwerk: > Currently we need this to distinguish from the "Show more" styling

That's not true, the `:after` content can be styled completely separately.

> And it is the title, so might have some benefit for accessibility tools?

The schema field is called "title" but in reality it's a one-sentence summary. `<summary>` is the correct semantics here (takes of pointed hat and goes back to work)

erictapen:
I'm finally starting to see some versions, even though they look really weird.

tmp XiJq6UUitf

erictapen:
tmp yzHE9uwAfP

erictapen:
I'd try to get this merge like this for now and iterate further in future PRs.

tmp HVR3r4It7X

src/website/webview/views.py
```diff
@@ -500,8 +501,40 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
+                            obj.affected_packages[a.package_name] = {
+                                "version_constraints": set(),
+                                "cpes": set(),
+                            }
+                        obj.affected_packages[a.package_name][
+                            "version_constraints"
+                        ].update(
+                            [
+                                (vc.status, vc.version_constraint_str())
+                                for vc in a.all_versions
```
erictapen, line 531: Using `a.all_versions` like this runs fine, but produces a pyright error, as it apparently doesn't understand how django adds fields via prefetch:

```
/home/kerstin/git/nix-security-tracker/src/website/webview/views.py                                                                    
  /home/kerstin/git/nix-security-tracker/src/website/webview/views.py:531:45 - error: Cannot access attribute "all_versions" for class 
"AffectedProduct"                                                                                                                      
    Attribute "all_versions" is unknown (reportAttributeAccessIssue)
  /home/kerstin/git/nix-security-tracker/src/website/webview/views.py:535:52 - error: Cannot access attribute "all_cpes" for class "Aff
ectedProduct"
    Attribute "all_cpes" is unknown (reportAttributeAccessIssue)
2 errors, 0 warnings, 0 informations
```

I've had this problem in the past, but forgot how/wether I solved it. Just not using `to_attr` doesn't help, as then I can't iterate on the field at runtime.

erictapen: Figured this out. The trick is to not use `to_attr` and then use `.all()` on the field.

erictapen:
Converting into draft again, until I fully incorporated https://github.com/Nix-Security-WG/nix-security-tracker/pull/378

erictapen:
Actually on second thought, let's try to get this merged and optimise later.

src/website/webview/templates/suggestion_list.html
erictapen: Done

fricklerhandwerk:
Usually a migration would be required, but we currently have zero deployments, so I suggest skipping it this time.

PR #366, opened 2024-11-17
Suggestions activity log

alejandrosame:
The activity log is implemented with the package 'django-pghistory', which uses PostgreSQL triggers.
    
    - Added python package 'django-pghistory'.
    - Updated 'django-pgtrigger' version, a dependency of the former. The old
      version was conflicting.


src/website/webview/views.py
```diff
@@ -482,6 +486,65 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
+    return get_username(user_id)
+
+
+def get_suggestion_activity_log(suggestion: CVEDerivationClusterProposal) -> list[str]:
+    log = []
+
+    # Suggestion creation, deletion and status updates
+    for event in CVEDerivationClusterProposalStatusEvent.objects.prefetch_related(
+        "pgh_context",
+    ).filter(pgh_obj_id=suggestion.pk):
```
RaitoBezarius, line 526: `.all().iterator()` should be preferred for iteration

src/website/webview/views.py
```diff
@@ -547,6 +612,8 @@ def get_context_data(self, **kwargs: Any) -> Any:
 
         for obj in context["object_list"]:
             obj.packages = channel_structure(obj.derivations.all())
+            obj.activity_log = get_suggestion_activity_log(obj)
```
RaitoBezarius, line 615: this is inefficient, you should have `get_suggestion_activity_logs(<list of object>)` and returns a map of `{pk: logs}` and then interleave them there.

src/website/webview/views.py
```diff
@@ -482,6 +486,65 @@ def get_context_data(self, **kwargs: Any) -> Any:
...
+
+
+def get_suggestion_activity_log(suggestion: CVEDerivationClusterProposal) -> list[str]:
+    log = []
+
+    # Suggestion creation, deletion and status updates
+    for event in CVEDerivationClusterProposalStatusEvent.objects.prefetch_related(
+        "pgh_context",
+    ).filter(pgh_obj_id=suggestion.pk):
+        user = get_user_from_context(event.pgh_context)
```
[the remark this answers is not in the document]
alejandrosame: For clarification, my guess is that you are referring to making it part of the ORM expression and not a RAW SQL call, right?

alejandrosame:
@RaitoBezarius I'm still massaging the data for the view. I'll take into account these query feedback on the last refactors before taking this PR out of draft.

fricklerhandwerk:
Tested, looking good. A few things currently on my mind:
- We wanted to show relative times first, so that still needs a conversion
- We only need to show only the creation timestamp in the suggestions view (there can't be any activity anyway)
- Showing the creation timestamp should always be possible (even before the logs were enabled), it's recorded with the suggestion in the `TimeStampMixin`
- The logs take 750ms per item on my machine
  - @RaitoBezarius says there could be some quick wins, so now it's time to harvest those
  - Eventually we probably want to not load the logs for the entire list, but only on demand
    - It should be enough to show last-modified in the `<summary>`
    - Last-modified can also be recorded in the `TimeStampMixin`, but the needs to be overridden to keep only the state changes, because IIUC it's not doing anything right now for suggestions

erictapen:
Happy to be able to look at it though! I'd keep design feedback until it's marked as ready for review.
Just one hint: With https://github.com/Nix-Security-WG/nix-security-tracker/pull/374 a color palette was introduced, that could be used for the coloring of state change activities.

src/website/webview/views.py
alejandrosame: From https://github.com/Nix-Security-WG/nix-security-tracker/pull/366#issuecomment-2487949573
>Eventually we probably want to not load the logs for the entire list, but only on demand 

@fricklerhandwerk Is this the case? Because in that case I'd skip this optimization, since it's the current logic is already tailored to "gather the activity log of this suggestion".

src/website/webview/views.py
alejandrosame: Implemented in https://github.com/Nix-Security-WG/nix-security-tracker/pull/366/commits/f11423ac1ddcbd944cbe4f9bbc60b02714c6728b

src/website/webview/views.py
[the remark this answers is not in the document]
alejandrosame: Implemented in https://github.com/Nix-Security-WG/nix-security-tracker/pull/366/commits/323d3f3318e8dd33aebec67c89ec19ebdab1c4d6

fricklerhandwerk:
I suppose this supersedes #126 and #101?

alejandrosame:
@fricklerhandwerk True, they should be closed and #126 can be used as reference whenever there's a decision to have an activity log for issues.

Issue #335, opened 2024-11-11
Separately list suggestions that have no matches


fricklerhandwerk:
As a security team member, I want to quickly clean out the triage queue from irrelevant suggestions. In order to do that, I want to separately go through suggestions which don't have any associated packages -- which typically means they're not affecting us -- and dismiss them in bulk.

Depends on:
- #187

PR #371, opened 2024-11-18
feat: only produce suggestions that have matches

fricklerhandwerk:
> NOT TESTED

This may be an alternative fix for https://github.com/Nix-Security-WG/nix-security-tracker/issues/335. Because, do we really need empty suggestions to begin with?

### DIFF
--- a/src/website/shared/listeners/automatic_linkage.py
+++ b/src/website/shared/listeners/automatic_linkage.py
@@ -55,24 +55,29 @@ def build_new_links(container: Container) -> None:
         return
 
     drvs = produce_linkage_candidates(container)
-    proposal = CVEDerivationClusterProposal.objects.create(cve=container.cve)
 
-    drvs_throughs = [
-        CVEDerivationClusterProposal.derivations.through(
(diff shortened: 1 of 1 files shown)


erictapen:
We briefly discussed this in the daily and I think it might be sensible to do this.

If we go for this, it might make sense to do it soon, as it might influence the design process for the suggestions view; Showing CVEs with loads of irrelevant CPEs would be less of a problem then.

src/website/shared/listeners/automatic_linkage.py
```diff
@@ -55,25 +55,30 @@ def build_new_links(container: Container) -> None:
         return
 
     drvs = produce_linkage_candidates(container)
-    proposal = CVEDerivationClusterProposal.objects.create(cve=container.cve)
```
RaitoBezarius, line 58: please, replace the flow by:

```
if not drvs: return
```

and keep the rest as-is, do not introduce and `if` for `drvs_through`, that's useless

fricklerhandwerk:
Superseded by #378

PR #374, opened 2024-11-19
feat: introduce a basic color palette for suggestion, dismissed, selected and draft

erictapen:
No time at all was spent on finding the colors here, so feel free to change in the future. This is only about consistency

tmp usEqLyTFUL
tmp 3BGnWV5GFq

### DIFF
--- a/src/website/webview/static/style.css
+++ b/src/website/webview/static/style.css
@@ -228,6 +228,10 @@
 :root {
   --grey: #d5d2d1;
   --dark-grey: #b5b2b1;
+  --automatic-suggestion: lightblue;
+  --dismissed-suggestion: #acb3c8;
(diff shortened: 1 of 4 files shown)


src/website/webview/static/style.css
```diff
@@ -533,14 +533,20 @@ span.hint {
...
+.draft-color {
+  background: var(--draft);
   color: white;
   cursor: pointer;
 }
 
-.green-button:hover {
-  background: darkgreen;
+button:hover {
+  filter: brightness(85%);
```
fricklerhandwerk, line 549: This is clever to abstract away, but will probably break down eventually. Let's leave it at that, but it will likely need follow-up work.

fricklerhandwerk (approved):

PR #378, opened 2024-11-20
feat: cache the suggestions in a JSON payload

RaitoBezarius:
which can be pulled back, paginated, etc.

This will run automatically upon every new insertion of pending proposal.


src/website/shared/listeners/cache_suggestions.py
```diff
@@ -0,0 +1,104 @@
...
+            "descriptions__value",
+        )
+        .all()
+    )
+
+    relevant_piece = [x for x in relevant_data if "affected__package_name" in x]
+    if not relevant_piece:
+        # No package name.
+        return
+    relevant_piece = relevant_piece[0]
```
fricklerhandwerk, line 56: @erictapen I guess we're already using more than just the first random package name (at least in some PR)?

src/website/shared/listeners/cache_suggestions.py
```diff
@@ -0,0 +1,104 @@
...
+
+    # This is not a suggestion we want to show.
+    if suggestion.derivations.count() > 1_000:
```
fricklerhandwerk, line 59: More like

```suggestion
    if suggestion.derivations.count() > 100:
```

fricklerhandwerk (approved):
Nice, this does the trick

PR #383, opened 2024-11-25
feat: use htmx to dynamically apply status changes

### LINKED ISSUE #379: Dynamic element loading for state changes
I'd like to replace https://github.com/Nix-Security-WG/nix-security-tracker/issues/321 with this.

As as security member, I want to change states of suggestions (automatic, dismissed, selected) with as little friction as possible. It is going to happen, that in my workflow I will dismissed hundreds of suggestions by clicking on it.

Therefore, the time until the next interaction is possible shouldn't be noticable. Currently, we have a page reload here which takes quite some time.

By not reloading the entire page via POST, but just swapping the suggestion tile with something that makes clear that its state got changed (and maybe even allows for a quick undo of a wrong action) we can make this much more of a pleasure to go through.

Also I think I understand HTMX sufficiently now that it wouldn't be so hard to accomplish this.

erictapen:
Closes #379 

Opening this for now so there is code to show. Still missing in scope:

- Make the undo action work
- Make a js-free workflow work again


erictapen:
tmp WySjCUm2MW

This is pretty much finished now, *but* depends on #365, as otherwise the merge conflicts would have meant twice the work.

Will mark as ready for review once #365 is merged.

erictapen:
Fixed the unwanted jump behaviour!

fricklerhandwerk:
@erictapen clicking "Purge deleted packages" in the *selected* view collapses a suggestion

erictapen:
@fricklerhandwerk Can we leave this for now and make the Purge button obsolete later by applying the same htmx logic to removing packages?

fricklerhandwerk:
Yeah should be fine.

fricklerhandwerk:
No, I'll do a last pass ASAP and merge.

src/website/webview/static/htmx.min.js
fricklerhandwerk, line 1: We do have Nix though, no need to vendor that.

fricklerhandwerk (approved):
LGTM except the shocking lack of Nixing

src/website/webview/templates/components/suggestion.html
```diff
@@ -4,35 +4,36 @@
...
 
     <details class="description">
       <summary>
-        {% if suggestion.title %}
-          {{ suggestion.title }}
+        {% if cached_suggestion.title %}
+          {{ cached_suggestion.title }}
         {% else %}
-          {{ suggestion.description|truncatewords:10 }}
+          {{ scached_uggestion.description|truncatewords:10 }}
```
fricklerhandwerk, line 25:
```suggestion
          {{ cached_suggestion.description|truncatewords:10 }}
```

Typo?

fricklerhandwerk (approved):
Awesome!

PR #384, opened 2024-11-27
fix: refresh suggestion cache entry on derivations update

erictapen:
Couldn't sleep with having this problem in my head.

I had to switch the arguments for the update_or_create function because they were wrong: The proposal_id is used for the lookup and the payload is then used on update.

This is going to make updates very slow, but doesn't affect page load and I'd argue that it is important to have correct behaviour at first. With https://github.com/Nix-Security-WG/nix-security-tracker/pull/383 merged update times wouldn't even be noticable.

### DIFF
--- a/src/website/webview/views.py
+++ b/src/website/webview/views.py
@@ -20,6 +20,7 @@ from django.contrib.postgres.search import (
     SearchRank,
 )
 from django.core.paginator import Page
+from django.db import transaction
 from django.db.models import (
     Case,
     Count,
@@ -37,6 +38,7 @@ from django.shortcuts import get_object_or_404, redirect
 from django.template.loader import render_to_string
 from django.utils.decorators import method_decorator
 from django.views.generic import DetailView, ListView, TemplateView
+from shared.listeners.cache_suggestions import cache_new_suggestions
 from shared.models import (
     AffectedProduct,
     Container,
@@ -530,7 +532,9 @@ class SuggestionListView(ListView):
         elif new_status == "ACCEPTED":
             suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
 
-        suggestion.save()
+        with transaction.atomic():
+            suggestion.save()
+            cache_new_suggestions(suggestion)
 
         # We provide graceful fallback for important workflows, when users have JavaScript disabled
         if "no-js" in request.POST:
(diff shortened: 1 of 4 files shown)


src/website/webview/views.py
```diff
@@ -545,7 +547,10 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         elif new_status == "ACCEPTED":
             suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
 
-        suggestion.save()
+        with transaction.atomic():
+            suggestion.save()
+            cache_new_suggestions(suggestion)
```
fricklerhandwerk, line 552: This should be done triggered async, otherwise we'll have to wait a long time.

RaitoBezarius:
I looked into that more seriously, I don't think triggering a recaching is the good approach here.

For one, this will trigger a weird experience where you will see temporarily (in case of heavy loads) the view non-sync with what's going on.

Recaching should be reserved for expensive transformations that are hard to re-express on their cached variants.

In this instance, there's only one operation: filtering out derivations from the payload.

This should be expressible in one SQL query which should not involve back'n'forth with Python. I have some local prototype I'm working on where I can do this sort of stuff.

fricklerhandwerk:
Are you suggesting something like storing a flag for the filtered-out derivations, but somewhere outside the actual suggestion entry (since that is cached)?

RaitoBezarius:
This is now ready.

RaitoBezarius:
Rebased.

RaitoBezarius:
The rebase broke displaying derivations.

RaitoBezarius:
OK this is beyond my paygrade, I do see the derivations threaded in some templates but I don't understand what's going on, can you weigh in @erictapen (whenever you are available ofc)?

RaitoBezarius:
In the meantime, fix #396 as intended in principle.

fricklerhandwerk:
This also breaks state change display, setting it to draft for now.

erictapen:
Fixed!

fricklerhandwerk (approved):
Tested, works. Displaying pages in the "selected" view for some reason is much slower than in the "suggestions" view, which doesn't make much sense to me, but we'll deal with that another time.

Issue #267, opened 2024-10-05
Delta catch up should optimize for bulk when the gap is too large


RaitoBezarius:
Delta CVE should error if the gap is too large.

We should also create a new toplevel management command called `ingest_cve` which does either bulk ingestion or delta ingestion based on the situation. Old commands can be kept for more advanced users.

Bulk ingestion should probably also skip existing CVEs.

This way, our CVE ingestion process becomes quite rock-solid.

PR #388, opened 2024-11-28
feat: the delta CVE ingester can replace the bulk CVE ingester

RaitoBezarius:
It supports parallelism but `make_cve` does not.

This will result in many deadlocks and issues during the transaction if
it's run with more than one process.

### DIFF
--- a/nix/web-security-tracker.nix
+++ b/nix/web-security-tracker.nix
@@ -16,6 +16,7 @@ let
     mkDefault
     concatStringsSep
     recursiveUpdate
+    optionalString
     ;
   inherit (pkgs) writeScriptBin writeShellApplication stdenv;
   cfg = config.services.web-security-tracker;
@@ -104,6 +105,18 @@ in
     secrets = mkOption {
       type = types.attrsOf types.path;
       default = { };
+    };
+    cve.startDate = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      defaultText = "the application default: January 1st of the prior year";
+      description = ''
+        The ingestion start date for CVE, most operators will care about CVEs of their last year until now.
+        Hence, this is the default.
+
+        If you need to obtain older CVEs for any reason, change this value.
+      '';
+      example = "2024-11-01";
     };
     maxJobProcessors = mkOption {
       description = ''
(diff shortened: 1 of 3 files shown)


src/website/shared/management/commands/ingest_delta_cve.py
```diff
@@ -64,40 +72,100 @@ def ingest_day(repo: Repository, day: datetime.datetime) -> CveIngestion:
...
-    help = "Ingest CVEs in bulk using the Mitre CVE repo"
+    help = "Ingest CVEs day per day using the MITRE CVE repo"
 
     def add_arguments(self, parser: argparse.ArgumentParser) -> None:
         parser.add_argument(
             "date",
             type=datetime.date.fromisoformat,
             help="End date until which we will download all the missing deltas.",
         )
+        # Always start from the past year's January.
```
fricklerhandwerk, line 146: That's evident from the code, but not *why*, please reword the comment to tell that.

src/website/shared/management/commands/ingest_delta_cve.py
```diff
@@ -64,40 +72,100 @@ def ingest_day(repo: Repository, day: datetime.datetime) -> CveIngestion:
...
+        logger.exception(
+            f"CVE ingestion on day {day} is impossible as there's no release, continuing for the next days"
+        )
+
+
+def parallel_ingestion(
+    repo: Repository,
+    next_ingestion: datetime.datetime,
+    date: datetime.datetime,
+    num_processes: int = 4,
```
fricklerhandwerk, line 110: Why do we have a default here, especially if it's unsafe with >1 as is? The CLI already sets a default

nix/web-security-tracker.nix
```diff
@@ -105,6 +106,12 @@ in
       type = types.attrsOf types.path;
       default = { };
     };
+    cve.startDate = mkOption {
+      type = types.nullOr types.str;
+      default = null;
+      defaultText = "the application default: last current year";
```
fricklerhandwerk, line 112:
```suggestion
      defaultText = "the application default: January 1st of the prior year";
```

Regarding "why", maybe the option's `description` the right place to put it because ideally this is the only place where anyone would ever have to manually interact with it.

fricklerhandwerk (approved):
A few maintenance-oriented questions, LGTM morally otherwise.

fricklerhandwerk:
Closes #267

RaitoBezarius:
@fricklerhandwerk ready

PR #389, opened 2024-11-28
fix(suggestion-activity-log): cover empty updates case

alejandrosame:


src/website/webview/templatetags/viewutils.py
```diff
@@ -62,14 +62,20 @@ def iso(date: datetime.datetime) -> str:
...
+        return next(reversed(od))
+    except StopIteration:
+        return None
 
 
 @register.filter
-def last_user(od: OrderedDict) -> str:
-    _, entry = next(reversed(od.items()))
-    return entry[0]["user"]
+def last_user(od: OrderedDict) -> str | None:
```
fricklerhandwerk, line 73: How will this `None` show in the template?

alejandrosame: image

alejandrosame: Mind that I never got into this case before because in the current implementation the calls for `last_user` and `last_key` were guarded by `{% if suggestion.activity_log.updates|length_is:0 %}`. So the previous screenshot shows an artificial check. Whenever the list of updates is empty the case of `created X` is the one to be shown.

fricklerhandwerk: Understood. Will merge it as is and @erictapen can figure out what to display in this case when we get into the situation. Right now the priority is not to throw an exception, we'll have to iron out the visual kinks later.

PR #393, opened 2024-11-29
fix: Add setuptools to dependencies

Erethon:
Sorry if I'm missing something obvious here, I discovered I need this change in order to build this on a minimal NixOS installation.


RaitoBezarius:
I think this is a matter of migrating the old codebase to the newest ways to package it, it's not the right solution for this.

Erethon:
Closing then, I'll discuss it with fricklerhandwerk to figure out what's needed.

PR #394, opened 2024-11-29
feat(shared/apps): run github sync state in production

### LINKED ISSUE #239: GitHub sync doesn't run in production
[Deployment through the module uses `daphne`](https://github.com/Nix-Security-WG/nix-security-tracker/blob/main/nix/web-security-tracker.nix#L217-L218), while the initial sync is [only run when doing `manage runserver`](https://github.com/Nix-Security-WG/nix-security-tracker/blob/main/src/website/shared/apps.py#L15).

Make sure deployment always syncs on startup and sync can be ran manually in development mode.

RaitoBezarius:
We just create a bespoke environment variable set by our packaging.

There's no other way as we cannot depend on any specific WSGI/ASGI
server argv0 here.

Fixes #239.

Depends on #392.

### DIFF
--- a/src/website/shared/apps.py
+++ b/src/website/shared/apps.py
@@ -13,6 +13,7 @@ class SharedConfig(AppConfig):
 
         # This hook is called on any `manage` subcommand.
         # Only connect to GitHub when the server is started.
+        # TODO: run this as a separate service, as this is almost exclusively a deployment concern
         if os.environ.get("RUN_MAIN", None) is None and (
             "runserver" in sys.argv
             or os.environ.get("SYNC_GITHUB_STATE_AT_STARTUP", False)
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
What happened to the idea of running sync as a separate cronjob? Too many moving parts?

src/website/shared/apps.py
```diff
@@ -13,7 +13,10 @@ def ready(self) -> None:
 
         # This hook is called on any `manage` subcommand.
         # Only connect to GitHub when the server is started.
-        if os.environ.get("RUN_MAIN", None) is None and "runserver" in sys.argv:
+        if os.environ.get("RUN_MAIN", None) is None and (
```
fricklerhandwerk, line 16:
```suggestion
        # TODO: run this as a separate service, as this is almost exclusively a deployment concern
        if os.environ.get("RUN_MAIN", None) is None and (
```

PR #395, opened 2024-11-29
fix: render package description in a better way

### LINKED ISSUE #333: View derivation description in suggestion view

erictapen:
Closes #333 

Good enough, will be refined with #209 anyways.
tmp co2OdPp0za

### DIFF
--- a/src/website/webview/static/style.css
+++ b/src/website/webview/static/style.css
@@ -494,6 +494,8 @@ article .nixpkgs-packages {
 
 .nixpkgs-package .package-description {
   flex-grow: 1;
+  font-family: "IBM Plex Sans";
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
For #209 we'll have to look into how to arrange the description so the view (especially on mobile) isn't too stuffed and scales with displaying more or less details, but that will clarify itself once we start considering all the other derivation metadata.

Issue #398, opened 2024-11-29
Set up GitHub parameters and secrets for staging


fricklerhandwerk:
The staging deployment currently doesn't start because it's not wired up with GitHub.

Follow-up on https://github.com/Nix-Security-WG/nix-security-tracker/issues/239

alejandrosame:
Fixed with https://github.com/Nix-Security-WG/nix-security-tracker/pull/460 and https://github.com/Nix-Security-WG/nix-security-tracker/pull/402.

PR #401, opened 2024-11-29
Revert "feat(shared/apps): run github sync state in production"

fricklerhandwerk:
Reverts Nix-Security-WG/nix-security-tracker#394

This breaks staging deployment without https://github.com/Nix-Security-WG/nix-security-tracker/issues/398. @alejandrosame please revert the revert once the GitHub setup is in place.


RaitoBezarius:
can you prepare the revert of this revert? so we can remerge once it's ready

Issue #390, opened 2024-11-29
Nixpkgs evaluation not robust enough


fricklerhandwerk:
Analsys by @RaitoBezarius:

There can be situations where a Nixpkgs evaluation is in progress, but gets cancel by a deployment or some other interruption, and there's no recovery of pending work. From there, further channel updates don't get processed and the evaluation system is deadlocked.

fricklerhandwerk:
The most critical items have been fixed, and the rest is not really actionable right now as we need monitoring. #345

PR #405, opened 2024-11-29
fix(evaluation): crash pending evaluations at worker restart

RaitoBezarius:
This way, it's not possible to have leftovers not taken care until the next worker restart.

Now, "pending" evaluations can happen only in-between two worker restarts.
For this, we can add a supervisor timer unit that will check if there's any evaluation taking far too long (> 24 hours) and will forcibly restart the worker.

With that, the loop is closed.

Counts towards #390.

### DIFF
--- a/nix/web-security-tracker.nix
+++ b/nix/web-security-tracker.nix
@@ -236,6 +236,7 @@ in
 
           script = ''
             # Before starting, crash all the in-progress evaluations.
+            # This will prevent them from being stalled forever, since workers would not pick up evaluations marked as in-progress.
             wst-manage crash_all_evaluations
             wst-manage listen --processes ${toString cfg.maxJobProcessors}
           '';
--- a/nix/web-security-tracker.nix
+++ b/nix/web-security-tracker.nix
@@ -235,6 +235,8 @@ in
           wantedBy = [ "multi-user.target" ];
 
           script = ''
+            # Before starting, crash all the in-progress evaluations.
+            wst-manage crash_all_evaluations
             wst-manage listen --processes ${toString cfg.maxJobProcessors}
           '';
         };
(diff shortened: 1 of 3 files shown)


RaitoBezarius:
This breaks the possibility for scaling the workers across multiple machines as we may crash ongoing evaluations which are done on another machine, there's no other way around. The good way to fix this is to move the evaluation in its own microservice and with its own handling.

nix/web-security-tracker.nix
```diff
@@ -235,6 +235,8 @@ in
           wantedBy = [ "multi-user.target" ];
 
           script = ''
+            # Before starting, crash all the in-progress evaluations.
```
fricklerhandwerk, line 238:
```suggestion
            # Before starting, crash all the in-progress evaluations.
            # This will prevent them from being stalled forever, since workers would not pick up evaluations marked as in-progress.
```

src/website/shared/management/commands/crash_all_evaluations.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Command(BaseCommand):
+    help = "Transition all in-progress evaluations to the crashed state"
+
+    def add_arguments(self, parser: ArgumentParser) -> None:
+        pass
+
+    def handle(self, *args: Any, **kwargs: Any) -> str | None:
+        # Look for all evaluations with a created_at time bigger than 24 hours.
```
fricklerhandwerk, line 15:
```suggestion
```

This does not apply here but is set in the channel update cron service config

src/website/shared/models/nix_evaluation.py
```diff
@@ -237,7 +237,7 @@ class EvaluationState(models.TextChoices):
             "IN_PROGRESS",
             _("In progress"),
         )
-        # Crash means resource exhaustion
+        # Crash means resource exhaustion or unexpected crash of the worker
```
fricklerhandwerk, line 240:
```suggestion
        # Crash means resource exhaustion, unexpected crash, or forced shutdown of the worker
```

fricklerhandwerk (approved):

PR #414, opened 2024-12-04
Sort suggestions by update

### LINKED ISSUE #356: Sort dismissed and selected views by last update
When triaging matching suggestions and inspecting dismisses/selected entries, I want to see those items I most recently moved around at the top of the list.

(Currently they are sorted in seemingly unpredictable order.)

alejandrosame:
The summary of changes is:
  - Sync suggestion `updated_at` with event tables for tracked status and derivations changes using signals.
  - Order the suggestion list based on latest updated first, and latest created second.
  - Place timestamp mixins in a separate module.

Closes #356


src/website/webview/views.py
```diff
@@ -530,7 +530,7 @@ def post(self, request: HttpRequest, *args: Any, **kwargs: Any) -> HttpResponse:
         elif new_status == "ACCEPTED":
             suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
 
-        suggestion.save()
+        suggestion.save(update_fields=["status"])
```
RaitoBezarius, line 533: What does this change has in relation with this commit?

src/website/shared/models/timestamps.py
```diff
@@ -0,0 +1,25 @@
+"""Timestamp mixins
```
RaitoBezarius, line 1: Arguably, it seems overkill to me to collect a file only of timestamp mixins, I'd recommend renaming this to mixins and making this a file of all models mixins we care about.

src/website/shared/models/linkage.py
```diff
@@ -68,3 +72,55 @@ class DerivationClusterProposalLink(models.Model):
...
+                DerivationClusterProposalLinkEvent.objects.filter(
+                    proposal_id=instance.pk
+                )
+                .order_by("-pgh_created_at")
+                .values_list("pgh_created_at", flat=True)[0]
+            )
+
+    if instance.updated_at is None or last_inserted_timestamp > instance.updated_at:  # type: ignore
+        instance.updated_at = last_inserted_timestamp
+        instance.save()
```
RaitoBezarius, line 126: Minimally, this should be protected against infinite looping by asserting that this save won't trigger further signal receives.

src/website/shared/models/linkage.py
```diff
@@ -68,3 +72,55 @@ class DerivationClusterProposalLink(models.Model):
...
+    """Keep proposal's `updated_at` field in sync with last relevant update.
+
+    From this function we can import `DerivationClusterProposalLinkEvent`
+    without import errors. When we try to import it at the module level,
+    there is an error about `shared.models` being partially initialized.
+    This is happening because the `Event` models get dynamically added by
+    the `@pghistory.track` decorators. The alternatives to having the import
+    in the function are:
+        - Creating the Event model explicitly.
+        - Using `instance.derivations.through.pgh_event_model` where
```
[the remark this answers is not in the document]
alejandrosame: > Furthermore, below you mention arguments about hardcoding table names,

For the record, below I mention about hardcoding table names in a Postgres trigger, which goes into a migration, and would need to be rewritten with the new table name. Here, the tradeoff is hardcoding the Model name, not the table name, to which I'm just leaving the documented default. No migrations are involved with the import.

src/website/shared/models/linkage.py
alejandrosame: This is what is achieved by having the check on "status" in "updated_fields".

src/website/shared/models/linkage.py
```diff
@@ -68,3 +72,55 @@ class DerivationClusterProposalLink(models.Model):
...
+          `DerivationClusterProposalLinkEvent` is used. However, that
+          expression is less readable.
+
+    Keeping `updated_at` in sync would be done more safely with Postgres triggers.
+    However, introducing them for this case would involve hardcoding table names.
```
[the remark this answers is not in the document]
alejandrosame: This is not well tested nor properly documented (for example how to reference other tables, which is needed for this denormalization). If I were to introduce the trigger table I'd simply do a normal interpolated Python string to introduce the trigger in the migration.

src/website/shared/models/timestamps.py
alejandrosame: I can rename the file for this PR. However I won't move other mixins here in this PR as that would increase even more the testing surface due to the general lack of automated tests.

The whole purpose of moving the timestamps together was to increase the visibility of the comments that hints at strategies to manage timestamps in Django according to the desired effect.

src/website/shared/models/linkage.py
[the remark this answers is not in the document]
alejandrosame: > I'm not sure also if the data consistency will be severely trashed in situations where this is seriously used,

According to the issue, this field is a denormalization of the actual timestamp registered with pghistory. Given this comment (talking about "severely trashed", "seriously used") makes me thing that there's a potential disagreement about what `updated_at` should mean.

fricklerhandwerk:
Important: We need to sort only by status update, not by metadata update (such as changed derivations), see #356! Because if we remove derivations in the "selected" view, we don't want the item to jump to the top on next page load. So really this should be a very simple change, AFAICS.

alejandrosame:
Superseded by #429. The signal vs trigger aspect of this PR was meant to enable further refactors in the business logic.

I've already worked around the fact that the timestamps don't agree from the beginning and they serve different purposes.

PR #423, opened 2024-12-05
feat: instant package removal

### LINKED ISSUE #411: Remove derivations instantly
Make it work the same as removing suggestions.

erictapen:
Depends on #422 as I refactored some stuff there.

Closes #411

### DIFF
--- a/src/website/webview/templates/components/suggestion.html
+++ b/src/website/webview/templates/components/suggestion.html
@@ -61,6 +61,14 @@
         <button type="submit" name="create-draft-not-implemented" class="draft-color">
           C̶̖͆r̶͍̚e̵͈̐ă̷̖t̷̄͜e̶͓̓ ̴̟̏d̵̳̅r̸̮̐a̵͍͑f̵̙̋t̴̹̃
         </button>
+        {% if cached_suggestion.packages %}
+          <div class="remove-when-js-is-enabled button-with-note">
+          <button type="submit" class="grey-button">
+            Purge deleted packages
+          </button>
+          <span class="note">(cannot be undone)<span>
+          </div>
+        {% endif %}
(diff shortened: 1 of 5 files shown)


fricklerhandwerk:
Draft until #422 is sorted out

src/website/webview/templates/components/suggestion.html
```diff
@@ -53,14 +61,6 @@
         <button type="submit" name="create-draft-not-implemented" class="draft-color">
           C̶̖͆r̶͍̚e̵͈̐ă̷̖t̷̄͜e̶͓̓ ̴̟̏d̵̳̅r̸̮̐a̵͍͑f̵̙̋t̴̹̃
         </button>
-        {% if cached_suggestion.packages %}
```
fricklerhandwerk, line 56: Doesn't this disable the scriptless workflow?

erictapen: Yeah thats right. From the commit message:

> This means that no-JS users will need to commit a status change when
removing packages, but at least that fact should be communicated well as
the page never reloads on package removal.

fricklerhandwerk: Why not just remove the element when scripting is enabled?

Didn't test it but this could work:

```
<div hx-get="#" hx-swap="delete" hx-trigger="load">
    This will be removed when HTMX loads
</div>
```

erictapen: I spent some time trying to do that in htmx, but didn't figure out how. Would be cool if htmx had some init event that can be used for `hx-on`.

For now I included one line of JS code in the `base.html` file.

fricklerhandwerk: I also had a look into that, but the classic seems to be the only reasonable thing.

PR #424, opened 2024-12-05
feat: show package maintainers that are going to be pinged

### LINKED ISSUE #416: Show maintainers that will be pinged
Before publishing a security issue, as a security team member I want to know which maintainers will be pinged.

erictapen:
Closes #416 

tmp LV8YRVXq6l

### DIFF
--- a/pyrightconfig.json
+++ b/pyrightconfig.json
@@ -1,0 +1,3 @@
+{
+  "reportAttributeAccessIssue": false
+}
--- a/src/website/shared/listeners/cache_suggestions.py
+++ b/src/website/shared/listeners/cache_suggestions.py
@@ -6,7 +6,7 @@ from typing import Any
 import pgpubsub
 
 from shared.channels import CVEDerivationClusterProposalChannel
-from shared.models import NixDerivation
+from shared.models import NixDerivation, NixMaintainer
 from shared.models.cached import CachedSuggestions
 from shared.models.cve import AffectedProduct, Metric, Version
 from shared.models.linkage import CVEDerivationClusterProposal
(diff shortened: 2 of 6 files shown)


src/website/shared/listeners/cache_suggestions.py
```diff
@@ -92,23 +92,28 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
...
+    derivations = list(
         suggestion.derivations.select_related("metadata", "parent_evaluation")
         .prefetch_related("outputs", "dependencies")
         .all()
     )
 
     prefetched_metrics = Metric.objects.filter(container__cve=suggestion.cve)
 
+    prefetched_maintainers = NixMaintainer.objects.filter(
+        nixderivationmeta__in=[d.metadata for d in derivations]
```
fricklerhandwerk, line 104: This will also show maintainers for "removed" (filtered-out) derivations

src/website/webview/templates/components/suggestion.html
```diff
@@ -38,6 +38,25 @@
       {% selectable_nixpkgs_package_list cached_suggestion.packages %}
     {% endif %}
 
+    {% if cached_suggestion.maintainers %}
+      <details class="maintainers">
+        <summary>
+          {{ cached_suggestion.maintainers | length }} package maintainer(s) would be pinged
```
fricklerhandwerk, line 44:
```suggestion
          Notify package maintainers: {{ cached_suggestion.maintainers | length }}
```

Before going full gettext we should just try to avoid any sort of grammar... I know this ugly but among all cosmetics this scales better until we have the time for polishing.

erictapen:
1sec, it still shows duplicates

erictapen:
Finally, should be ready now.

pyrightconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "reportAttributeAccessIssue": false
```
erictapen, line 2: So this is a super annoying lint, that produces unusable errors imo.

Everytime I try to use `Prefetch` with `to_attr`, I can't use the attr that is assigned to the object, as I otherwise get this type error. So I'd rather just disables it.

fricklerhandwerk (approved):

PR #428, opened 2024-12-08
refactor(suggestions-log): implement as on db bulk aggregation query

### LINKED ISSUE #427: Dismissed view takes forever to load
Using the staging DB dump from 2024-12-05, the list of dismissed suggestions takes a minute for domLoading.

Not sure what causes it, could be a very nasty suggestion item.

alejandrosame:
Closes #427

Some local benchmark comments:
    - From main, before refactor, running the queries in the selected page gave me: + CPU 3124ms DOMloading 4.23s.
    - After changing the query to a union an on db aggregation:
        + CPU 322ms DOM 638ms
    - After bulk filtering in the query (that is, adding pgh_obj_id__in filtering): + CPU 409ms DOM 831ms
    - After adding final dictionary aggregation:
        + CPU 308ms DOM 802ms

Therefore, with more data in the future, it can be reconsidered whether the bulk grouping is offering a good speedup tradeoff over filtering by single suggestion IDs.

The biggest speedup in this refactor was achieved by eliminating data roundtrips by using union and subqueries in the activity log fetch.

### DIFF
--- a/src/website/shared/logs/logs.py
+++ b/src/website/shared/logs/logs.py
@@ -1,10 +1,14 @@
-from collections import OrderedDict
-from typing import Any
+import datetime
+import typing
+from typing import Any, TypedDict
 
 from django.contrib.auth.models import User
+from django.contrib.postgres.aggregates import ArrayAgg
 from django.db.models import (
     BigIntegerField,
     Case,
+    Count,
+    F,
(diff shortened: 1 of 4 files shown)


fricklerhandwerk (approved):
It ain't pretty but it works

PR #432, opened 2024-12-09
fix: rename /selected to /drafts

erictapen:
Follow-up to #430 

I choose "drafted" in the activity log, even though I'm not sure how good of a fit it is. "To draft a letter" is a thing though.

### DIFF
--- a/src/website/webview/templates/components/suggestion_activity_log.html
+++ b/src/website/webview/templates/components/suggestion_activity_log.html
@@ -54,7 +54,7 @@
         {% else %}
           {# Status update entries #}
           {% if "accepted" in log_entry.status_value  %}
-            <span class="suggestion-status-draft">drafted</span>
+            <span class="suggestion-status-draft">accepted as draft</span>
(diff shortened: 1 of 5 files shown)


src/website/webview/templates/components/suggestion_activity_log.html
```diff
@@ -54,7 +54,7 @@
         {% else %}
           {# Status update entries #}
           {% if "accepted" in log_entry.status_value  %}
-            <span class="suggestion-status-selected">selected</span>
+            <span class="suggestion-status-draft">drafted</span>
```
fricklerhandwerk, line 57:
```suggestion
            <span class="suggestion-status-draft">accepted as draft</span>
```

How about this?

fricklerhandwerk (approved):

src/website/webview/templates/components/suggestion_activity_log.html
erictapen: cool :+1:

PR #435, opened 2024-12-09
fix: ingest newest CVEs, not oldest CVEs

RaitoBezarius:
This explains why we were never catching up, we were always restarting from scratch.


fricklerhandwerk (approved):
Yep. Software is hard.

PR #437, opened 2024-12-09
feat(suggestion): aggregate package updates by 5 min intervals

### LINKED ISSUE #425: Summarise changes in the activity log display
tmp wOxkuJahh3

alejandrosame:
Closes #425

### DIFF
--- a/src/website/shared/logs/logs.py
+++ b/src/website/shared/logs/logs.py
@@ -8,6 +8,7 @@ from django.db.models import (
     BigIntegerField,
     Case,
     Count,
+    DateTimeField,
     F,
     OuterRef,
     Q,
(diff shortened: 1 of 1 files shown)


alejandrosame:
Related to timestamp to display (https://github.com/Nix-Security-WG/nix-security-tracker/issues/425#issuecomment-2526297440):

Since I ended up using `date_bin` to group timestamps by an arbitrary 5 min interval, the timestamp now shows rounded to the closest 5 min interval. I hope it's relatively clearer than timestamp to easily signal this is a bucketed aggregation.

If we end up having to do more passes over the aggregated packages in Python, then the latest timestamp could also be reintroduced relatively easily.

alejandrosame:
Closing since https://github.com/Nix-Security-WG/nix-security-tracker/pull/445 reflects better the proposed idea in #425.

From that other implementation, it shouldn't be too hard to add an extra check to also group by some arbitrary time interval, if that seems appealing in the future.

PR #443, opened 2024-12-09
feat: show username of logged in user in navbar

alejandrosame:
image

### DIFF
--- a/src/website/shared/templates/base.html
+++ b/src/website/shared/templates/base.html
@@ -28,7 +28,16 @@
 
         {% block auth %}
         {% if user.is_authenticated %}
+        {% get_social_accounts user as accounts %}
         <li>
+          {% if accounts and "github" in accounts %}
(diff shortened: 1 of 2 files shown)


src/website/shared/templates/base.html
```diff
@@ -29,6 +29,7 @@ <h1>
         {% block auth %}
         {% if user.is_authenticated %}
         <li>
+          <span>{{user.username}}</span>
```
fricklerhandwerk, line 32:
```suggestion
          {% if user.provider == "github" %}
            <a href="https://github.com/{{ user.username }}" target="_blank">{{ user.username }}</a>
          {% else %}
            <span>{{ user.username }}</span>
          {% endif %}
```

What we we do something like this, but corrected for the right login name from socialaccounts?

alejandrosame: Amended this approach in the commit

fricklerhandwerk (approved):
Nice, thanks

PR #451, opened 2024-12-10
feat: make eval store configurable

fricklerhandwerk:
evaluations write garbge we don't use to disk if we hit the physical store, so this should speed things up a bit

PR #452, opened 2024-12-10
feat: display src file location next to the version for each derivation

### LINKED ISSUE #440: Link source location from derivation per channel
As a security team member I want to quickly inspect a derivation source, right from the record linkage suggestion.

Implementation note: link from the version field per channel.

erictapen:
Closes #440 

tmp YSrXBRubZ0

~~The code should be robust for e.g. `nixpkgs-darwin-24.11` and `nixos-24.11-small` to have different revisisions, though I never see any difference between these. Probably just because they evaluation did actually happen on the same commits?~~

### DIFF
--- a/src/website/shared/listeners/cache_suggestions.py
+++ b/src/website/shared/listeners/cache_suggestions.py
@@ -189,7 +189,11 @@ def get_src_position(derivation: NixDerivation) -> str | None:
         rev = urllib.parse.quote(derivation.parent_evaluation.commit_sha1)
         # position is something like `/tmp/tmpfh7ff2xs/pkgs/development/python-modules/qemu/default.nix:67`
         position_match = re.match(
-            r"/tmp/[^/]+/(.+):(\d+)", derivation.metadata.position
+            # FIXME the location of the eval store is going to be configurable in the future.
+            # https://github.com/Nix-Security-WG/nix-security-tracker/pull/451
+            # Ideally the position field is already relative to the location.
+            r"/tmp/[^/]+/(.+):(\d+)",
+            derivation.metadata.position,
         )
         if position_match:
             path = urllib.parse.quote(position_match.group(1))
--- a/src/website/shared/listeners/cache_suggestions.py
+++ b/src/website/shared/listeners/cache_suggestions.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import urllib.parse
 from itertools import chain
 from typing import Any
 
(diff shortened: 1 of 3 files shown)


src/website/shared/listeners/cache_suggestions.py
```diff
@@ -179,6 +180,24 @@ def is_version_affected(version_statuses: list[str]) -> Version.Status:
...
+def get_src_position(derivation: NixDerivation) -> str | None:
+    """
+    Get the GitHub URL pointing to the exact source file used for the evaluation of this derivation.
+    E.g. https://github.com/NixOS/nixpkgs/blob/0e8be3827d0298743ba71b91eea652d43d7dc03d/pkgs/by-name/he/hello/package.nix#L47
+    """
+    if derivation.metadata and derivation.metadata.position:
+        rev = urllib.parse.quote(derivation.parent_evaluation.commit_sha1)
+        # position is something like `/tmp/tmpfh7ff2xs/pkgs/development/python-modules/qemu/default.nix:67`
+        position_match = re.match(
+            r"/tmp/[^/]+/(.+):(\d+)", derivation.metadata.position
```
fricklerhandwerk, line 192: This won't scale with https://github.com/Nix-Security-WG/nix-security-tracker/pull/451, because where we evaluate is essentially a runtime configuration, but which may change between deployments... :/

@raitobezarius maybe we can scrape that out already at evaluation time?

erictapen:
tmp lJ8D0cvkKB
Removed the `</>`, as it takes too much screenspace. The version is the link now.

src/website/shared/listeners/cache_suggestions.py
erictapen: Oh I understand. Makes the distinction pretty much impossible then, right? I'll add a FIXME for now.

RaitoBezarius: If the eval store is in memory, I don't see why we cannot do that computation, it's just a string clean up?

erictapen: We would need to know the location then. Is the config value accessible in cache_suggestions? Also this would mean that all evaluations get invalidated, once you change the location of the eval store.

fricklerhandwerk (approved):

src/website/shared/listeners/cache_suggestions.py
fricklerhandwerk: The config value is accessible but in case of config change will get invalidated. This is why we should do it earlier where we know the value at the time. But also I just realised this is not necessary config store but location of the Nixpkgs checkout, so this can definitely just stripped off during eval-ingestion. We'll leave it at that now because it won't help for the demo.

PR #461, opened 2024-12-11
WIP: feat: tracking derivations via attribute paths

### LINKED ISSUE #307: Package versions in the suggestions view seem to repeat themselfes
I have no idea yet how it works or wether it's an actual issue, but there seems to be the pattern, that the old stable version and the unstable version seem to be equal, while the new stable version is newer. Hints towards a bug.

tmp J9lSQSs1Lo

RaitoBezarius:
Fixes #307.


fricklerhandwerk:
Superseded by #1035

Issue #399, opened 2024-11-29
Allow configuring a repo on which to post issues


fricklerhandwerk:
In order to post GitHub issues for our security records, we need to specify a repo where this should happen.

- possibly generate new secrets for the GitHub app in this repo
- encrypt secrets into the repo
- deploy

Issue #138, opened 2023-12-15
Open a GitHub issue for a security record


fricklerhandwerk:
As a security team member, I want to publish a triaged security issue to GitHub so it can be collaborated on. Doing this should:

> [!NOTE]
> For now, escape maintainer names in backticks so we don't actually ping people yet

- [ ] Post data from the tracker into the GitHub issue description

  - Title: CVE title
  - Body: CVE Description, CVE ID with link, CVSS as details
  - Package names with descriptions
    - channels in details, with source links
  
- [ ] Ping or assign maintainers of affected packages in the issue

Depends on:
- #208
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/399

Implementation notes:
- We should eventually be able to overwrite the issue body from the tracker, but only if it hasn't changed since our last write (maintainers may have edited the issue text and we don't want to lose that data)
- We want to demo this on some dummy repository, and not ping random people while playing around. Maybe there should be a safety switch that will escape the GitHub handles

yannham:
Should we close this now? There's still the question of customization, and a bunch of improvements that can be implemented, but is looks like the original scope of the issue has been implemented.

fricklerhandwerk:
Yep, looks good.

PR #462, opened 2024-12-11
feat: configure repository to post issues to

alejandrosame:
Closes https://github.com/Nix-Security-WG/nix-security-tracker/issues/399. Towards https://github.com/Nix-Security-WG/nix-security-tracker/issues/138


fricklerhandwerk (approved):
We should probably make the Django settings all crash with an error message to set the values in the module. Otherwise without the right config it will crash with more inscrutable GitHub errors. Opened https://github.com/Nix-Security-WG/nix-security-tracker/issues/464

PR #479, opened 2025-01-09
Remove old staging deployment, automatically deploy tracker.security.nixos.org

Erethon:
Opening as a draft PR for now as I want to verify that GH actions work as expected first.

This completely removes the old staging setup, @RaitoBezarius I believe you can now destroy that installation if you want to reclaim your resources. CC'ing @fricklerhandwerk to verify this is true.

This also updates GH actions to automatically update the new infra on code changes in the repo. For now it removes the database dumps because they depended on various infra components that were part of Raito's setup, but if we think they're important we can re-introduce them in a different manner. ~It also breaks the development using a container, but we can re-introduce this easily if it was used.~


fricklerhandwerk:
Dismantling the staging configuration is alright. We just need to keep track of what needs to get reconstructed. That gets kinda annoying when the original code is removed. I recommend opening issues for each feature and linking to the old commits so it's easy to find the previous state.

Erethon:
Marked this as ready to review, but I think we can merge as is and see if the CI fails after this (it shouldn't).

fricklerhandwerk (approved):
Yep. Just make sure to recreate the DB eventually. We can take care of auto-updating dependencies later, the code never built with unstable anyway.

PR #498, opened 2025-04-16
feat: open a GitHub issue

yannham:
**UPDATE**: this PR has been split into smaller ones. I'll keep this one open as a tracking device and to refer to this detailed description.

# Open a GitHub issue

Partly address #138 (some work left to do on labels and whatnot).

Screenshot From 2025-04-17 12-18-54

Screenshot From 2025-04-17 12-19-08

Sample of generated test issues: https://github.com/nix-sec-tracker-test/nixpkgs/issues

## Description

This PR implements an additional stage for suggestions: publication. When a suggestion is in the state `accepted`, a new (more precisely existing button that was disabled) "Publish issue" button is available. When clicked, the following actions are performed:

- the suggestion's status changes to a new "published" status
- a `NixpkgsIssue` object is created, taking the data from the suggestion
- a GitHub issue is created, taking the data from the suggestion (with title, description, cvss details, link to the CVE, link the created issue on the tracker, and list of maintainers of affected packages). The maintainer username is fetched from GH so that it's up to date
- the status change component contains a link to the created GH issue in place of the Undo button

## Content

- `utils.py` has been renamed to `github.py`, as it contained a single helper for creating a `GitHub` connection object
- an endpoint `create_gh_issue` has been added to `utils.py -> github.py`
- a new `PUBLISHED` status has been added for `CVEDerivationClusterProposal` and its accompanying migration
- the `CachedSuggestions` model has a new computed property `all_maintainers()`, which gather all maintainers an dedup-them
- The "Publish issue" button has been enabled, triggering a state change to "published"
- the suggestion state change component takes an additional `gh_issue_link` to link to the GitHub issue when it's been published
- a method `create_nixpkgs_issue()` has been added to `CVEDerivationClusterProposal` to derive a fresh `NixpkgsIssue` from the former

## TODO

- [x] Tag maintainers (escaped for now) in the GH issue
- [ ] [EDIT: left for follow-up PR] Include package names with description in the GH issue description

PR #510, opened 2025-04-28
refactor: add GH_ISSUES_PING_MAINTAINERS setting

yannham:
Split from #498: preliminary work adding a setting to escape maintainers' names when creating a GitHub issue to avoid pinging them while the feature isn't yet well tested.

### DIFF
--- a/src/website/tracker/settings.py
+++ b/src/website/tracker/settings.py
@@ -336,6 +336,12 @@ DB_SECURITY_TEAM = "security_team"
 
 GH_WEBHOOK_SECRET = get_secret("GH_WEBHOOK_SECRET")
 
+# When set to False, the application will escape package maintainers' name when
+# mentioning them in the issue body to avoid actually pinging them. This is used
+# as a safety measure during development. When the GitHub issue story is ironed
+# out, this should be set to True in production.
+GH_ISSUES_PING_MAINTAINERS = False
+
 TEST_RUNNER = "tracker.test_runner.CustomTestRunner"
 
 # Make history log immutable by default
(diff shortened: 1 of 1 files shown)


src/website/tracker/settings.py
```diff
@@ -336,6 +336,12 @@ def get_secret(name: str, encoding: str = "utf-8") -> str:
 
 GH_WEBHOOK_SECRET = get_secret("GH_WEBHOOK_SECRET")
 
+# When set to False, the application will escape package maintainers' name when
+# mentioning them in the issue body to avoid actually pinging them. This is used
+# as a safety measure during development. When the GitHub issue story is ironed
+# out, this should be set to True in production.
+GH_ISSUES_PING_MAINTAINERS = False
```
fricklerhandwerk, line 343: Please add that to DjangoSettings (~l.46), otherwise it's really annoying to keep track of those options.

fricklerhandwerk (changes requested):

src/website/tracker/settings.py
yannham: Done in latest commit.

src/website/tracker/settings.py
```diff
@@ -45,6 +45,11 @@ class DjangoSettings(BaseModel):
             of Nixpkgs maintainers in the evaluation database.
             """
         )
+        # When set to False, the application will escape package maintainers' name when
+        # mentioning them in the issue body to avoid actually pinging them. This is used
+        # as a safety measure during development. When the GitHub issue story is ironed
+        # out, this should be set to True in production.
+        GH_ISSUES_PING_MAINTAINERS: bool = False
```
fricklerhandwerk, line 52:
```suggestion
        GH_ISSUES_PING_MAINTAINERS: bool = Field(
            description="""
            When set to False, the application will escape package maintainers' name when
            mentioning them in a GitHub issue to avoid actually pinging them.
            This is used as a safety measure during development. Set to True in production.
            """
        )
```

fricklerhandwerk (approved):

PR #514, opened 2025-04-28
feat: add create_gh_issue() helper

yannham:
Split off from #498. This PR adds an endpoint to the `github` module to create and publish a GitHub issue on the Nixpkgs repository (actually the one configured in `settings`) from an accepted suggestion.

### DIFF
--- a/src/website/shared/github.py
+++ b/src/website/shared/github.py
@@ -1,8 +1,18 @@
 import logging
 from os import environ as env
+from urllib.parse import quote
 
 from github import Auth, Github
-from tracker.settings import get_secret
+from github.Issue import Issue as GithubIssue
+from tracker.settings import (
+    GH_ISSUES_PING_MAINTAINERS,
+    GH_ISSUES_REPO,
+    GH_ORGANIZATION,
+    get_secret,
+)
+from webview.templatetags.viewutils import severity_badge
+
+from shared.models import CachedSuggestions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


src/website/shared/github.py
```diff
@@ -42,3 +52,96 @@ def get_gh(per_page: int = 30) -> Github:
             )
 
     return Github(auth=gh_auth, per_page=per_page)
+
+
+def create_gh_issue(
+    cached_suggestion: CachedSuggestions,
```
fricklerhandwerk, line 58: Side note, ignore: why is the class name plural? 😮‍💨

fricklerhandwerk (commented):
Will merge this after the setting

src/website/shared/github.py
yannham: Yeah, I wondered the same :slightly_smiling_face:

src/website/shared/github.py
```diff
@@ -42,3 +52,96 @@ def get_gh(per_page: int = 30) -> Github:
...
+        else:
+            return ""
+
+    repo = github.get_repo(f"{GH_ORGANIZATION}/{GH_ISSUES_REPO}")
+    title = cached_suggestion.payload["title"]
+    logger.error("all maintainers: %s", cached_suggestion.all_maintainers)
+
+    body = f"""\
+- [{cached_suggestion.payload['cve_id']}](https://nvd.nist.gov/vuln/detail/{quote(cached_suggestion.payload['cve_id'])})
+- [Nix security tracker issue]({tracker_issue_uri})
```
fricklerhandwerk, line 125:
```suggestion
- [Nixpgks security tracker issue]({tracker_issue_uri})
```

src/website/shared/github.py
```diff
@@ -42,3 +52,96 @@ def get_gh(per_page: int = 30) -> Github:
...
+{cvss_details()}"""
+
+    return repo.create_issue(title, body)
+
+
+def get_maintainer_username(maintainer: dict, github: Github = get_gh()) -> str:
+    """
+    Get the current GitHub username of a maintainer given their user ID. If the
+    request failed, fallback to the github handle stored in the maintainer
+    object that comes from Nixpkgs, which might be out of date.
```
fricklerhandwerk, line 139:
```suggestion
    object that comes from Nixpkgs, which might be out of date.
    # TODO: Cache the mapping, e.g. on initial sync and when receiving GitHub events
    # on username change, or simply when doing these calls for resolving the user ID.
```

fricklerhandwerk: https://github.com/Nix-Security-WG/nix-security-tracker/issues/521

src/website/shared/github.py
RaitoBezarius:
```suggestion
- [Nixpkgs security tracker issue]({tracker_issue_uri})
```

src/website/shared/github.py
yannham: Good point; I've tried to find a way to at least do a bulk request for a set of IDs but it doesn't seem to exist.

PR #516, opened 2025-04-28
feat: published status in the view

yannham:
Split off from #498. Adds a "Publish issue" button to update a suggestion from accepted to published. Update the status change component to link to the opened Github issue (instead of the undo button). Note that this commit doesn't actually implement any backend logic, so the button has no effect currently: it's purely a view thing.


yannham:
Superseded by #523.

PR #523, opened 2025-04-29
feat: re-enable "Publish issue" button

yannham:
This is the last chunk of #498, which adds the "Publish issue" button and implements its behavior in the backend. Includes and supersedes #516.


fricklerhandwerk (approved):
Modulo error handling, but can be done in a follow-up

PR #524, opened 2025-04-29
feat: add status change error component

yannham:
Follow-up of #516. This PR adds a small component, similar to the one handling state changes, to report errors during suggestion state transitions. More specifically here, the potentially fallible transition is to publish a GitHub issue, as this implies an external call to the GitHub API. In that case, if any step of the publication fails, the component will show that an error has occurred:

image

## TODO

- [x] The suggestion still disappears temporarily from the current view, which isn't desirable. It's still here if we refresh the page, but upon error, the suggestion shouldn't go anywhere in the first place.

### DIFF
--- a/src/website/webview/static/style.css
+++ b/src/website/webview/static/style.css
@@ -358,6 +358,7 @@
   --automatic-suggestion: lightblue;
   --dismissed-suggestion: #acb3c8;
   --draft-suggestion: #7bca7b;
+  --error-suggestion: #ffabab;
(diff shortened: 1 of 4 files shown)


fricklerhandwerk (approved):
Yep. This feels like we're still slowly piling up tech debt, but I believe an obvious pattern to separate the pure-HTML code path will soon emerge. And once we start catching up with tests, refactorings will be safe again.

PR #527, opened 2025-04-30
feat: (un)subscribe button

yannham:
Partially address #173. Depends on #526. Adds a (un)subscribe button on the issue detail view to subscribe or unsubscribe to a specific issue. The front-end part is currently really basic, let's use this PR as a basis for discussion.

## TODO/To discuss

- [ ] Add a status change message?
- [ ] Avoid a reload of the page when posting?
- [ ] (I'd like to leave that for a follow-up) add (un)subscribe buttons in the issue list view as well


fricklerhandwerk:
The simple stupid way to do separate the code path for pure-HTML is to have a regular POST response for the full page reload, and let the htmx annotation on the button call a different endpoint that serves just that purpose. It's slightly noisy because you'd need to add an extra URL and view, and eventually we'd better build an abstraction.

PR #528, opened 2025-04-30
feat: attach labels to github issue

### LINKED ISSUE #501: Add GitHub issue labels based on the tracker data
As a security team member of package maintainer, I want to find issues published through the security tracker on GitHub by label, in order to keep using existing workflows.

## Acceptance criteria

Given there's a draft issue,
When that issue is published to GitHub
Then the security tracker adds issue labels as configured.

## Implementation notes

- Currently there seems to be only one label or security issues on Nixpkgs: https://github.com/NixOS/nixpkgs/issues?q=is%3Aopen%20is%3Aissue%20label%3A%221.severity%3A%20security%22
- We need to make a list of labels configurable at deployment time, since the upstream label may be renamed
- We probably shouldn't have a default (and neither for the other deployment-specific parameters, such as the repo and teams) since if you want these values to be set very deliberately

yannham:
Closes #501. The code looks fine but I have a permission issue to solve to properly test it; labels are just ignored currently, but if I try to add them manually on the created issue with `add_to_labels`, I get a proper 403, so something needs to be done (and documented) around GitHub access rights/PATs.

### DIFF
--- a/src/website/tracker/settings.py
+++ b/src/website/tracker/settings.py
@@ -79,6 +79,16 @@ class Settings(BaseSettings):
             Set `nixpkgs-committers` for the production deployment.
             """
         )
+        GH_ISSUES_LABELS: list[str] = Field(
+            description="""
+            Labels to attach to Github issues created from the tracker, making
+            it easier to filter them on the nixpkgs repo.
+            """,
+            # It's always ok to operate with an empty list of labels both in
+            # production and in development mode. Override accordingly depending
+            # on the environment.
+            default=[],
+        )
 
     DJANGO_SETTINGS: DjangoSettings
 
--- a/src/website/tracker/settings.py
+++ b/src/website/tracker/settings.py
@@ -82,7 +82,7 @@ class Settings(BaseSettings):
         GH_ISSUES_LABELS: list[str] = Field(
             description="""
             Labels to attach to Github issues created from the tracker, making
-            it easier to filter them on the nixpkgs repo.
+            it easier to filter them on the target repository.
             """,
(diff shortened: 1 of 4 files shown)


src/website/tracker/settings.py
```diff
@@ -336,6 +336,9 @@ def get_secret(name: str, encoding: str = "utf-8") -> str:
 GH_SECURITY_TEAM = "security"
 # Repository to post issues to
 GH_ISSUES_REPO = "nixpkgs"
+# Labels to set automatically on created Github issues, allowing users to filter
+# them.
+GH_ISSUES_LABELS = []
```
fricklerhandwerk, line 341: Please add to the DjangoSettings instead so we can configure it uniformly. Everything beneath the TODO in that file is legacy and I'll clean it up eventually.

yannham:
Example issue with labels: https://github.com/nix-sec-tracker-test/nixpkgs/issues/58

src/website/tracker/settings.py
```diff
@@ -79,6 +79,16 @@ class DjangoSettings(BaseModel):
             Set `nixpkgs-committers` for the production deployment.
             """
         )
+        GH_ISSUES_LABELS: list[str] = Field(
+            description="""
+            Labels to attach to Github issues created from the tracker, making
+            it easier to filter them on the nixpkgs repo.
```
fricklerhandwerk, line 85:
```suggestion
            it easier to filter them on the target repository.
```

fricklerhandwerk (approved):

PR #530, opened 2025-05-02
chore: formatting

fricklerhandwerk:


yannham (approved):

fricklerhandwerk:
It's weird how this keeps re-appearing. `nix-build -A pre-commit-check` didn't complain *before* or after the change, but the commit hook does complain.

Issue #532, opened 2025-05-02
Subscribe to notifications on a package


fricklerhandwerk:
As a user I want to be able to subscribe to notifications for issues published that affect some package.

## Acceptance criteria

Given I have found some package,
When I subscribe to notifications for it,
Then there's an indication that I did and it shows up in my list of subscriptions.

Given I'm subscribed to some package,
When I unsubscribe from notifications for it,
Then there's an indication that I did and it's gone from my list of subscriptions.

Given I'm subscribed to some package,
When a security record that affects it *or any of its buildtime and runtime dependencies* is **published** ,
Then I get a notification according to my set preferences.

Given I'm subscribed to some package,
When a security record that affects it *or any of its buildtime and runtime dependencies* is **closed** ,
Then I get a notification according to my set preferences.

The notifications should have links to the security record in question and the corresponding GitHub issue.

Depends on:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/215

florentc:
To sum up and confirm, is this mockup of the user profile setting view (on the left) and package list view (on the right) the kind of thing we are targetting eventually?

![Image](https://github.com/user-attachments/assets/dc896229-494a-49fa-9ac6-19f43baadc9a)

fricklerhandwerk:
Yes, and we may arrange the currently left box on top of the currently right one depending on screen size/format, but structurally this is it. As @balsoft proposed those boxes are supposed to belong together so we could in principle have multiple such collections in the future, but visually it doesn't matter right now.

PR #535, opened 2025-05-05
Add relations in Profile for subscribed channels and packages

florentc:
As a step to implement #532, this adds relations between:
- profiles and channels
- profiles and derivations

### DIFF
--- a/src/website/webview/models.py
+++ b/src/website/webview/models.py
@@ -5,7 +5,8 @@ from django.contrib.auth.models import User
 from django.db import models
 from django.db.models.signals import post_save
 from django.dispatch import receiver
-from shared.models import NixpkgsIssue
+from shared.models import NixChannel
+from shared.models.nix_evaluation import NixDerivation
 
 
 class Profile(models.Model):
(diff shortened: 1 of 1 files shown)


src/website/webview/models.py
```diff
@@ -15,7 +16,8 @@ class Profile(models.Model):
     """
 
     user = models.OneToOneField(User, on_delete=models.CASCADE)
-    subscriptions = models.ManyToManyField(NixpkgsIssue, related_name="subscribers")
+    channels = models.ManyToManyField(NixChannel, related_name="subscribers")
```
fricklerhandwerk, line 19: @balsoft suggested to actually have something like "collections" or "profiles" (see https://github.com/Nix-Security-WG/nix-security-tracker/issues/532#issuecomment-2851394198), which I think is a good idea if we eventually want users to be able to subscribe programmatically from clients on different machines/setups. Therefore how about we put this into a container?

fricklerhandwerk:
Subscription to "packages" (attrnames) has been done in https://github.com/NixOS/nix-security-tracker/pull/644.

Issue #215, opened 2024-09-27
Access affected packages by attribute name


fricklerhandwerk:
As a user of the security tracker, I want to be able to search for particular derivations.

## Acceptance criteria

Given I have an attribute path for a package in Nixpkgs,
When I enter that string into a text field (or URL),
Then I get a list of suggestions where that package is active

## Implementation notes

- Also add a link in suggestion details from package name to the per-package list
- Show all suggestions where that package appears, show the status in each item.
  - We can still put things into sub-views if this turns out too messy in practice

yannham:
Naive question: can't we just reuse search.nixos.org in one way or another (including updating the latter if needed to expose some kind of API)? I remember when Rok and other people implemented search.nixos.org, I think it's not trivial to do efficiently; duplicating this work on the sec tracker would be a bit sad.

RaitoBezarius:
That website uses an externally donated Elastic Search instance I believe.

Plus it doesn't fit the needs for the queries used in this project.

So it's a matter of choosing whether to redeploy ELK, reuse PGSQL native
FTS engine or something else.

Le lun. 5 mai 2025, 14:55, Yann Hamdaoui ***@***.***> a
écrit :

fricklerhandwerk:
Closed by #782

PR #537, opened 2025-05-07
Add a model for package

florentc:
As a step to implement #532, this WIP PR introduces a db model for packages. For now, a package only has a name. The nix derivation table gains a column, package, for the many-to-one relation from derivation to package. The provided migration ensures the creation of packages associated to all derivations and links each derivation to the right package.

The idea is that user profiles will be in relation with packages to model subscriptions. This will make it possible to subscribe to a package name as we discussed with @fricklerhandwerk  and make queries easier to link a "package" and the many derivations. Eventually, we could envision to migrate some metadata from derivations, which are in fact shared and package-related, to this package model.


src/website/shared/migrations/0049_nixpackage_nixderivation_package.py
```diff
@@ -0,0 +1,44 @@
+from django.db import migrations, models, transaction
+import django.db.models.deletion
+
+def populate_package_field(apps, schema_editor):
+    NixPackage = apps.get_model('shared', 'NixPackage')
+    NixDerivation = apps.get_model('shared', 'NixDerivation')
+
+    with transaction.atomic():
```
fricklerhandwerk, line 8: This will probably stop the world for a long time

florentc:
@fricklerhandwerk I wanted your opinion/advice about implementing setting up the package of a derivation automatically and eventually making the package field non nullable. What would be the go-to way to automatically get_or_create a package and assign it to the package field?

Should I override the `save` method of the model or use some signal (post_save or other?)

This would look something like this in both cases:

```python
package, _ = NixPackage.objects.get_or_create(name=self.attribute)
self.package = package
```

src/website/shared/migrations/0049_nixpackage_nixderivation_package.py
florentc: It did take something like 5 min. Do you know how it could be optimized?

fricklerhandwerk:
To be honest, I don't know, but I agree we'd need something like that.

Although I wonder whether it's a good idea to store the "package" aggregation at the relational layer. We need that information very quickly for (anything like) search (#215), so it needs to be cached anyway. So why not instead let the `NixPackage` model have a nullable `JSONField` that would store the derivations already together with whatever data we need, channel and version in particular? Then we could populate it on first access (e.g. when it's `None` and then update whenever a new `NixDerivation` is ingested, which would need a `post_save` signal on `NixDerivation`).

src/website/shared/migrations/0049_nixpackage_nixderivation_package.py
fricklerhandwerk: Generally something like here: https://github.com/Nix-Security-WG/nix-security-tracker/pull/538/files

But see my top level comment for why we may not need that anyway

fricklerhandwerk:
Towards https://github.com/NixOS/nix-security-tracker/issues/617

adekoder (approved):

PR #539, opened 2025-05-09
CI: add actionlint

smelc:
## Context

While working on plugging CodeQL in this repository, I noticed there was no check on the quality of workflow files, so I'm proposing to add one.

## How to trust this PR

* [actionlint](https://github.com/rhysd/actionlint) is a well-established checker
* There was only one warning found by actionlint, which I fixed in the second commit: well done team 👏
* I ran this pipeline manually on my fork, it looks like this (when only warning is not corrected: [this run](https://github.com/smelc/nix-security-tracker/actions/runs/14882894784/job/41794975378)):

image

### DIFF
--- a/.github/workflows/actionlint.yaml
+++ b/.github/workflows/actionlint.yaml
@@ -1,0 +1,31 @@
+name: Actionlint
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/**"
+  push:
+    branches: main
+    paths:
+      - ".github/workflows/**"
+  workflow_dispatch:
+
+# Restrict permissions for GITHUB_TOKEN to minimum required
+permissions:
+  contents: read
+  # All other permissions are implicitly none
+
+jobs:
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        with:
+          # Discard `GITHUB_TOKEN` after this step, as we don't need it
+          persist-credentials: false
+      - uses: cachix/install-nix-action@v31
+      - name: Check workflow files
+        run: |
+          NIXPKGS=$(nix-instantiate --eval -A pkgs.path)
+          nix-shell -I nixpkgs="$NIXPKGS" -p actionlint --run actionlint


.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,23 @@
+name: Actionlint
+
+on:
+  pull_request:
+  push:
+    branches: main
```
raboof, line 6: should we restrict this to PRs that touch the workflows?

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,23 @@
...
+  # allow "manual" triggering from automatic PRs
+  workflow_dispatch:
```
raboof, line 8: I'm not sure I understand this comment - does that apply for this repo?

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,23 @@
...
+  actionlint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      # Taken from https://github.com/rhysd/actionlint/blob/v1.7.7/docs/usage.md#use-actionlint-on-github-actions
+      - name: Download actionlint
+        id: get_actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+        shell: bash
```
raboof, line 20: This basically runs arbitrary code from a 3rd-party repo. When triggered with 'push' or 'workflow_dispatch', I'm pretty sure a GitHub Action runs with access to privileges (the 'permissive' set in https://docs.github.com/en/actions/security-for-github-actions/security-guides/automatic-token-authentication#permissions-for-the-github_token).

Could we lock down the token for this workflow? And disable `persist-credentials` for the checkout action above? (https://github.com/actions/checkout/)

.github/workflows/actionlint.yaml
smelc: Indeed sorry, this is coming from another repo of mine. I removed the comment :+1:

.github/workflows/actionlint.yaml
smelc: That makes sense :+1:

Done that using a [path filter](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore) in the additional commit `ci: run actionlint only when .github/workflows/** changes`.

@raboof> please make sure to check that the workflow indeed runs on this PR when you approve them.

.github/workflows/actionlint.yaml
smelc: @raboof> set `persist-credentials` to false (thanks, didn't know about this one) and restricted `GITHUB_TOKEN`'s rights to the minimum :+1:

Erethon: > @raboof> set `persist-credentials` to false (thanks, didn't know about this one) and restricted `GITHUB_TOKEN`'s rights to the minimum 👍 See the last two commits.

You might be interested in https://github.com/woodruffw/zizmor that catches this. If you want an example of how it's already used in nixos/infra, https://github.com/NixOS/infra/blob/b464b1d87008ce8b61838a7d6ae55fa9b2cd5f11/.github/workflows/zizmor.yml

smelc: Cool, didn't know about zizmor, thanks for sharing @Erethon :heart:

smelc:
Rebased and CI is green. @raboof, @Erethon> can I get a rereview and an approval? cc @fricklerhandwerk

.github/workflows/actionlint.yaml
```diff
@@ -0,0 +1,33 @@
...
+        with:
+          persist-credentials: false # Because installing actionlint runs arbitrary code, we discard the GITHUB_TOKEN after this step
+      # Taken from https://github.com/rhysd/actionlint/blob/v1.7.7/docs/usage.md#use-actionlint-on-github-actions
+      - name: Download actionlint
+        id: get_actionlint
+        run: bash <(curl https://raw.githubusercontent.com/rhysd/actionlint/main/scripts/download-actionlint.bash)
+        shell: bash
+      - name: Check workflow files
+        run: ${{ steps.get_actionlint.outputs.executable }}
+        shell: bash
```
fricklerhandwerk, line 33:
```suggestion
      - name: Check workflow files
        run: nix-shell -p actionlint --run actionlint

```

I wonder whether we can just do something like this? Ideally using the same version of Nixpkgs as for anything else to avoid another download.

smelc:
@fricklerhandwerk> good suggestion :heart:

.github/workflows/actionlint.yaml
fricklerhandwerk: Ah, we need the nix-install-action here first...

fricklerhandwerk: btw here's a cumbersome but workable way to sync the nixpkgs version: https://github.com/ngi-nix/ngipkgs/blob/52c0b887994e3f497ef21878ccfc2c7b8b292d5c/.github/workflows/test-demo.sh#L34-L39

smelc: @fricklerhandwerk> sorry I'm not Nix-savvy enough to fully adapt the snippet to this repo. Can you correct the following:

```shell
      - name: Check workflow files
        run: |
          nixpkgs_revision=$(nix-instantiate --eval --attr sources.nixpkgs.rev /ngipkgs | jq --raw-output)
          NIXPKGS="https://github.com/NixOS/nixpkgs/archive/$nixpkgs_revision.tar.gz"
          nix-shell --include nixpkgs="$NIXPKGS" --packages nix --run "actionlint"
```

Can you adapt the `nixpkgs_revision` definition to this repo :grimacing:?

smelc: @fricklerhandwerk> sorry this is beyond me, I have no idea what `nix-instantiate` does for example. So either we should drop this PR, or ask someone more savvy to tackle this thread :slightly_smiling_face:

fricklerhandwerk: No problem, I'll pick it up then. It's low priority, so no timeline.

Issue #544, opened 2025-05-13
Match visual style of published vuln. records to suggestions and drafts


fricklerhandwerk:
The template for published records is out of date because we haven't been working on it in the last development phase. Match its style to the up-to-date views so it looks uniform (and the code is a bit cleaner).

PR #547, opened 2025-05-13
style: unify page width for all views

fricklerhandwerk:
towards #544

this simplifies the style sheet a bit, removing superfluous classes.


MurdoqZodoq (commented):
Ruby

Issue #500, opened 2025-04-21
Navigate to moved record


fricklerhandwerk:
As a user of the security tracker, whenever I change the state of a suggestion or issue (e.g. to "draft" or "dismissed"), I want to be able to view that new state immediately, in order not to have to search for it in a different place.

## Acceptance criteria

- New -> Dismissed

  Given there's a new suggestion,
  When I dismiss it,
  Then there's a way to navigate directly to the dismissed suggestion.

- Dismissed -> Draft

  Given there's a dismissed suggestion,
  When I mark it as draft,
  Then there's a way to navigate directly to the draft issue.

- New -> Draft

  Given there's a new suggestion,
  When I mark it as draft,
  Then there's a way to navigate directly to the draft issue.

- Draft -> Dismissed

  Given there's a draft issue,
  When I mark it as dismissed,
  Then there's a way to navigate directly to the dismissed suggestion.

- Draft -> Published

  Given there's a draft issue,
  When I publish it,
  Then there's a way to navigate directly to the published GitHub issue **and** to the security tracker issue.

## Implementation notes

Since we also support a pure HTML workflow, these are actually two things to keep parity. To make that somewhat tractable there currently needs to be a bit of cleanup to separate the htmx and the pure HTML code paths and test them separately.

PR #552, opened 2025-05-13
Add a button to navigate to moved records

florentc:
This addresses #500 

- This adds a button to navigate to a suggestion after its status changes:

- When clicked on, it goes to the page corresponding to suggestions of the new status (e.g. when dismissing, it goes to "dismissed suggestions") and highlights briefly the suggestion in question

As for browsing published issues, it will work automatically once the same nice view is used for the  "published issues" tab as for the other tabs.


src/website/webview/static/style.css
```diff
@@ -489,6 +490,19 @@ article .suggestion {
   flex-grow: 1;
 }
 
+@keyframes suggestion-quick-highlight {
+  0% {
+    background: var(--highlighted);
+    border-color: black;
```
fricklerhandwerk, line 496:
```suggestion
```

Nit, but let's not focus on the skin too much for now.

fricklerhandwerk (approved):
Nice!

Issue #417, opened 2024-12-05
Remove package maintainers from issue draft


fricklerhandwerk:
As a security team member, before publishing a security issue, I want to be able to remove maintainers that shouldn't be pinged.

## Acceptance criteria

Given I have a draft issue,
When I remove one of the listed maintainers
Then that maintainer won't be pinged when the issue is published.

Given I have a draft issue and JavaScript is enabled in my browser,
When I remove one of the listed maintainers
Then I can restore them immediately.

Given I have a draft issue,
When I remove one of the listed maintainers
Then that removal is tracked in the activity log for that suggestion.

## Implementation notes

It should look and feel just like for packages.

Depends on:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/416

PR #553, opened 2025-05-14
refactor: add maintainers and maintainers edits to (cached) suggestions

yannham:
First step toward #417. This PR adds a new model `MaintainersEdit`, which corresponds to the addition or the removal of a maintainer from the maintainer list, and a proper field `maintainers` to cached suggestions which aggregates all the maintainers of all the derivations (deduped) and the edits. The full list of maintainers used to be derived in two places beforehand; now it's done once and for all (well, at least as long as there's no edit) and stored in the cached suggestion.

For now, edits can't be made, so they shouldn't impact the list.

**Warning**: to have things work after this PR, one needs to regenerate cached suggestions.


src/website/shared/listeners/cache_suggestions.py
```diff
@@ -295,3 +301,32 @@ def parse_drv_name(name: str) -> tuple[str, str]:
...
+def maintainers_list(packages: dict, edits: list[MaintainersEdit]) -> list[dict]:
+    """
+    Returns a deduplicated list (by GitHub ID) of all the maintainers of all the
+    affected packages linked to this suggestion, modified by potential
+    user-supplied edits.
+    """
+
+    to_remove = {
+        m.maintainer for m in edits if m.type == MaintainersEdit.EditType.REMOVE
+    }
```
fricklerhandwerk, line 315:
```suggestion
    to_remove = [m.maintainer for m in edits if m.type == MaintainersEdit.EditType.REMOVE]
```

Was this the linter trying to be clever?

src/website/shared/listeners/cache_suggestions.py
```diff
@@ -295,3 +301,32 @@ def parse_drv_name(name: str) -> tuple[str, str]:
...
+    to_add = [m.maintainer for m in edits if m.type == MaintainersEdit.EditType.ADD]
```
fricklerhandwerk, line 316: Why is this a list and the other thing a set?

src/website/shared/listeners/cache_suggestions.py
fricklerhandwerk, line 315:
```suggestion
    to_remove = {m.maintainer for m in edits if m.type == MaintainersEdit.EditType.REMOVE}  # noqa
```

Slightly easier on the eye?

src/website/shared/listeners/cache_suggestions.py
```diff
@@ -295,3 +301,32 @@ def parse_drv_name(name: str) -> tuple[str, str]:
...
+        if m["github_id"] not in seen_ids and m["github_id"] not in to_remove:
+            seen_ids.add(m["github_id"])
+            maintainers.append(m)
+
+    for m in to_add:
+        if m.github_id not in seen_ids and m.github_id not in to_remove:
+            seen_ids.add(m.github_id)
+            maintainers.append(m)
+
+    return maintainers
```
fricklerhandwerk, line 332: Why not... (not tested)

```suggestion
    maintainers = {m for pkg in packages.values() for m in pkg["maintainers"]}
    return list(maintainers - to_remove | to_add)
```

src/website/shared/models/linkage.py
```diff
@@ -62,6 +63,26 @@ def create_nixpkgs_issue(self) -> NixpkgsIssue:
...
+        ADD = "add", _("add")
+        REMOVE = "remove", _("remove")
+
+    edit_type = models.CharField(
+        max_length=text_length(EditType), choices=EditType.choices
+    )
+    maintainer = models.ForeignKey(NixMaintainer, on_delete=models.CASCADE)
+    suggestion = models.ForeignKey(
+        CVEDerivationClusterProposal,
+        related_name="maintainers_edits",
```
fricklerhandwerk, line 81: Can we enforce in a constraint that a maintainer edit only appears once on a suggestion? We'd still need to ensure our UI logic won't run into the constraint, but at least the data model would make invalid state impossible to represent and we won't need to do post-checks.

fricklerhandwerk:
Related (but not a blocker): https://github.com/Nix-Security-WG/nix-security-tracker/issues/555

src/website/shared/models/linkage.py
yannham: Do you mean that for example we don't have two edits that add the same maintainer to the same suggestion? Otherwise, reading things literally, I believe `ForeignKey` does enforce that an edit only belongs to one suggestion (it's a one-to-many relation).

src/website/shared/listeners/cache_suggestions.py
yannham: We use them differently: `to_remove` is basically a hashset that we query to know if we should add a maintainer to the working list. I realize, by the way, that it shoud be only github ids and not the whole maintainers. Maybe a better name is `to_skip` and it should be merged with `seen` below; they have the same role.

`to_add` is a list of maintainer to process and add at the end, but we deduplicate already when adding (and we need to do it by github_id, and not on the whole maintainer object as I mentioned below in another comment). We could make `to_add` a set but it's useless. Deduplication happen later.

src/website/shared/listeners/cache_suggestions.py
yannham: This is how I wrote it initially but `ruff` insist on reformatting it this way :slightly_smiling_face:

src/website/shared/listeners/cache_suggestions.py
yannham: I've reworked this part to make it simpler, and relatively idiomatic (and actually fix issues with `dict` vs `non-dict` that were undetected because addition/deletion is just unimplemented for now). There is a comprehension version but I don't like it honestly, it hides a side-effect in an `if`(relying on the fact that `add` returns `None` and this the condition is always `False`):

```
deduped_objects = [obj for obj in objects if obj['id'] not in seen_ids and not seen_ids.add(obj['id'])]
```

I think the current version is reasonably light, but let me know what you think: https://github.com/Nix-Security-WG/nix-security-tracker/pull/553/commits/5b4dd63c2fa73d1879cd09b95ec06175a243b747

fricklerhandwerk: I think one can easily identify values in the set by adding to the `NixMaintainer` model

```python
    def __hash__(self):
        return hash(self.github_id)
        
    def __eq__(self, other):
        if not isinstance(other, Maintainer):
            return False
        return self.github_id == other.github_id
```

yannham: The thing is, at this point, we handle dictionaries, that have been generated by the other parts of suggestion caching (the list of maintainers to edit), so those are dicts and not `NixMaintainer`s. IMHO it's not really worth it to bend the rest of the code just to get a one-liner here, as the function might not be the most elegant, but it's relatively short, simple and understandable.

That being said, maybe we want to add those functions anyway, orthogonally; this probably makes sense as a `hash` and `eq` implementation.

src/website/shared/models/linkage.py
yannham: > Well, it's a one-to-many for edits, not for maintainers in those edits. You could have multiple edits with the same maintainer, right?

Yes, but this is ok... ? You might want to independently add maintainer `@tartempion` to suggestion A and remove it from suggestion B. You initially wrote 

> Can we enforce in a constraint that a maintainer edit only appears once on a suggestion

which I believe is indeed enforced by the foreign key. So, if I understand correctly, the only remaining constraint that isn't enforced and that we want to enforce is the one you said, that several edits to the _same package_ with the _same maintainer_ shouldn't be allowed. Is that correct?

src/website/shared/listeners/cache_suggestions.py
yannham: (as a small aside, it's also less magic - with the custom hash instance, a reader must known that `__hash__` and `__eq__` have be redefined to understand why the code works as intended)

src/website/shared/models/linkage.py
fricklerhandwerk: > Is that correct?

Yes, that's what I meant.

fricklerhandwerk (approved):
Do you think you can still add the relation constraint?

PR #554, opened 2025-05-14
Add remove button (trash) on maintainers + small refactor

yannham:
Depends on #553. This PR refactors the components displaying the list of maintainers by having a dedicated `maintainer` template (instead of inlining this logic in the `maintainer_list`), and makes it possible to have the trash button (selectable). We also avoid having selectable and non-selectable version of the same component (that is, nixpkgs packages and the newly added maintainer) in two separate file, but rather use a context boolean, which deduplicates a bit of code.

Note that nothing is properly handling the form if you click on the trash button for a maintainer; this is planned as future work.


fricklerhandwerk:
Code looks pretty good from a cursory glance, I'll try it out soon(tm). But let's not add broken UI. Let's either skip displaying the new thing and keep it a pure refactor for now, or add the logic already. :)

yannham:
It was mostly for review purpose (small PR) that I split the two, but I can merge this PR and the next one - I'm confident the backend part of it will be relatively small.

src/website/webview/static/style.css
```diff
@@ -850,6 +850,56 @@ a:focus {
...
+  font-size: 1.2em;
+  color: var(--dark-grey);
+}
+
+.maintainer-container input ~ article {
+  margin: 0;
+  padding: 0;
+}
+
+.maintainer-container input:not(:checked) ~ article {
```
fricklerhandwerk, line 898:
```suggestion
.maintainer-container input:not(:checked) ~ article,
.maintainer-container input:not(:checked) ~ article a.github-id
{
```

So the handle also gets greyed out

PR #556, opened 2025-05-15
Implement consistent views for published issues

florentc:
This addresses #544 by implementing:
- a component for issues (as in `NixpkgsIssue`, which differ from "suggestions" that are listed in the other tabs, and is the reason why there was no view until now)
- a paginated list view of published issues (the landing page of "Published issues")
- a detailed view per issue, each having a URL

<details>
<summary>Screenshots</summary>

</details>

### Remarks

- The `NixpkgsIssue` model and the `CveRecord` model do not carry the same info as suggestions used in other tabs, there is not the same degree of continuity as there is between the other tabs. Still, when publishing an issue from a draft, you can directly browse to the issue link.
- In the list view, the "related derivations" are not passed down to the view. This is a first draft only and I'm open to discussing this as I was a bit puzzled about how the info about related derivations is fetched for the detailed view and whether or how we want to lift that to reuse it in list view.
- [ ] As you can see, the maintainers are not yet displayed. There is absolutely no technical difficulty to do that. It just happens that @yannham worked on a refactor of the maintainers list component in a PR of his (the component weirdly takes a list of packages right now and yann is making it take a list of maintainers as expected), so once it is merged it can be reused here to avoid duplication.

### DIFF
--- a/src/website/webview/static/style.css
+++ b/src/website/webview/static/style.css
@@ -365,8 +365,8 @@
   --issue-unknown: #fcee92;
   --issue-affected: #99e87d;
   --issue-notaffected: #e8867d;
-  --issue-notforus: #e2e2e2
-  --issue-wontfix: #ccccdd
+  --issue-notforus: #e2e2e2;
+  --issue-wontfix: #ccccdd;
 }
 
 body {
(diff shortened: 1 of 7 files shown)


fricklerhandwerk (approved):
This is a lot better than before. Let's merge it as is and keep iterating on the details.

PR #560, opened 2025-05-26
refactor: streamline activity log data model

yannham:
This is preliminary work for the follow-up of #558: add maintainer edits to the activity log.

It turns out the activity log main class is trying to accommodate all use-cases by providing a common structure that fits very different things (typically a status update and a package edition update), leading to those kind of unsatisfying fillers and default values:

```
    {'action': 'derivations.remove',    
      'package_count': 1,    
      'package_names': ['{"name": "perl5.36.3-GSSAPI-0.28", "attribute": "perl536Packages.GSSAPI"}'],    
      'status_value': 'NOT_A_STATUS_CHANGE',    
      'suggestion_id': 11,    
      'timestamp': datetime.datetime(2024, 12, 5, 16, 15, 7, 899037, tzinfo=datetime.timezone.utc),    
      'username': 'ANONYMOUS'},    
     {'action': 'update',    
      'package_count': 0,    
      'package_names': [],    
      'status_value': 'accepted',    
      'suggestion_id': 11,    
      'timestamp': datetime.datetime(2024, 12, 5, 16, 14, 19, 520312, tzinfo=datetime.timezone.utc),    
      'username': 'ANONYMOUS'}]
```

This is not great, but adding maintainer edits in the mix - which will have yet a different structure - will just make this horrible. Looking at the code, it seems that one reason for cramming everything into one schema is that it makes it possible to use Django query sets to aggregate and sort all the data together while staying in the SQL world.

This PR instead gathers the data from individual event, and combine them into dicts that may have different shapes (although they have a common frame: `suggestion_id`, `username`, `action`, and `timestamp`, and a varying part depending on `action`). It'll be much simpler to add maintainer edits on top of that without trying to artificially find a schema that would fit all of the three type of edits, with many impossible states.

One consequence is that the sorting can't happen in SQL, and must happen on the Python side instead. This might have performance implications - I couldn't notice a difference during manual testing. I don't expect a page to need to handle a huge volume of log entries, but on the other hand I think removing one package item on the front end can lead to removing hundreds of derivations in the log, so it might not entirely be transparent. In any case, I still expect Python to be able to quickly sort a list with a few thousands elements quickly. If sorting becomes an issue, I'm still convinced that the previous situation wasn't a good solution anyway: in that case, maybe we should cache the activity log, or pre-compute the folding in the database for example. Another solution would be to group events by suggestion_id first and then sort, while we do the contrary here, which is more costly.


fricklerhandwerk:
Side note: we should absolutely cache the activity log

src/website/shared/logs/logs.py
```diff
@@ -100,95 +72,155 @@ def _annotate_username(self, query: EventQuerySet) -> EventQuerySet:
...
+                    "username": package_event.username,
+                    "action": package_event.pgh_label,
+                    "package_names": package_event.package_names,
+                    "package_count": package_event.package_count,
+                }
+            )
+
+        return sorted(raw_events, key=lambda event: event["timestamp"])
 
     def get_dict(
```
fricklerhandwerk, line 153: The schema should simply be e.g. a Pydantic model, which will provide a `model_dump()` method out of the box. Then we don't need any comments to document how it's supposed to work.

fricklerhandwerk (dismissed):

MurdoqZodoq (commented):
She'll

src/website/shared/logs/logs.py
yannham: Ah, great idea, I just forgot that we have unions in Python :sweat_smile:

yannham: Improved in [606de4f](https://github.com/Nix-Security-WG/nix-security-tracker/pull/560/commits/606de4f8474d9d39b6ec1244c72adcc39de45972)

fricklerhandwerk:
We discussed this yesterday, notes:
- Use type constructor when appending to the collection, makes it more obvious what's happening
- Better to type the events in Pydantic directly, since then later we can serialize them with one line into a JSONField for caching
  - There's substantial a risk the activity logs will become a performance/storage liability over time
  - There would be more to do to mitigate it, but caching would be a first step, and here we'd prepare that
- Would also slightly reduce the amount of visual noise in the views
- Generally more types are good for readability, maintainability, robustness

fricklerhandwerk (changes requested):
Ah damn, something about aggregating packages gets broken by this change:

image

May or may not be related to https://github.com/Nix-Security-WG/nix-security-tracker/issues/591

yannham:
Have you regenerated your cached suggestions?

yannham:
Sorry, in fact I'm not sure what got broken? Is that just the names of the packages?

yannham:
Hmm. I can't really see any logical change related to channels here (albeit after just a quick look), so maybe it's worth looking at #591 first and sees if that fixes the issue here. I can't remember if that happened before the rebase or not.

fricklerhandwerk:
Aha it seems the broken aggregation of events has been there for a while. Will investigate further.

PR #562, opened 2025-05-27
feat: add maintainers edit to activity log

yannham:
Depends on #560.

Tracks event related to maintainer edition, and properly display them in the activity log.

Currently, we don't do any aggregation (as is done with packages), because as opposed to packages, maintainers edit are atomic (in the sense that one click actually corresponds to one POST request and thus to one event). Additionally, it's not possible to bulk edits maintainers, while removing one package on the front end can lead to actually removing hundreds of corresponding derivations.

We could aggregate log entries, but we would have to decide:

1. A time threshold for them to be considered simultaneous (or at least aggregating, and for which we could find a single timestamp that would be a reasonable approximation for each event).
2. Should we annihilate dual events (e.g. removing a maintainer and adding it back), or do we consider this to be worth logged even if it's a no-op?

In any case, I think it's not highly important and best left to a follow-up work if needed.

image

### DIFF
--- a/src/webview/templatetags/viewutils.py
+++ b/src/webview/templatetags/viewutils.py
@@ -55,7 +55,7 @@ class SuggestionActivityLog(TypedDict):
 
 class Maintainer(TypedDict):
     name: str
-    email: str
+    email: str | None
     github: str
     matrix: str | None
     github_id: int


fricklerhandwerk:
> * A time threshold for them to be considered simultaneous (or at least aggregating, and for which we could find a single timestamp that would be a reasonable approximation for each event).
> 
> * Should we annihilate dual events (e.g. removing a maintainer and adding it back), or do we consider this to be worth logged even if it's a no-op?

Yes, both. I think this is how we're doing it for removing/undoing-removal-of packages as well. There simply should be a debounce so we don't store garbage data.

src/webview/templates/components/suggestion_activity_log.html
```diff
@@ -51,8 +51,17 @@
...
+          {% if ".add" in log_entry.action %}
+            added
+          {% else %}
+            removed
+          {% endif %}
+          maintainer
+          <strong>@{{ log_entry.maintainer.github }}</strong>
         {% else %}
-          {# Status update entries #}
+        {# Status update entries #}
```
fricklerhandwerk, line 64:
```suggestion
          {# Status update entries #}
```

src/shared/logs/logs.py
```diff
@@ -191,6 +200,27 @@ def get_raw_events(
...
+            raw_events.append(
+                MaintainerChangeEvent(
+                    suggestion_id=maintainer_event.suggestion.id,
+                    timestamp=maintainer_event.pgh_created_at,
+                    username=maintainer_event.username,
+                    action=maintainer_event.pgh_label,
+                    # TODO: we should use Pydantic model for maintainers and
+                    # friends as well, at some point
+                    maintainer=cast(
+                        Maintainer, model_to_dict(maintainer_event.maintainer)
```
fricklerhandwerk, line 219: image

`model_to_dict` will trip up on an unset `email` in the [`Maintainer` model](https://github.com/Nix-Security-WG/nix-security-tracker/pull/562/files#diff-120d76390978d1d6de1ddd431d367056180c1c0e7cb522e53e91bab736fe786bR58)

fricklerhandwerk: Added a quick fix

PR #566, opened 2025-06-03
simplify wiring of secrets and GitHub authentication

fricklerhandwerk:
this is a preliminary to test the UI side of the login workflow, which still
requires a few simplifications to separate the moving parts.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,24 +35,7 @@ python3 -c 'import secrets; print(secrets.token_hex(100))' > .credentials/SECRET
       It will correspond to [`nixpkgs`](https://github.com/nixos/nixpkgs).
       In the **Settings** tab on that repository, in the **Features** section, ensure that _Issues_ are enabled.
 
-2.  For your GitHub user, in **Developer Settings**, generate a new [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
-
-    This is not strictly necessary just to run the service, but allows for more API calls and is therefore important for a production deployment.
-
-    - Generate new token
-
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -35,24 +35,7 @@ python3 -c 'import secrets; print(secrets.token_hex(100))' > .credentials/SECRET
       It will correspond to [`nixpkgs`](https://github.com/nixos/nixpkgs).
       In the **Settings** tab on that repository, in the **Features** section, ensure that _Issues_ are enabled.
 
-2.  For your GitHub user, in **Developer Settings**, generate a new [personal access token](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/managing-your-personal-access-tokens)
```
yannham, line 38: Why is this removed? Doesn't this make it impossible to e.g. create a GH issue in a dev environment?

src/website/shared/github.py
```diff
@@ -20,37 +13,13 @@
...
-    gh_auth: Auth.Auth | None = None
-
-    if credentials_dir is None:
-        logger.warning("No credentials directory available, using unauthenticated API.")
-    else:
-        logger.info(f"Using credentials directory: {credentials_dir}")
-
-        try:
-            with open(f"{credentials_dir}/GH_TOKEN", encoding="utf-8") as f:
-                gh_auth = Auth.Token(f.read().rstrip("\n"))
```
fricklerhandwerk, line 37: @yannham https://github.com/Nix-Security-WG/nix-security-tracker/pull/566/files#r2123702381 this is a noop if `GH_APP_PRIVATE_KEY` is set, which it is in the deployment. So removing the personal access token makes no difference.

PR #569, opened 2025-06-04
Cache issues and show all infos in list view

florentc:
Part of #544 

Previously, showing related derivations on a published issue involved a very long database query. This was bearable for individual issues (issue detail view) but did not scale at all in the issue list view.

This introduces cached issues, similar to cached suggestions. This means there is:

- a database migration that adds cached issues and a hook to populate the cached issue after a new issue is created or modified
- a new `manage` command, `manage regenerate_cached_issues` to regenerate all cached issues. This needs to be ran once to cache issues that were in the db prior to the introduction of the hook @Erethon 

The views are updated accordingly. Now the info about related derivations is available everywhere and the requests take a decent time to process.


src/website/shared/listeners/cache_issues.py
```diff
@@ -0,0 +1,80 @@
...
+        class Maintainer(BaseModel):
+            github: str
+            name: str
+            email: str
+            # email: EmailStr
+
+        name: str
+        maintainers: list[Maintainer]
+
+    status: str
```
fricklerhandwerk: This should be an enum that follows the model

```suggestion
    status: IssueStatus
```

change above

```
from shared.models.cve import NixpkgsIssue, IssueStatus
```

src/website/shared/listeners/cache_issues.py
```diff
@@ -0,0 +1,80 @@
...
+    created_at: date
+    description: str
+    vulnerabilities: list[Vulnerability]
+    related_derivations: list[RelatedDerivation]
+
+
+def cache_new_issue(issue: NixpkgsIssue) -> None:
+    status = issue.get_status_display()  # type: ignore
+    created_at = issue.created
+    description = issue.description.value
```
fricklerhandwerk: I think these can go into the type constructor, they're short enough

src/website/shared/models/cached.py
```diff
@@ -19,3 +20,18 @@ class CachedSuggestions(TimeStampMixin):
...
+    A cached view of published issues
+    """
+
+    issue = models.OneToOneField(
+        NixpkgsIssue,
+        related_name="cached",
+        on_delete=models.CASCADE,
+        primary_key=True,
+    )
+
```
fricklerhandwerk, line 67: Ideally we'd have a way to know whether the stored format is still valid. One way would be adding a positive integer `version` field that defaults to a class attribute of our pydantic schema. As long as contributors don't forget to bump the version when the schema is changed, we'll be able to figure out the rest programmatically

src/website/webview/views.py
```diff
@@ -409,31 +411,20 @@ def get_object(self, queryset: QuerySet | None = None) -> NixpkgsIssue:
...
     template_name = "issue_list.html"
     model = NixpkgsIssue
     paginate_by = 10
 
     def get_queryset(self) -> BaseManager[NixpkgsIssue]:
-        return NixpkgsIssue.objects.all().order_by("-created")
+        issues = NixpkgsIssue.objects.all().order_by("-created")
+        for issue in issues:
+            issue.cached_payload = CachedNixpkgsIssuePayload(**issue.cached.payload)  # type: ignore
+        return issues
```
fricklerhandwerk, line 428: Why not

```suggestion
        return [CachedNixpkgsIssuePayload(**issue.cached.payload) for issue in issues]
```

(and change the return type to the pydantic schema)?

That would avoid a lot of visual noise in the template

src/website/shared/models/cached.py
fricklerhandwerk: We don't have to do anything about the version yet, adding a version and a TODO comment would be enough for now.

CONTRIBUTING.md
```diff
@@ -269,6 +269,19 @@ manage ingest_bulk_cve --subset 100
 This will take a few minutes on an average machine.
 Not passing `--subset N` will take about an hour and produce ~500 MB of data.
 
+### Caching suggestions and issues
+
+Fetching information about individual suggestions or published issues is very frequent when browsing through the security tracker. For each view requested, this would typically involve database queries that would take way too long to execute.
+
+Therefore, we pre-execute these queries and cache the results on every suggestion and published issue. The cached information is used when navigating in the tracker. A django hook on inserting or updating suggestions or published issues ensures caching for new elements but the operation needs to be performed once for pre-existing elements in the database.
```
fricklerhandwerk: Please move this explanatory section into the code. It's important information but not relevant for using the system. We need to watch out not to overload readers who just need to get the job done here.

CONTRIBUTING.md
```diff
@@ -269,6 +269,19 @@ manage ingest_bulk_cve --subset 100
...
+
+To compute or re-compute the cached information from scratch:
```
fricklerhandwerk, line 276: For the contribution guide, it's enough to state the fact and what to do about it.

```suggestion
Contents of published vulnerability records are displayed from a cache to avoid latency from complex database queries.
To compute or re-compute the cached information from scratch:
```

CONTRIBUTING.md
florentc: When we onboarded on the project with @yannham, we missed having the information about generating cached suggestions though.

fricklerhandwerk: Yeah yeah the instructions should be here. Just not the explanation on why things are built the way they are! That's for in-code comments :)

src/website/webview/views.py
florentc: The view uses the db key for issues as a title and it is not part of the pydantic schema. It used to be but it turns out it is only present when manually generating cached issues, not when it happens via the hook.

Would you like to return pairs of pk/cached issue in the view?

CONTRIBUTING.md
florentc: In your suggestion, you add a sentence (new line 278) but I feel like it is intended to replace the paragraph in line 276, do you confirm? Also we should mention suggestions.

Would you like to replace the entire section with:

```markdown
Suggestion and issue contents are displayed from a cache to avoid latency from complex database queries.

To compute or re-compute the cached information from scratch:
```

src/website/shared/listeners/cache_issues.py
florentc: Fixed by https://github.com/Nix-Security-WG/nix-security-tracker/pull/569/commits/7c41feb736238712c205c7f55db7d49df9688075

src/website/shared/listeners/cache_issues.py
florentc: Fixed by https://github.com/Nix-Security-WG/nix-security-tracker/pull/569/commits/eb7b5b127c1d7ad03656dc94dc568717449d4e65

src/website/shared/models/cached.py
florentc: Fixed by https://github.com/Nix-Security-WG/nix-security-tracker/pull/569/commits/5a3289eece4ce44a9ef0e68024d8028570201777

CONTRIBUTING.md
fricklerhandwerk: Yes exactly

src/website/webview/views.py
fricklerhandwerk: > it turns out it is only present when manually generating cached issues, not when it happens via the hook

Now this is weird!

> Would you like to return pairs of pk/cached issue in the view?

Not sure. If that works for you? If you have a way to be milder on the eyes in the template, I'm fine with that. But we can also leave a TODO for later and merge ahead.

fricklerhandwerk (approved):
Perfect

src/website/webview/views.py
florentc: In fact the code is not the pk.

I think it's because the code is generated post save. In `cve.py`

```python
@receiver(post_save, sender=NixpkgsIssue)
def generate_code(
    sender: type[NixpkgsIssue], instance: NixpkgsIssue, created: bool, **kwargs: Any
) -> None:
    if created:
        number = sender.objects.filter(
            created__year=instance.created.year, pk__lte=instance.pk
        ).count()
        instance.code = f"NIXPKGS-{str(instance.created.year)}-{str(number).zfill(4)}"
        instance.save()
```

While in `cache_issues.py`

```python
@pgpubsub.post_insert_listener(NixpkgsIssueChannel)
def cache_new_issue_following_insert(
    old: NixpkgsIssue, new: NixpkgsIssue
) -> None:
    cache_new_issue(new)
```

So the code doesn't exist yet.

I tried to use post_save too for caching new issues but it did not work.

I also tried to cache issues with a `@receiver(post_save, sender=NixpkgsIssue)` but I could no longer even create issues because it complained about duplicates.

There is certainly an elegant solution but it will need more investigation so for the time being in this PR I'll add a todo.

fricklerhandwerk (approved):

PR #570, opened 2025-06-05
Show current git revision on every page

### LINKED ISSUE #358: Show the version we're running
As a user of or contributor to the security tracker, I want to be able to check which version is currently running in production in order to know whether a given change is already deployed.

## Acceptance criteria

Given the security tracker runs on some machine,
When I call the appropriate endpoint (could be any page or a certain page),
Then it shows the Git revision it's running.

florentc:
This fixes #358. It mentions the current git revision in the footer of every page.

### DIFF
--- a/src/website/shared/templates/base.html
+++ b/src/website/shared/templates/base.html
@@ -81,6 +81,7 @@
         <a href="https://github.com/Nix-Security-WG/nix-security-tracker">Nixpkgs Security Tracker</a> is part of a project funded by the
         <a href="https://sovereigntechfund.de/en/">Sovereign Tech Fund</a>.
       </p>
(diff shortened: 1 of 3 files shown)


fricklerhandwerk:
This *should* work in production (fingers crossed to a hexagon)

florentc:
@fricklerhandwerk I see you have reverted to using `fetchGit`. I thought it was something @balsoft mentioned we should try to avoid doing (hence my attempt at running `git rev-parse` in prod, and fetching from `.git/ref/HEAD` in dev shell).

Shouldn't we at least add a TODO about it?

In any case, whatever we do, it seems we'll need to investigate further because in staging it doesn't show the revision.

PR #580, opened 2025-06-11
Reuse package list view for issues

### LINKED ISSUE #574: Collapse packages per vulnerability
Currently it looks like this

Image

But we want it to be like that:

Image

florentc:
Fixes #574 

This is a draft and very naive adaptation (not ready for merge) of reusing the package list component from suggestions for issues. It needs to adapt the content of the cached issues.

### DIFF
--- a/src/website/shared/listeners/cache_issues.py
+++ b/src/website/shared/listeners/cache_issues.py
@@ -1,10 +1,13 @@
 import logging
 
 import pgpubsub
+from django.db.models import Prefetch
 
 from shared.channels import NixpkgsIssueChannel
+from shared.listeners.cache_suggestions import channel_structure
 from shared.models.cached import CachedNixpkgsIssue, CachedNixpkgsIssuePayload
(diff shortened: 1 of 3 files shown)


florentc:
@fricklerhandwerk The PR was draft because there were remaining things to be done before merge.

- [ ] In particular, we need to remove the former `related_derivations` field from the cached issue pydantic schema (that is no longer used).
- [ ] We need to assert whether we can/want an actual detailed type for the `packages` field: that is return type for the `channel_structure` (which for now was dict)
- [ ] Maybe move `channel_structure` outside cache suggestions because it is used both for caching suggestions, and now caching issues
- [ ] Also, we needed to bump the schema version number for the "packages" field.

Rght now, the issues are broken in staging because there needs to be a new regenerate_cached_issues. However, I think we should first at least bump the version number of the schema before we do it.

fricklerhandwerk:
Yup sorry, or just looked too good... :)

PR #586, opened 2025-06-16
ci: bump nixpkgs and use it everywhere

### LINKED ISSUE #584: Bump Nixpkgs to 25.05
We've been avoiding bumping Nixpkgs so we don't have to deal with looming build failures, but now before production deployment it's finally time.

- [ ] Update dependencies
- [ ] Fix build failures
- [ ] Let the auto-update run for the regular dependencies (currently it's a no-op since it's only bumping dead code)

fricklerhandwerk:
closes #584

historically there was a split between sources for development and
deployment. this had never been resolved, because up to some point in
the past, bumping nixpkgs would result in nasty build failures.
apparently these have been resolved.

now that we keep nixpkgs at unstable, there's no acute reason to have
two environments.


fricklerhandwerk:
@Erethon yeah it's not helping, could you please massage the deployment into shape?

Erethon:
I applied this manually and fixed a bug that popped in our usage of PyGitHub ([changelog](https://pygithub.readthedocs.io/en/latest/changes.html?highlight=get_organization#others), [commit](https://github.com/Nix-Security-WG/nix-security-tracker/pull/586/commits/6ff724c3cb330e51c2cb841e27087795e13119d1)). Everything works fine now as far as I can see.

Erethon (approved):

PR #588, opened 2025-06-19
feat: redirect to same page after login

fricklerhandwerk:


src/webview/tests.py
```diff
@@ -1 +1,53 @@
...
+from django.test import TestCase
+from django.urls import reverse
+
+
+class Login(TestCase):
+    def setUp(self) -> None:
+        # location to start navigation
+        self.entry_point = reverse("webview:suggestions_view")
+
+    def test_foo(self) -> None:
```
Erethon: A more descriptive name for this test would be better.

Issue #418, opened 2024-12-05
Add maintainers to an issue draft


fricklerhandwerk:
As a security team member, I want to be able to select *additional maintainers* to be notified when a security issue is published. Those may be others than just the original maintainers of the affected packages.

## Acceptance criteria

Given I have a draft issue,
When I search for a maintainer among the collection of all known maintainers, and add them to the draft issue,
Then that maintainer will be pinged when the issue is published.

Given I have a draft issue,
When I add an extra maintainer to be pinged,
Then that change will be recorded in the activity log.

## Implementation notes

We may want to display additional pings in a separate list for clarity.

Depends on:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/416

PR #593, opened 2025-06-26
Add extra maintainers from database or GitHub in issue drafts

florentc:
This addresses #418 and implements an input field for extra maintainers in issue drafts. When not found in database, it fetches the info from GitHub via user handle (exact match).

### DIFF
--- a/src/shared/github.py
+++ b/src/shared/github.py
@@ -136,3 +136,20 @@ def get_maintainer_username(maintainer: dict, github: Github = get_gh()) -> str:
             f"Couldn't retrieve the GitHub username for maintainer {maintainer['github_id']}, fallback to {maintainer['github']}: {e}"
         )
         return maintainer["github"]
+
+
+def fetch_user_info(github_handle: str, github: Github = get_gh()) -> dict | None:
+    """
+    Fetch GitHub user info by handle.
+    """
+    try:
+        user = github.get_user(github_handle)
+        return {
+            "id": user.id,
+            "login": user.login,
+            "name": user.name,
+            "email": user.email,
+        }
+    except Exception as e:
+        logger.error(f"Could not fetch GitHub user '{github_handle}': {e}")
+        return None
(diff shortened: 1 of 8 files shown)


src/webview/tests.py
```diff
@@ -51,3 +57,117 @@ def mock_login(
...
+        self.assigner = Organization.objects.create(uuid=1, short_name="foo")
+        self.cve_record = CveRecord.objects.create(
+            cve_id="CVE-2025-0001",
+            assigner=self.assigner,
+        )
+        self.suggestion = CVEDerivationClusterProposal.objects.create(
+            status=CVEDerivationClusterProposal.Status.PENDING,
+            cve_id=self.cve_record.id,
+        )
+        self.cached = CachedSuggestions.objects.create(
```
fricklerhandwerk: Since we're at it, can we maybe go the whole hog and build the `DerivationClusterFProposal` with an entire derivation that has a maintainer, compile the cached suggestion, and only then test the interaction? That would increase line coverage a lot in this one change.

src/webview/tests.py
```diff
@@ -51,3 +57,117 @@ def mock_login(
...
+                url,
+                {
+                    "suggestion_id": self.suggestion.pk,
+                    "new_maintainer_github_handle": "newuser",
+                },
+            )
+        self.assertEqual(response.status_code, 200)
+        # Reload cached suggestion and check maintainers
+        self.cached.refresh_from_db()
+        maintainers = self.cached.payload["maintainers"]
```
fricklerhandwerk: Maybe we should also check that we're not failing at cache invalidation and compare cached values with database contents.

src/webview/tests.py
florentc: https://github.com/Nix-Security-WG/nix-security-tracker/pull/593/commits/4f262dc3bdc9276a5923e293d463609a221e5a56

src/webview/tests.py
florentc: https://github.com/Nix-Security-WG/nix-security-tracker/pull/593/commits/f1871640907dcf42b87e78751c51cd5a3a71741d

PR #597, opened 2025-07-03
Hide maintainer edition widgets when logged out or unauthorized

florentc:
Maintainer widgets (add / remove / restore) would be shown even when logged out. This fixes it and add regression tests.

### DIFF
--- a/src/webview/tests.py
+++ b/src/webview/tests.py
@@ -2,6 +2,7 @@ from typing import Any
 from unittest.mock import patch
 
 from allauth.account.utils import get_login_redirect_url
+from allauth.socialaccount.models import SocialAccount
 from allauth.socialaccount.providers.oauth2.views import OAuth2LoginView
 from allauth.socialaccount.templatetags.socialaccount import provider_login_url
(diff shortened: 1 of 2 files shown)


src/webview/tests.py
```diff
@@ -252,3 +262,16 @@ def test_add_maintainer_invalid_github_handle(self) -> None:
         self.assertEqual(
             response.context["error_msg"], "Could not fetch maintainer from GitHub"
         )
+
+    def test_add_maintainer_widget_visible_when_admin(self) -> None:
+        url = reverse("webview:drafts_view")
+        response = self.client.get(url)
+        response_content = response.content.decode("utf-8")
+        self.assertIn("maintainer-add-container", response_content)
```
fricklerhandwerk, line 270: I'm slightly uneasy about checking rendered response content against a string that we define in the template, especially because refactoring the template will require changing the test. But everything else I can come up with is more complicated, so it's probably fine.

fricklerhandwerk (approved):

Issue #592, opened 2025-06-23
Undoing package removal doesn't work


fricklerhandwerk:
This seems to be a regression in the view logic: Removing a package and undoing it leaves the package removed. There's nothing in the log indicating an error.

![Image](https://github.com/user-attachments/assets/569a2016-40c4-49a4-8535-cf92d58797a0)

Better test that properly.

PR #599, opened 2025-07-09
Fix undoing package removal not working

florentc:
This aims at fixing #592 

For now the PR only provides a test suite for the feature, including a test that shows the issue.

This implements package edits akin to maintainers edits, along with a new `original_packages` field in cached suggestion.

This means this PR requires to apply a migration and regenerate the cached suggestions.


src/webview/tests.py
```diff
@@ -36,6 +36,267 @@
...
+                head_sha1_commit="deadbeef",
+                state=NixChannel.ChannelState.STABLE,
+                release_version="24.05",
+                repository="https://github.com/NixOS/nixpkgs",
+            ),
+            commit_sha1="deadbeef",
+            state=NixEvaluation.EvaluationState.COMPLETED,
+        )
+
+        # Create two derivations for the same package
```
fricklerhandwerk:
```suggestion
```

artefact or typo? the two derivations are simply being linked to the same suggestion

src/webview/tests.py
```diff
@@ -36,6 +36,267 @@
...
+        # Verify derivation2 has been removed from the suggestion
+        self.suggestion.refresh_from_db()
+        remaining_derivation_ids = set(
+            self.suggestion.derivations.values_list("id", flat=True)
+        )
+        self.assertIn(self.derivation1.id, remaining_derivation_ids)
+        self.assertNotIn(self.derivation2.id, remaining_derivation_ids)
+
+        # Restore package2 by including both derivation IDs again
+        restore_response = self.client.post(
```
fricklerhandwerk, line 244: Can we maybe here somewhat tidily obtain the undo action URL from the response, to make sure it's actually made available in the UI? One way that avoids parsing the entire response HTML would be executing the template tag on the response context

src/shared/listeners/cache_suggestions.py
```diff
@@ -110,10 +128,12 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
     )
 
     prefetched_metrics = Metric.objects.filter(container__cve=suggestion.cve)
-    packages = channel_structure(all_versions, derivations)
+    original_packages = channel_structure(all_versions, derivations)
```
fricklerhandwerk, line 131: Note from review meeting: This is the same but has a different name as `packages` in the underlying model

src/shared/models/linkage.py
```diff
@@ -93,6 +93,34 @@ class Meta:  # type: ignore[override]
...
+    """
+
+    class EditType(models.TextChoices):
+        REMOVE = "remove", _("remove")
+        # ADD reserved for future use if needed
+
+    edit_type = models.CharField(
+        max_length=text_length(EditType), choices=EditType.choices
+    )
+    package_attribute = models.CharField(max_length=255)
```
fricklerhandwerk, line 108: We may want to go a step further and having a dedicated package model that will bundle derivations and channels and whatnot. Especially since attribute names may change over time, and what we really want is a stable identifier.

src/shared/listeners/cache_suggestions.py
florentc: I got it mixed up, in fact there is neither packages nor original_packages in the underlying model so it's all fine

PR #601, opened 2025-07-09
infra: Optimise the nix store by default

Erethon:
Automatically [optimise the nix store](https://wiki.nixos.org/wiki/Storage_optimization#Automatic) by using a timer. We could argue that this should be set in `nix/web-security-tracker.nix`, but I don't want to "force" this on all installations and I think its best to allow admins to make this decision themselves. This reasoning shall be documented in #594.


fricklerhandwerk:
Side note: please add the "why" to the commit message as well, I squashed it in this time based on the PR description.

Issue #438, opened 2024-12-09
Restore removed packages from the activity log


fricklerhandwerk:
As a security team member, when assessing the activity log for a suggestion (or issue draft), it could happen that I notice some package was removed but shouldn't have been. I want to be able to restore it by selecting it right were I find it.

PR #624, opened 2025-07-23
Fix broken activity log after implementation of new PackageEdit paradigm

florentc:
Following the implementation of a new paradigm for editing packages in a suggestion, akin to editing maintainers in a way that leaves the suggestions immutable, the activity log was no longer reporting those package changes events.

This PR implements the necessary changes to get back to a working activity log.

Event folding is still working like before for packages.

This is intended to be merged AFTER #562 (as it uses it as a base branch for this PR)

Regarding remaining activity log user friendly features (grouping maintainers edits, implementing debouncing to hide close "delete + restore" events), I plan them for a future PR.

### DIFF
--- a/src/webview/tests.py
+++ b/src/webview/tests.py
@@ -13,7 +13,6 @@ from django.test import Client, TestCase
 from django.urls import reverse
 
 from shared.listeners.cache_suggestions import cache_new_suggestions
-from shared.logs import SuggestionActivityLog
 from shared.models.cve import (
     AffectedProduct,
     CveRecord,
@@ -26,7 +25,6 @@ from shared.models.linkage import (
     CVEDerivationClusterProposal,
     DerivationClusterProposalLink,
     MaintainersEdit,
-    PackageEdit,
     ProvenanceFlags,
 )
 from shared.models.nix_evaluation import (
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
This definitely needs a test though.

src/webview/tests.py
```diff
@@ -496,3 +498,695 @@ def test_add_maintainer_widget_not_visible_when_logged_out(self) -> None:
         response = self.client.get(url)
         response_content = response.content.decode("utf-8")
         self.assertNotIn("maintainer-add-container", response_content)
+
+
+class PackageEditActivityLogTests(TestCase):
+    def setUp(self) -> None:
```
fricklerhandwerk, line 502: We may want to adopt e.g. https://factoryboy.readthedocs.io to produce these objects.

src/webview/tests.py
```diff
@@ -496,3 +498,695 @@ def test_add_maintainer_widget_not_visible_when_logged_out(self) -> None:
...
+        # Should have one event for package removal
+        package_events = [
+            e for e in events if e.get("action", "").startswith("package.")
+        ]
+        self.assertEqual(len(package_events), 1)
+
+        removal_event = package_events[0]
+        self.assertEqual(removal_event["action"], "package.remove")
+        self.assertEqual(removal_event["package_attribute"], "package2")
+        self.assertEqual(removal_event["username"], "admin")
```
fricklerhandwerk: Maybe we shouldn't test internals here? I imagine these tests to nail down surface features visible to users. While it's surely cleaner to look at the database, it's more important that we're covering things end-to-end. But I'm open for discussion here.

```suggestion
```

src/webview/tests.py
```diff
@@ -496,3 +498,695 @@ def test_add_maintainer_widget_not_visible_when_logged_out(self) -> None:
...
+        url = reverse("webview:suggestions_view")
+        self.client.post(
+            url,
+            {
+                "suggestion_id": self.suggestion.pk,
+                "attribute": ["package1"],
+            },
+        )
+
+        # Then restore package2 by including both packages
```
fricklerhandwerk, line 668: There's no UI for that yet, except through the undo button, right? We probably want to be explicit here how the restoration works, i.e. go through the output of this view.


This test might come in handy for https://github.com/Nix-Security-WG/nix-security-tracker/issues/438

src/webview/tests.py
```diff
@@ -496,3 +498,695 @@ def test_add_maintainer_widget_not_visible_when_logged_out(self) -> None:
...
+        self.client.post(
+            url,
+            {
+                "suggestion_id": self.suggestion.pk,
+                "attribute": ["package1", "package2"],
+            },
+        )
+
+        # Check that no PackageEdit entries exist (removal was undone)
```
fricklerhandwerk: Same as above, probably should not test internals?

PR #629, opened 2025-07-25
Activity log: hide close events that cancel each other

florentc:
This is based on #624 so it should be merged after.

This implements a feature that hides event that cancel each other when they are consecutive and happened within 5min from each other.

For instance add a maintainer then remove it. Remove a package then restore it.

For now the implementation is very basic: events are only cancelled out if consecutive. If you start to interlace them then they will all show up in the rendered activity log. But this fits the main use case: when you click on add/remove package/maintainer then undo.


src/shared/logs/logs.py
```diff
@@ -149,50 +152,106 @@ def get_raw_events(
...
             )
 
         return sorted(raw_events, key=lambda event: event.timestamp)
 
+    def _remove_canceling_events(
+        self,
+        events: list[
+            PackageChangeEvent | SuggestionChangeEvent | MaintainerChangeEvent
+        ],
+        time_threshold_seconds: int = 300,
```
fricklerhandwerk:
```suggestion
        time_threshold_seconds: int = 30,
```

5 min is probably too wide of a window.

src/shared/logs/logs.py
```diff
@@ -149,50 +152,106 @@ def get_raw_events(
...
+        if (
+            event1.username != event2.username
+            or event1.suggestion_id != event2.suggestion_id
+            or (event2.timestamp - event1.timestamp).total_seconds()
+            > time_threshold_seconds
+        ):
+            return False
+
+        # Check for canceling pairs
+        if isinstance(event1, MaintainerChangeEvent) and isinstance(
```
fricklerhandwerk, line 250: How about instead implementing a comparison operator on the class?

src/shared/logs/logs.py
```diff
@@ -149,50 +152,106 @@ def get_raw_events(
...
+        self,
+        event1: PackageChangeEvent | SuggestionChangeEvent | MaintainerChangeEvent,
+        event2: PackageChangeEvent | SuggestionChangeEvent | MaintainerChangeEvent,
+        time_threshold_seconds: int,
+    ) -> bool:
+        """Check if two consecutive events cancel each other."""
+        # Same user, same suggestion, within time threshold
+        if (
+            event1.username != event2.username
+            or event1.suggestion_id != event2.suggestion_id
```
fricklerhandwerk, line 243: The way we're using it would always give the same suggestion. Is this simply defensive programming?

src/shared/logs/logs.py
```diff
@@ -204,11 +263,15 @@ def get_dict(
 
         raw_events = self.get_raw_events(suggestion_ids)
 
+        # Cancellation pass - remove events that cancel each other within a given time window (5 min by default)
+        filtered_events = self._remove_canceling_events(raw_events)
+
         grouped_activity_log: dict[
-            int, list[PackageChangeEvent | SuggestionChangeEvent]
+            int,
+            list[PackageChangeEvent | SuggestionChangeEvent | MaintainerChangeEvent],
```
fricklerhandwerk: As often as we use the type, maybe let's aggregate this in a class (or simply a type that is made up of this union)?

src/webview/tests.py
```diff
@@ -620,6 +622,31 @@ def setUp(self) -> None:
...
+        url = reverse("webview:suggestions_view")
+        self.client.post(
+            url,
+            {
+                "suggestion_id": self.suggestion.pk,
+                "attribute": ["package1"],
+            },
+        )
+
+        with patch(
```
fricklerhandwerk: For readability and making it more obvious that changing the test doesn't break existing behavior, I'd recommend against extracting this procedure and instead
- change the existing test to use the `with patch()` so it's clear the behavior is preserved if the time delta is right
- add another test with a smaller delta and otherwise copypasta, but the changed observation that the activity log is now collapsed

This would also make obvious which interactions may indeed need abstraction. I don't think that "add and then remove X" is a meaningful simplification here. Instead it should be more along the lines of what users would actually do: [click that button to] trigger some action, then [click the resulting button to] trigger the next action. Then test cases would pretty much read like user instructions. But that sort of refactoring can come when it starts hurting the eyes. Extracting a pattern for <3 instances is typically premature abstraction.

src/shared/logs/logs.py
florentc: You mean a comparison to check if two events are self-cancellable (in which case it would be weird IMO because it would not be reflexive or transitive) or that they have same type?

src/shared/logs/logs.py
florentc: Yes totally. For now it's a method in `SuggestionActivityLog` but nothing type-wise guarantees that and it seems to me like an expected implicit check given the function name/doc/type signature. I would find it weird to remove the guard just because we know that we would compare events local to a suggestion, especially since this guard is a small trivial O(1) and doesn't cost us anything. Besides, if someday we extract this method to be a standalone function for events in general (for instance if we implement a view for a global activity log) this could bite us back in the face.

src/webview/tests.py
florentc: This was intended as deduplicating, not abstracting (I agree that remove & restore is not meaningful) to make the 4 test scenarios where those are used more readable. I am not sure I understood what you would like to achieve.

Do you want me to remove `_perform_package_removal_and_restoration`/`_perform_maintainer_removal_and_restoration` and inline them at their call sites?

src/shared/logs/logs.py
fricklerhandwerk: That they cancel, which is a superset of having the same type.

src/webview/tests.py
fricklerhandwerk: > Do you want me to remove ...

Yes. Then changes to existing tests will be obvious (currently I can barely make out whether they stayed the same) and future refactorings will be as well. And as mentioned, we can still collapse the verbose explicit notation into a method such as `self.remove_package(self.example_package)`.

Issue #502, opened 2025-04-21
Remove release branch from a suggestion


fricklerhandwerk:
As a security team member, I want to be able to remove a release branch (channel) from a suggestion, to signify that a given vulnerability does not affect it.

## Acceptance criteria

Given I have a matching suggestion with some affected channels,
When I remove a channel,
Then it will not show up in the published issue.

Given I have a matching suggestion with some affected channels, and I have JavaScript enabled in my browser,
When I remove a channel,
Then I can bring it back immediately, before changing the state of that suggestion.

Given I have a matching suggestion with some affected channels,
When I remove a channel,
Then that change will be recorded in the activity log for that suggestion.

fricklerhandwerk:
As a follow-up we may want to have any of the following:
- Put back a channel into a suggestion outside of an undo action
  - see https://github.com/Nix-Security-WG/nix-security-tracker/issues/632#issuecomment-3144768881 for details
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/618

PR #631, opened 2025-07-30
Activity log overhaul & maintainers edit folding

florentc:
This refactors the event management into something saner and more adaptable to future features (no longer suggestion log centric, could be any activity log through implementation of other "fetchers" such as user-centric logs or global logs). This also implements folding maintainer edits.

### DIFF
--- a/src/shared/logs/batches.py
+++ b/src/shared/logs/batches.py
@@ -1,3 +1,4 @@
+from abc import ABC
 from datetime import datetime
 from typing import Literal
 
@@ -13,7 +14,7 @@ from shared.logs.events import (
 )
 
 
-class FoldedEvent(BaseModel):
+class FoldedEvent(BaseModel, ABC):
     """Base class for folded events that can represent single or bulk operations."""
 
     suggestion_id: int
@@ -25,6 +26,8 @@ class FoldedStatusEvent(FoldedEvent):
     """A folded status change event (always singular)."""
 
     action: Literal["insert", "update"]
+    # TODO This should eventually be restricted to the few literal values a
+    # suggestion status can have
     status_value: str
 
 
@@ -46,29 +49,20 @@ FoldedEventType = FoldedStatusEvent | FoldedPackageEvent | FoldedMaintainerEvent
 
 
 def batch_events(
-    events: list[RawEventType], pre_sort: bool = False
+    events: list[RawEventType], sort: bool = False
 ) -> list[FoldedEventType]:
     """
     Batch consecutive events of the same type from the same user into bulk operations.
-    Events must be sorted chronologically. Use the pre_sort flag if not.
+    Events must be sorted chronologically. Use the sort flag if not.
     """
     folded_events = []
     accumulator = None
 
-    events = sort_events_chronologically(events) if pre_sort else events
+    events = sort_events_chronologically(events) if sort else events
 
     for event in events:
         if isinstance(event, RawPackageEvent):
-            if not accumulator:
-                # Start new package accumulator
-                accumulator = FoldedPackageEvent(
-                    suggestion_id=event.suggestion_id,
-                    timestamp=event.timestamp,
-                    username=event.username,
-                    action=event.action,
-                    package_names=[event.package_attribute],
-                )
-            elif (
+            if (
                 isinstance(accumulator, FoldedPackageEvent)
                 and event.action == accumulator.action
                 and event.username == accumulator.username
@@ -79,7 +73,8 @@ def batch_events(
                 accumulator.timestamp = event.timestamp
             else:
                 # End current accumulator, start new one
-                folded_events.append(accumulator)
+                if accumulator:
+                    folded_events.append(accumulator)
                 accumulator = FoldedPackageEvent(
                     suggestion_id=event.suggestion_id,
                     timestamp=event.timestamp,
@@ -89,16 +84,7 @@ def batch_events(
                 )
 
         elif isinstance(event, RawMaintainerEvent):
-            if not accumulator:
-                # Start new maintainer accumulator
-                accumulator = FoldedMaintainerEvent(
-                    suggestion_id=event.suggestion_id,
-                    timestamp=event.timestamp,
-                    username=event.username,
-                    action=event.action,
-                    maintainers=[event.maintainer],
-                )
-            elif (
+            if (
                 isinstance(accumulator, FoldedMaintainerEvent)
                 and event.action == accumulator.action
                 and event.username == accumulator.username
@@ -109,7 +95,8 @@ def batch_events(
                 accumulator.timestamp = event.timestamp
             else:
                 # End current accumulator, start new one
-                folded_events.append(accumulator)
+                if accumulator:
+                    folded_events.append(accumulator)
                 accumulator = FoldedMaintainerEvent(
                     suggestion_id=event.suggestion_id,
                     timestamp=event.timestamp,
@@ -136,7 +123,6 @@ def batch_events(
                     )
                 )
 
-    # Don't forget the final accumulator
     if accumulator:
         folded_events.append(accumulator)
 
--- a/src/shared/logs/events.py
+++ b/src/shared/logs/events.py
@@ -25,7 +25,7 @@ class RawEvent(BaseModel, ABC):
     def precedes_close_related_event(
         self, other: "RawEvent", time_threshold_seconds: int = 30
     ) -> bool:
-        """Checks the event is followed by one related to the same suggestion by the same user within a give time window"""
+        """Checks if the event is followed by one related to the same suggestion by the same user within a given time window"""
         return (
             self.username == other.username
             and self.suggestion_id == other.suggestion_id
@@ -113,16 +113,16 @@ def sort_events_chronologically(events: list[RawEventType]) -> list[RawEventType
 
 
 def remove_canceling_events(
-    events: list[RawEventType], time_threshold_seconds: int = 30, pre_sort: bool = False
+    events: list[RawEventType], time_threshold_seconds: int = 30, sort: bool = False
 ) -> list[RawEventType]:
     """
     Remove consecutive events that cancel each other out within a time window.
-    Events must be sorted chronologically. Use the pre_sort flag if not.
+    Events must be sorted chronologically. Use the sort flag if not.
     """
     filtered_events = []
     i = 0
 
-    events = sort_events_chronologically(events) if pre_sort else events
+    events = sort_events_chronologically(events) if sort else events
 
     while i < len(events):
         if i + 1 < len(events) and events[i].is_canceled_by(
--- a/src/shared/logs/fetchers.py
+++ b/src/shared/logs/fetchers.py
@@ -52,6 +52,9 @@ def _annotate_username(query: EventQuerySet) -> EventQuerySet:
     )
 
 
+# TODO In the future we may add other fetchers, for instance fetch all events
+# of a given user (for personal dashboard or admin panel) or all events
+# globally (for admin)
 def fetch_suggestion_events(suggestion_id: int) -> list[RawEventType]:
     """Fetch all raw events for a suggestion and return them sorted by timestamp."""
     all_events: list[RawEventType] = []
--- a/src/webview/views.py
+++ b/src/webview/views.py
@@ -526,7 +526,7 @@ class SuggestionListView(ListView):
         for obj in context["object_list"]:
             raw_events = fetch_suggestion_events(obj.proposal_id)
             obj.activity_log = batch_events(
-                remove_canceling_events(raw_events, pre_sort=True)
+                remove_canceling_events(raw_events, sort=True)
             )
 
         context["adjusted_elided_page_range"] = context[
@@ -560,7 +560,7 @@ class SuggestionListView(ListView):
 
         # Activity log
         raw_events = fetch_suggestion_events(suggestion.pk)
-        activity_log = batch_events(remove_canceling_events(raw_events, pre_sort=True))
+        activity_log = batch_events(remove_canceling_events(raw_events, sort=True))
 
         cached_suggestion = get_object_or_404(
             CachedSuggestions, proposal_id=suggestion_id


florentc:
This is ready for review @fricklerhandwerk 

I'm having a few remarks about the approach (this is a big refactor but it doesn't change existing features):
- We have a time window to have events cancel each other or not, which makes sense. I am wondering if we should have the same for event batching.
- When events are batched / folded, we only keep the most recent timestamp, but I feel uncomfortable losing the info. I think we should still keep track of the previous timestamps and show them in the detailed view.

Those would belong in other PRs for other user stories though.

fricklerhandwerk:
> We have a time window to have events cancel each other or not, which makes sense. I am wondering if we should have the same for event batching.

Yes, would be reasonable.

> When events are batched / folded, we only keep the most recent timestamp, but I feel uncomfortable losing the info. I think we should still keep track of the previous timestamps and show them in the detailed view.

Since all this is happening at the presentation layer, we should look into it through the lens of user needs.

> Those would belong in other PRs for other user stories though.

Yes, absolutely.

src/shared/logs/collections.py
```diff
@@ -0,0 +1,51 @@
...
+        """Initialize with a list of raw events, optionally skipping sort if already sorted."""
+        if _skip_sort:
+            self._events = events
+        else:
+            self._events = sorted(events, key=lambda event: event.timestamp)
+
+    @property
+    def events(self) -> list[RawEventType]:
+        """Get the current list of events."""
+        return self._events.copy()
```
fricklerhandwerk: Side note: the underlying list elements are still mutable, so if used carelessly this stays a footgun. I doubt it's worth going full pedantic mode to prevent this. Also since it's an internal interface I would even doubt it's worth being very OOPy here: we could just as well use the class as a means of code organisation, and simply store the list of `RawEvent`, access it directly when needed, and have methods that will return it filtered for various purposes.

src/shared/logs/collections.py
```diff
@@ -0,0 +1,51 @@
...
+        """Return the number of events in the collection."""
+        return len(self._events)
+
+    def __iter__(self) -> Iterator[RawEventType]:
+        """Make the collection iterable."""
+        return iter(self._events)
+
+    def remove_canceling_events(
+        self, time_threshold_seconds: int = 30
+    ) -> "EventCollection":
```
fricklerhandwerk:
```suggestion
    ) -> list[RawEventType]:
```

could also do it

src/shared/logs/raw_events.py
```diff
@@ -0,0 +1,213 @@
...
+class RawStatusEvent(RawEvent):
+    """Raw status change event."""
+
+    action: Literal["insert", "update"]
+    status_value: str
+
+    def is_canceled_by(
+        self, other: "RawEvent", time_threshold_seconds: int = 30
+    ) -> bool:
+        """Status events don't cancel each other currently."""
```
fricklerhandwerk:
```suggestion
        """Status events don't cancel each other."""
```

"currently" is always the case for the given piece of code

src/shared/logs/raw_events.py
```diff
@@ -0,0 +1,213 @@
...
+        """Check if this event is canceled by another event.
+
+        Must be implemented by subclasses to define specific cancellation logic.
+        """
+        pass
+
+    def _basic_cancellation_check(
+        self, other: "RawEvent", time_threshold_seconds: int = 30
+    ) -> bool:
+        """Helper method for common cancellation checks."""
```
fricklerhandwerk: This check is not generic, it means something very specific and should be captured in the name. Dumb suggestion:

```suggestion
    def user_interaction_in_short_succession(
        self, other: "RawEvent", time_threshold_seconds: int = 30
    ) -> bool:
        """Helper method for filtering out interactions by the same user that happened in short succession."""
```

src/shared/logs/raw_events.py
```diff
@@ -0,0 +1,213 @@
...
+
+
+class RawStatusEvent(RawEvent):
+    """Raw status change event."""
+
+    action: Literal["insert", "update"]
+    status_value: str
+
+    def is_canceled_by(
+        self, other: "RawEvent", time_threshold_seconds: int = 30
```
fricklerhandwerk: I find it slightly smelly that we have to pass through the threshold everywhere, but have no good alternative here; therefore feel free to ignore.

src/shared/logs/raw_events.py
```diff
@@ -0,0 +1,213 @@
...
+                "maintainers.remove",
+            }
+
+        return False
+
+
+RawEventType = RawStatusEvent | RawPackageEvent | RawMaintainerEvent
+
+
+class EventFetcher:
```
fricklerhandwerk: I'm a bit confused why this is a class. The way it's used makes it appear as the object holds some sort of state, but really it just obtains data. We could put this as a function into a separate module if you want namespacing, but doing namespacing via classes seems unidiomatic to me.

src/shared/logs/raw_events.py
fricklerhandwerk, line 1: We may as well just call it `events.py` or `event_types.py` to make it clear that this is essentially the data model with very little logic

src/shared/logs/folding.py
```diff
@@ -0,0 +1,162 @@
...
+    suggestion_id: int
+    timestamp: datetime  # Timestamp of the most recent event of the collection
+    username: str
+
+
+class FoldedStatusEvent(FoldedEvent):
+    """A folded status change event (always singular)."""
+
+    action: Literal["insert", "update"]
+    status_value: str
```
fricklerhandwerk: Why isn't this an enum that follows the DB model? Was this always a string or is this a remnant from the times where we were using dicts for everything?

src/shared/logs/folding.py
```diff
@@ -0,0 +1,162 @@
...
+                else:
+                    # End current accumulator, start new one
+                    folded_events.append(accumulator)
+                    accumulator = FoldedPackageEvent(
+                        suggestion_id=event.suggestion_id,
+                        timestamp=event.timestamp,
+                        username=event.username,
+                        action=event.action,
+                        package_names=[event.package_attribute],
+                    )
```
fricklerhandwerk:
```suggestion
                if (
                    isinstance(accumulator, FoldedPackageEvent)
                    and event.action == accumulator.action
                    and event.username == accumulator.username
                    and event.suggestion_id == accumulator.suggestion_id
                ):
                    # Continue accumulating packages
                    accumulator.package_names.append(event.package_attribute)
                    accumulator.timestamp = event.timestamp
                else:
                    # End current accumulator if it exists
                    if accumulator:
                        folded_events.append(accumulator)
                    
                    # Start new accumulator
                    accumulator = FoldedPackageEvent(
                        suggestion_id=event.suggestion_id,
                        timestamp=event.timestamp,
                        username=event.username,
                        action=event.action,
                        package_names=[event.package_attribute],
                    )
```

...seems a bit easier on the eyes

src/shared/logs/folding.py
```diff
@@ -0,0 +1,162 @@
...
+
+    def __iter__(self) -> Iterator[FoldedEventType]:
+        """Make the collection iterable."""
+        return iter(self._folded_events)
+
+    def __getitem__(self, index: int) -> FoldedEventType:
+        """Make the collection subscriptable (supports indexing and slicing)."""
+        return self._folded_events[index]
+
+    def _fold_events(self, collection: EventCollection) -> list[FoldedEventType]:
```
fricklerhandwerk: Pretty sure this can be a standalone procedure. Wrapping what amounts to a list in `EventCollection` seems overkill. Maybe we can have this as a method on a much dumber `EventCollection` (as noted in https://github.com/Nix-Security-WG/nix-security-tracker/pull/631#discussion_r2245658175)?

src/shared/logs/folding.py
```diff
@@ -0,0 +1,162 @@
...
+from shared.logs.collections import EventCollection
+from shared.logs.raw_events import (
+    Maintainer,
+    RawMaintainerEvent,
+    RawPackageEvent,
+    RawStatusEvent,
+)
+
+
+class FoldedEvent(BaseModel):
```
fricklerhandwerk: This is indistinguishable from a `RawEvent`. How about we inherit from that here as well, and also maybe just call it `Event`?

I may be falling victim to premature abstraction myself with the following thought, but: it seems more ergonomic to me to derive folded events and annihilating events from the same base classes that pertain to the object of interest (suggestion, package, maintainer). Arguably these will be tiny classes, and arguably they still may change independently in the future, so I don't know. But the overlap in semantics makes it compelling to coalesce them along *that* dimension rather than anything else.

fricklerhandwerk:
Overall this is a very productive change, and the tests add a lot of clarity as to what happens in the UI. The machinery to sort out the underlying semantics is still a bit clunky, and arguably a bit overengineered, but I can see the thought process that went into this and have tried to find a direction towards something more light-weight in the review comments.

fricklerhandwerk:
This change should make this one rather quick to implement:
- https://github.com/Nix-Security-WG/nix-security-tracker/issues/502

src/shared/logs/raw_events.py
florentc: Initially, I wanted some kind of accumulator through the EventFetcher class where you would initialize the fetcher f, then f.fetch_suggestion(foo), then f.fetch_suggestion(bar), or f.fetch_most_recent(time_in_days), for instance.

Eventually I ruled this out but the class stayed. It's removed in the applied suggestions.

src/shared/logs/raw_events.py
florentc: Done with another name

src/shared/logs/folding.py
florentc: This is a remnant from before.

src/shared/logs/folding.py
florentc: Regarding the first remark: `RawEvent` and `FoldedEvent` are indistinguishable in content if you isolate them and compare them, but not in meaning nor usage. Having a collection of `RawEvent` and having a collection of `FoldedEvent` is meant to convey that they are to be treated differently, unlike in the design before where singular events and folded events shared the same classes, leading to confusion in the model. IMO making `FoldedEvent` inherit from `Event` would not make sense. `FoldedEvent` is to be seen as the type of what's sent to the view template. Also, the `timestamp` attribute has a different meaning for FoldedEvents. Maybe `ViewEvent`/`BatchedEvents` would be better.

I'm not sure I understand the second paragraph correctly.

src/shared/logs/raw_events.py
florentc: Done

src/shared/logs/collections.py
florentc: Done

src/shared/logs/collections.py
florentc: Done, now collections of events (raw events or folded events) no longer have a dedicated class but are simple lists.

src/shared/logs/folding.py
florentc: The suggestion doesn't work as is, we can see this together if you want as I am not sure I get the spirit and target shape of the refactor

src/shared/logs/folding.py
florentc: Now it is

src/shared/logs/batches.py
```diff
@@ -0,0 +1,143 @@
...
+
+FoldedEventType = FoldedStatusEvent | FoldedPackageEvent | FoldedMaintainerEvent
+
+
+def batch_events(
+    events: list[RawEventType], pre_sort: bool = False
+) -> list[FoldedEventType]:
+    """
+    Batch consecutive events of the same type from the same user into bulk operations.
+    Events must be sorted chronologically. Use the pre_sort flag if not.
```
yannham: What is the use of `pre_sort`? To avoid performing work if the events are already sorted? The wording makes it sound like there are cases where we don't want events to be sorted _at all_, which I guess is not the case. If my initial guess is right, then I would suggest to make it clear, like: "Set the pre_sort flag to False if the events are already sorted", or something like that.

src/shared/logs/events.py
```diff
@@ -0,0 +1,138 @@
...
+        """Check if this event is canceled by another event.
+
+        Must be implemented by subclasses to define specific cancellation logic.
+        """
+        pass
+
+    def precedes_close_related_event(
+        self, other: "RawEvent", time_threshold_seconds: int = 30
+    ) -> bool:
+        """Checks the event is followed by one related to the same suggestion by the same user within a give time window"""
```
yannham:
```suggestion
        """Checks if the event is followed by one related to the same suggestion by the same user within a given time window"""
```

src/shared/logs/events.py
```diff
@@ -0,0 +1,138 @@
...
+    """
+    return sorted(events, key=lambda event: event.timestamp)
+
+
+def remove_canceling_events(
+    events: list[RawEventType], time_threshold_seconds: int = 30, pre_sort: bool = False
+) -> list[RawEventType]:
+    """
+    Remove consecutive events that cancel each other out within a time window.
+    Events must be sorted chronologically. Use the pre_sort flag if not.
```
yannham: Ditto about `pre_sort` documentation.

yannham (approved):
I like this refactoring. It's much cleaner, and while the abstract class and all can sound a bit overkill at first, the code is rather simple to read and understand, with a clear separation of concerns.

Regarding testing, maybe we could add a bit more, like testing the cancellation logic for example?

src/shared/logs/folding.py
yannham: @florentc I think @fricklerhandwerk 's point, which I agree with, is that most of the code in `if not accumulator` and `else` is the same. The only difference is that you add the accumulator to the folded events in the `else` branch, but the whole `FoldedPackageEvent` creation is duplicated, so it should be possible to factor it out, as Valentin did (it probably doesn't work for silly reasons, not deep ones, right?)

src/shared/logs/folding.py
yannham: Just FYI, since I approved the PR, I must note that this is still a `str` currently. Do you intend to fix that @florentc ? Or maybe in a follow-up PR?

src/shared/logs/folding.py
florentc: Done

src/shared/logs/batches.py
florentc: I renamed it to `sort`, it makes more sense indeed.

The operations that have a `pre_sort` flag (for now, remove events that cancel each other, and group similar events together) require sorted lists of events, and preserve that order. So when we chain those operations, it isn't useful to re-sort. Hence the flag to be used in only the first operation.

src/shared/logs/events.py
florentc: Done

src/shared/logs/events.py
florentc: Done with the renaming to `sort`

src/shared/logs/folding.py
florentc: Added a TODO

src/shared/logs/raw_events.py
florentc: Done

PR #635, opened 2025-09-15
Dynamic updates of activity log

### LINKED ISSUE #630: Dynamically update activity log on changes
As a security tracker user, when I change something about a suggestion that would show up in the activity log, I want the activity log to update instantly.

## Acceptance criteria

Given a suggestion,
When I {{add,remove} a {maintainer, package}, undo an action},
Then that change shows up in the activity log without a page reload.

florentc:
Fixes #630


src/webview/views.py
```diff
@@ -593,6 +593,7 @@ def suggestion_view_context() -> dict:
             }
 
         # We only have to modify derivations when they are editable
+        package_changes_made = False
```
yannham, line 255: Nitpick: the previously existing comment should be moved one line down, because here it looks like it's documenting `package_changes_made`.

src/webview/views.py
```diff
@@ -679,11 +681,27 @@ def suggestion_view_context() -> dict:
...
             if undo_status_change:
+                # Fetch fresh activity log for undo status change
+                raw_events = fetch_suggestion_events(suggestion.pk)
+                fresh_activity_log = batch_events(
+                    remove_canceling_events(raw_events, sort=True)
+                )
+
                 snippet = render_to_string(
                     "components/suggestion.html",
-                    suggestion_view_context(),
```
yannham, line 684: Do we still use `suggestion_view_context()` somewhere?

florentc: We do

src/webview/views.py
florentc: The comment is about the whole block rather than just the next line. If we moved it one line down, the `package_changes_made` initialization would appear to be unrelated, yet it is part of this block.

PR #640, opened 2025-09-30
Notification center

florentc:
This PR introduces a notification center for each user.

To access, click on the bell on the top right next to your name.

Three new `manage` commands are introduced to inject new notifications (this PR doesn't introduce any mechanism to generate notifications from events):
- `create_notification` (specify `--user` and optionally `--title` `--message`)
- `list_notifications` (specify `--user`)
- `clear_notifications` (specify `--user`)
- `resync_notification_counters` (optionally `--user`)
- `create_missing_profiles`

There is a new db model for notifications so you'll need to run the migrations. The model is defined in webview, not in shared, as notifications seemed to me like a web viewer feature only.

The views are in their own module to start a trend of no longer adding stuff in the overcrowded `views.py` in webview.

There is full compatibility for no-javascript and live reload (when clicking on "mark as read") when js is enabled.

I didn't use `render_to_string` so that writing tests will be easier by checking data in the returned context instead.

### DIFF
--- a/src/pyproject.toml
+++ b/src/pyproject.toml
@@ -3,5 +3,5 @@ name = "web-security-tracker"
 version = "0.0.1"
 
 [tool.setuptools]
-packages = [ "project", "shared", "webview" ]
+packages = { find = {} }
 include-package-data = true


src/webview/management/commands/create_missing_profiles.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        username = options.get("user")
+
+        if username:
+            try:
+                user = User.objects.get(username=username)
+                users = [user]
+                self.stdout.write(f"Checking profile for user '{username}'")
```
yannham: Or change the other below, but the two message are distinct for one users and many.

```suggestion
                self.stdout.write(f"Creating missing profile for user '{username}'")
```

src/webview/models.py
```diff
@@ -25,3 +33,138 @@ def create_profile(
 ) -> None:
     if created:
         Profile.objects.create(user=instance)
+
+
+class NotificationManager(models.Manager):
+    def create_for_user(
```
yannham, line 39: it's a stylistic question, so feel free to entirely ignore it. I find that `_for_user` is a bit heavy and not necessarily required, as it is 1. conveyed by the argument (similarly, we could also push the logic and call the function `create_notification_for_user_with_title_and_message_and_is_read`), and 2. there's no `create` that could be confused with `create_for_user`. So I would just call it `create`.

src/webview/notifications/views.py
```diff
@@ -0,0 +1,102 @@
...
+class NotificationCenterView(LoginRequiredMixin, ListView):
+    """Main notification center view showing user's notifications."""
+
+    template_name = "notifications/notification_center.html"
+    model = Notification
+    context_object_name = "notifications"
+    paginate_by = 10
+
+    def get_queryset(self) -> QuerySet[Notification]:
+        """Return notifications for the current user, ordered by creation date."""
```
yannham: I suppose ? The `-` in `-created_at` ?


```suggestion
        """Return notifications for the current user, ordered by decreasing creation date."""
```

yannham (approved):
LGTM: clear code, small and understandable endpoints, good documentation. It was easy to follow, and looks nice on the frontend. A gotcha is that I haven't tested it yet, as I don't currently have my  NixOS laptop with the right setup with me, but as far as code is concerned, it looks fine.

src/webview/management/commands/create_missing_profiles.py
florentc: Done

src/webview/models.py
florentc: There would be a name clash with a method from a parent class.

src/webview/notifications/views.py
florentc: Done

PR #641, opened 2025-10-02
Add setting to configure whether a user needs to be a committer or a maintainer in order to create issues

Erethon:
This PR:
- Introduces a new setting and code to configure whether a user needs to be a maintainer or a committer in order to create an issue.
- Fixes a bug that prevented users from being added to their groups.


src/shared/auth/utils.py
```diff
@@ -18,3 +18,14 @@ def ismaintainer(user: Any) -> bool:
...
+
+def can_publish_github_issue(user: Any) -> bool:
+    if isadmin(user):
+        return True
+
+    else:
+        if settings.GH_ISSUES_COMMITTERS_ONLY:
+            return iscommitter(user)
+        else:
+            return ismaintainer(user)
```
yannham: Nitpick (feel free to ignore): this could be expressed as one boolean condition `isadmin(user) || iscommitter(user) || (!settings.GH_ISSUES_COMMITTERS_ONLY && ismaintainer(user))`. Not entirely sure it's more readable, though.

balsoft: Seems more readable to me

Issue #647, opened 2025-10-29
Automatic notifications for package maintainers


balsoft:
Given I am a package maintainer,
Given I am logged in with my GitHub account,
Given I have not opted out of automatic notifications,
When a suggestion affecting my package is created,
Then I receive a notification in my [Overview](https://github.com/Nix-Security-WG/nix-security-tracker/issues/174)

Given I am logged in with my GitHub account,
When I open [Subscription Center](https://github.com/Nix-Security-WG/nix-security-tracker/pull/644)
Then I can opt out of automatic notifications

PR #655, opened 2025-11-06
Automatic subscription to maintained packages

florentc:
This addresses #647.

This feature introduces a toggle (by default on) to be notified about new suggestions involving packages a user maintains, even if they have not manually added it in the subscriptions center.

Notifications mention that you are notified because you are a maintainer

### DIFF
--- a/src/webview/subscriptions/views.py
+++ b/src/webview/subscriptions/views.py
@@ -155,10 +155,7 @@ class ToggleAutoSubscribeView(LoginRequiredMixin, TemplateView):
 
         profile = request.user.profile
 
-        if action == "enable":
-            profile.auto_subscribe_to_maintained_packages = True
-        else:  # disable
-            profile.auto_subscribe_to_maintained_packages = False
+        profile.auto_subscribe_to_maintained_packages = action == "enable"
 
         profile.save(update_fields=["auto_subscribe_to_maintained_packages"])
 


src/shared/listeners/notify_users.py
```diff
@@ -30,32 +40,71 @@ def create_package_subscription_notifications(
...
-            for pkg in user.profile.package_subscriptions
-            if pkg in affected_packages
-        ]
+    for user in all_users_to_notify:
+        # Determine notification reason and affected packages for this user
+        user_affected_packages = []
+        notification_reason = []
+
+        # Check if user is subscribed to any affected packages
+        if user in subscribed_users:
```
YorikSar: Note that subscribed_users here is a list, which would make this and the ckeck o line 91 a bottleneck for popular packages.

src/shared/listeners/notify_users.py
```diff
@@ -30,32 +40,71 @@ def create_package_subscription_notifications(
...
 
-    if not subscribed_users.exists():
-        logger.debug(f"No subscribed users found for packages: {affected_packages}")
-        return
+    # Find maintainers of affected packages from cached suggestion
+    maintainer_users = set()
+    if maintainers_github:
+        maintainer_users = set(
+            User.objects.filter(
+                username__in=maintainers_github,
```
YorikSar: Would it be better to have another JOIN here instead of fetching all GitHub handles into Python just to send them back?

src/webview/subscriptions/views.py
```diff
@@ -141,6 +141,60 @@ def _handle_error(self, request: HttpRequest, error_message: str) -> HttpRespons
...
+
+        if action not in ["enable", "disable"]:
+            return self._handle_error(request, "Invalid action.")
+
+        profile = request.user.profile
+
+        if action == "enable":
+            if profile.auto_subscribe_to_maintained_packages:
+                return self._handle_error(
+                    request, "Auto-subscription is already enabled."
```
YorikSar: I don't think this should be an error: if response to one query was lost and user clicked on the button twice, it should still be fine as long as we reach expected state.

florentc: Addressed in [dcc4cc3](https://github.com/NixOS/nix-security-tracker/pull/655/commits/dcc4cc3ef402ef6ab83b0af527c355b426d62aff)

src/shared/listeners/notify_users.py
florentc: Addressed in [7fd96cb](https://github.com/NixOS/nix-security-tracker/pull/655/commits/7fd96cb8b780cced082da516058ef1364887bb6b)

src/shared/listeners/notify_users.py
florentc: Addressed in [82ee088](https://github.com/NixOS/nix-security-tracker/pull/655/commits/82ee088c6b0674166039d8692ad5d3653f176983)

YorikSar (approved):
Looks fine to me.

Issue #654, opened 2025-11-05
Suggestion comments


balsoft:
Given I am logged in:

When I dismiss a suggestion
Then I MUST add a comment explaining the reason for dismissal

When I draft a suggestion
Then I CAN add a comment adding information to the issue

When I publish a suggestion
Then I SHOULD add a comment with additional information about the issue 

Given I am logged in
When I view a suggestion
Then I see the comment with the suggestion

When I publish a suggestion
When the suggestion has an associated comment
Then the comment must be added prominently to the corresponding GitHub issue

florentc:
I think this is a partial duplicate of #621.

Issue #621, opened 2025-07-18
Confirm dismissing automatic suggestions


fricklerhandwerk:
As a security team member, I want more friction when dismissing a suggestion, in order to avoid error corrections. I also want to provide a reason for the dismissal.

## Acceptance criteria

Given an automatic suggestion,
When I dismiss it,
There is a confirmation prompt with an optional text field for the reason.

PR #656, opened 2025-11-12
Editable comments on suggestions

florentc:
This addresses #654 and #621.

**WARNING**
This PR adds a db migration that should be applied after #655 is merged (or after editing the migration numbers accordingly).

### DIFF
--- a/src/shared/github.py
+++ b/src/shared/github.py
@@ -108,14 +108,28 @@ def create_gh_issue(
 
     def additional_comment() -> str:
         if comment:
-            escaped_comment = comment.replace("`", "\\`")
+            # Find the maximum number of consecutive backticks in the comment
+            max_backticks = 0
+            current_backticks = 0
+
+            for char in comment:
+                if char == "`":
+                    current_backticks += 1
+                    max_backticks = max(max_backticks, current_backticks)
+                else:
+                    current_backticks = 0
+
+            # Use at least 3 backticks, or one more than the maximum found in
+            # order to escape accidents or attempts at escaping the code block
+            fence_backticks = "`" * max(3, max_backticks + 1)
+
             return f"""
 
 ## Additional comment
 
-```
-{escaped_comment}
-```"""
+{fence_backticks}
+{comment}
+{fence_backticks}"""
         else:
             return ""
 


src/shared/github.py
```diff
@@ -105,6 +106,18 @@ def affected_nix_packages() -> str:
...
 </details>"""
 
+    def additional_comment() -> str:
+        if comment:
+            return f"""
+
+## Additional comment
+
+```
+{comment}
```
balsoft, line 131: Maybe it's better to have it as a codeblock, and instead go with a blockquote or something similar?

In any case, we should escape the comment so that it never leaks out of the backticks

florentc: Good catch! I escaped the backticks. I think for now it's more relevant to use the current code block rather than blockquote because on sec-tracker the comment field is plain text with no intention of supporting markdown formatting. Using a blockquote we would either have to acknowledge that users write markdown in the comment sections, or escape everything which would look weird. I think escaping the backticks in code block is a good compromise for now.

src/shared/github.py
```diff
@@ -108,12 +108,13 @@ def affected_nix_packages() -> str:
 
     def additional_comment() -> str:
         if comment:
+            escaped_comment = comment.replace("`", "\\`")
```
YorikSar: This will turn ` into
````
```
\`
```
````
Which renders as
```
\`
```

To escape triple backticks, you need more backticks around the block:

`````
````
```
````
`````
renders as
````
```
````
(yes, I added 5 backticks before and after to show this)

Comment (florentc) [reply src/shared/github.py]: Solved in [dc452ae](https://github.com/NixOS/nix-security-tracker/pull/656/commits/dc452ae8c3690c08c40c06d74b24bb73bb36f40e)

File: src/shared/github.py [hunk 3]
```diff
@@ -105,6 +106,33 @@ def affected_nix_packages() -> str:
...
+            for char in comment:
+                if char == "`":
+                    current_backticks += 1
+                    max_backticks = max(max_backticks, current_backticks)
+                else:
+                    current_backticks = 0
+
+            # Use at least 3 backticks, or one more than the maximum found in
+            # order to escape accidents or attempts at escaping the code block
+            fence_backticks = "`" * max(3, max_backticks + 1)
```
YorikSar, line 124: This is fine, and I know that fixing a problem with a regex doubles the problem, but iterating over a comment in Python seems scary to me. This could also work to count backtick runs:
```
max(m.end() - m.start() for m in re.finditer('`+', comment))
```

PR #669, opened 2025-12-05
fix: Improve performance of suggestion generation

Erethon:
While testing a management command for deleting [older releases](https://github.com/NixOS/nix-security-tracker/issues/606), I had to regenerate suggestions and I got frustrated because it's a slow operation. This patch tries to improve this a bit. I know it seems like a weird diff, but it's mostly rearranging things so we can use `prefetch_related`. Here's the reasoning behind the changes:

1) Switch the `count() == 0` to an `not exists()` query as it's easier to read and more django-like. While it might look like we can remove this check completely because we're doing a similar query further down the line, this is not the case, as it's faster to return early here before doing more complex queries.
2) Use `prefetch_related` to avoid N+1 queries on `AffectedProducts`. Then, use the prefetched fields instead of doing more queries for cpes and versions. This is the main reason this diff is faster.

This brought down the time it takes to regenerate suggestions to less than 6 minutes, while it was taking a bit over 10 minutes earlier on staging.

### DIFF
--- a/src/shared/listeners/cache_suggestions.py
+++ b/src/shared/listeners/cache_suggestions.py
@@ -85,28 +85,29 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
 
     affected_products = dict()
     all_versions = list()
+
     prefetched_affected_products = AffectedProduct.objects.filter(
-        container__cve=suggestion.cve
-    )
+        container__cve=suggestion.cve, package_name__isnull=False
+    ).prefetch_related("versions", "cpes")
+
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
Ready for review. I added a basic test that exercises the code path, but doesn't check any interesting properties since (as noted in the comment) most of that would be covered by a Pydantic type check. I didn't touch the existing fixtures to keep the diff reasonable, but surely we can throw away a lot of duplicate code once someone™ does that cleanup (likely I'll do that as one of the next moves).

PR #671, opened 2025-12-05
fix: error creating GitHub issue due to empty title

### LINKED ISSUE #653: Cannot publish a drafted issue
CVE-2025-8396 which is currently drafted on the production instance https://tracker.security.nixos.org/drafts/#suggestion-1332 cannot be published.

The publication fails with the error message: "Error: unable to change status of this suggestion to published"

Image

fricklerhandwerk:
Closes #653

This also fixes the display component for suggestion status change that is affected by the same problem. Long term this fallback logic should be handled in one place.

### DIFF
--- a/src/shared/models/linkage.py
+++ b/src/shared/models/linkage.py
@@ -65,7 +65,9 @@ class CVEDerivationClusterProposal(TimeStampMixin):
             # end.
             status=IssueStatus.AFFECTED,
             description=Description.objects.create(
-                value=self.cached.payload["description"]
+                # FIXME(@fricklerhandwerk): Ensure the cached payload is correct beforehand instead of an ad-hoc fallback here. [ref:title-fallback-hack]
+                # Or maybe the data model can be relaxed to have an optional description.
+                value=self.cached.payload["description"] or ""
             ),
         )
         issue.cve.add(self.cve)
--- a/src/webview/tests/test_issues.py
+++ b/src/webview/tests/test_issues.py
@@ -166,3 +166,42 @@ class IssueTests(TestCase):
             self.suggestion.status,
             CVEDerivationClusterProposal.Status.PUBLISHED,
         )
+
+    def test_publish_gh_issue_empty_description(self) -> None:
+        """Test that creating a GitHub issue will succeed and update suggestion status, despite no CVE description"""
+        # [tag:test-github-create_issue-description]
+
+        url = reverse("webview:drafts_view")
+        self.cve_container.descriptions.clear()
+        self.cve_container.save()
+        self.suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
+        self.suggestion.save()
+        cache_new_suggestions(self.suggestion)
+
+        # FIXME(@fricklerhandwerk): Mock Github's `create_issue()` here, not our own procedure! [ref:todo-github-connection]
+        with patch("webview.views.create_gh_issue") as mock:
+            mock.side_effect = lambda *args, **kwargs: create_gh_issue(
+                *args,
+                github=MockGithub(expected_issue_title="Dummy Title"),  # type: ignore
+                **kwargs,
+            )
+
+            response = self.client.post(
+                url,
+                {
+                    "suggestion_id": self.suggestion.pk,
+                    "new_status": "published",
+                    "comment": "",  # Empty comment
+                },
+            )
+
+        messages = list(get_messages(response.wsgi_request))
+        self.assertFalse(
+            any(m.level_tag == "error" for m in messages),
+            "Errors on issue submission",
+        )
+        self.suggestion.refresh_from_db()
+        self.assertEqual(
+            self.suggestion.status,
+            CVEDerivationClusterProposal.Status.PUBLISHED,
+        )


src/shared/github.py
```diff
@@ -134,7 +138,14 @@ def additional_comment() -> str:
...
     repo = github.get_repo(f"{settings.GH_ORGANIZATION}/{settings.GH_ISSUES_REPO}")
-    title = cached_suggestion.payload["title"]
+
+    # XXX(@fricklerhandwerk): 3/4 of CVEs have no title, 1/2 have no description, but none has neither.
+    # This hack, which we're also using in the template, can -- for now -- be expected to always work.
+    # Users can still change the GitHub issue title on GitHub if the truncated description is not informative.
+    if cached_suggestion.payload["title"]:
+        title = cached_suggestion.payload["title"]
+    else:
+        title = truncatewords(cached_suggestion.payload["description"], 10)
```
florentc, line 148: Even if in practice we have not encountered any case where there's no title nor description, I think it's worth having a catch-all that either names it "Untitled" or fails after logging an error to avoid possible silent failures.

src/webview/tests/test_comments.py
florentc, line 1: This module (`test_comments`) was intended to gather tests about the user-provided comment feature (checking it's required when dismissing, and so on). The new test doesn't belong in this module IMO. The only test related to publishing issues that could belong here is checking that the user-provided comment ends up appearing in the published issue on GitHub.

I think the new test should belong to the already-too-long-but-we'll-split-eventually `test_suggestions` module which deals with suggestions lifecycle and status change, or something new.

src/webview/views.py
```diff
@@ -378,6 +378,8 @@ def suggestion_view_context() -> dict:
                     {
                         "suggestion_id": suggestion.pk,
                         "title": cached_suggestion.payload["title"],
+                        # XXX(@fricklerhandwerk): Same hack to fall back in case of empty title.
```
florentc: This comment may be confusing because there isn't really a hack here (nor a similar one close by that would be "Same"). I imagine you refer to using a truncated description to replace a missing title, but that's not happening right here. Maybe you used to do the hack here and forgot to remove the comment?

```suggestion
```

src/shared/github.py
```diff
@@ -28,6 +29,9 @@ def create_gh_issue(
     cached_suggestion: CachedSuggestions,
     tracker_issue_uri: str,
     comment: str | None = None,
+    # FIXME(@fricklerhandwerk): [tag:todo-github-connection] Make an application-level "GitHub connection" object instead.
```
florentc, line 32:
```suggestion
    # TODO(@fricklerhandwerk): [tag:todo-github-connection] Make an application-level "GitHub connection" object instead.
```

src/shared/github.py
```diff
@@ -134,7 +138,14 @@ def additional_comment() -> str:
             return ""
 
     repo = github.get_repo(f"{settings.GH_ORGANIZATION}/{settings.GH_ISSUES_REPO}")
-    title = cached_suggestion.payload["title"]
+
+    # XXX(@fricklerhandwerk): 3/4 of CVEs have no title, 1/2 have no description, but none has neither.
```
florentc:
```suggestion
    # NOTE(@fricklerhandwerk): 3/4 of CVEs have no title, 1/2 have no description, but none has neither.
```

src/shared/tests/test_github_sync.py
```diff
@@ -20,6 +20,10 @@ class MockGithubUser:
     def __init__(self, id_value: str) -> None:
         self.id: int = int(id_value)
 
+    @property
+    def login(self) -> str:
+        return str(self.id)
+
```
florentc, line 28: 1. Is the property used somewhere? I imagine it is, indirectly, when mocking, in which case maybe there should be a mention of which tests / kinds of tests depend on it as it won't be clear from grepping.
2. Isn't the `id` (number) different from the `login` (string handle)

src/shared/tests/test_github_sync.py
```diff
@@ -49,6 +53,24 @@ def get_team_by_slug(self, slug: str) -> MockGithubTeam:
         return self.teams[slug]
 
 
+class MockGithubIssue:
+    @property
+    def html_url(self) -> str:
+        return "mock"
```
florentc, line 63: 1. Same as "1" in previous review comment
2. Out of curiosity, what is this supposed to be when not mocking? Simply the full url to the issue?

src/webview/tests/test_comments.py
```diff
@@ -307,3 +311,45 @@ def test_updating_comment_on_existing_suggestion(self) -> None:
 
         suggestion = drafts_response.context["object_list"][0].proposal
         self.assertEqual(suggestion.comment, updated_comment)
+
+    def test_publish_gh_issue(self) -> None:
+        """Test that creating a GitHub issue will succeed and update the suggestion status"""
+        # TODO(@fricklerhandwerk): [tag:test-github-create_issue] Move this test to a less surprising location.
+        # That would require consolidating the setup logic quite a bit.
```
florentc: Oh ok you noted it too. I still think we should take care of moving it to a better place before merging as it's really easy in this project to start mixing up everything once there is a seed of disorganization to build upon.

src/shared/github.py
fricklerhandwerk: We should agree on semantics here. See [PEP 350](https://peps.python.org/pep-0350/) for convention that I've seen actually used in other places as well.

src/shared/tests/test_github_sync.py
fricklerhandwerk: Yes, both. That's okay; we just need *some* sting right now.

But as noted in one of the comments, we should have the stub behavior in the test that uses it. This here is just bad-bad-bad but I don't want to rewrite the entire thing in this PR.

I'll add another comment...

src/shared/tests/test_github_sync.py
fricklerhandwerk: Yes exactly

src/webview/tests/test_comments.py
fricklerhandwerk: I know, this is why I added the comment: there's a lot of machinery that would require cleanup if we wanted to avoid yet another copy of essentially the same setup code. I prefer to do the refactoring after getting the fix deployed, right now users can't publish issues.

src/shared/github.py
fricklerhandwerk: Thanks for the prompt to be more defensive. Weeding this out had me find another place where we're just assuming that a description will always exist. Adding a fallback and a check...

src/webview/tests/test_comments.py
florentc: By far, I'd rather have another module that duplicates the exact same setup code just for this single test, rather than having an unrelated test not where it belongs.

"To be sorted later" things tend to never be addressed, and down the line, we either suffer figuring out and maintaining disorganized code, or worse, having to refactor (e.g. the styling PR or the future view.py / suggestions refactor PR).

src/shared/github.py
fricklerhandwerk: continued here: https://github.com/NixOS/nix-security-tracker/pull/673

PR #676, opened 2025-12-09
Rename `Restore` to `Convert to draft` in dismissed issues view

Erethon:
~When restoring an issue from a dismissed state it should go back to untriaged, not drafts.~

After the discussion in this issue I believe renaming this button is closer to the expected flow.

### DIFF
--- a/src/webview/templates/components/suggestion.html
+++ b/src/webview/templates/components/suggestion.html
@@ -74,7 +74,7 @@
         </button>
       {% elif status_filter == "rejected" %}
         <button type="submit" name="new_status" value="accepted" class="draft-color">
-          Restore
+          Convert to draft
         </button>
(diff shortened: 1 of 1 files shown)


florentc:
I have wondered about this in the past too, it didn't feel right to me either. But now I feel like that is intended and desirable.

- Untriaged is a suggestion nobody has had an opinion about.
- Dismissing is when you don't care about it
- Draft is when you care about it
- You must choose one or the other
- If you regret dismissing, it means you consider it worth being published, and hence go to drafts for adjustments (maintainers, packages, comments) before publishing.

I don't think (or rather I no longer think) it should go back to untriaged, because it's the inbox for new suggestions.

Maybe we should just name the button "Draft issue" on the "dismissed" page?
Any opinion @fricklerhandwerk ?

PR #683, opened 2025-12-15
track stable Nixpkgs

fricklerhandwerk:
as discussed with @erethon. cc @raitobezarius in case you have opposing opinions, because we had this topic back in the day.


fricklerhandwerk:
It would be a nice forcing function to dogfood what the tracker is supposed to be about... But in any case, given the build breaks so hard, I guess let's not bother for now.

Issue #550, opened 2025-05-13
Show activity log for published issues


fricklerhandwerk:
As a user of the security tracker, I want to see the history of changes on a published issue, in order to know who worked on it when.

## Acceptance criteria

Given there's a published vulnerability record,
When I view its details,
There's an activity log that goes back to when it still was a suggestion or draft.

PR #690, opened 2025-12-16
fix: Removed packages not removed from published issues

### LINKED ISSUE #652: Removed packages are kept once the issue is published
For example in https://tracker.security.nixos.org/issues/NIXPKGS-2025-0004 I removed `python312Packages.llama-index-readers-txtai` and `python313Packages.llama-index-readers-txtai` but they are still visible/shown in the same way than the impacted package.

The corresponding issue on GitHub does not show the removed packages as expected (https://github.com/NixOS/nixpkgs/issues/457571).

florentc:
- Reference the origin suggestion in the issue model
- Simplify the issue model by removing now redundant info
- Remove no longer needed cached issues

This fixes #652
This is a step forward towards solving #550 and #544

Note:
During development I went through a few db migrations. I wanted to roll them back to regenerate then tweak one single migration from scratch but reverting migrations failed on my local setup, so I could not do it reliably.
Tell me in review if this matters for you or not.

### DIFF
--- a/src/shared/migrations/0061_nixpkgs_issue_relate_to_suggestion.py
+++ b/src/shared/migrations/0061_nixpkgs_issue_relate_to_suggestion.py
@@ -30,13 +30,6 @@ def initialize_suggestion_field(apps, schema_editor):
                 issue.suggestion = suggestion
                 issue.save(update_fields=['suggestion'])
 
-
-def reverse_initialize_suggestion_field(apps, schema_editor):
-    """Reverse migration - clear suggestion field."""
-    NixpkgsIssue = apps.get_model('shared', 'NixpkgsIssue')
-    NixpkgsIssue.objects.update(suggestion=None)
-
-
 class Migration(migrations.Migration):
 
     dependencies = [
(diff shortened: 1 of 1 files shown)


src/webview/admin.py
```diff
@@ -136,5 +135,5 @@ class NixpkgsIssueAdmin(AutocompleteMixin, admin.ModelAdmin):
     readonly_fields = ["code"]
 
 
-admin.site.register(CVEDerivationClusterProposal)
+# admin.site.register(CVEDerivationClusterProposal)
```
fricklerhandwerk: We should do something about that... we haven't looked at the data through the admin view in a while, but that's no reason to keep it working.

fricklerhandwerk (approved):
Looks good except for that one workaround in the admin view. Out of curiosity, in which way did the reverse migrations fail?

florentc:
`django.db.utils.ProgrammingError: multiple primary keys for table "shared_cachednixpkgsissue" are not allowed`
On rolling back the penultimate migration

PR #693, opened 2025-12-18
chore: add correct label when publishing GitHub issue

fricklerhandwerk:


infra/production.nix
```diff
@@ -94,6 +94,7 @@ in
       GH_ISSUES_REPO = "nixpkgs";
       GH_SECURITY_TEAM = "security";
       GH_COMMITTERS_TEAM = "nixpkgs-committers";
+      GH_ISSUE_LABELS = [ "1.severity:security" ];
```
Erethon: I believe this should be [`ISSUES` (plural)](https://github.com/NixOS/nix-security-tracker/blob/77e8067c6fbfca7db5d805a3e85b7259f3f4125c/src/project/settings.py#L129).

Erethon (changes requested):

infra/production.nix
fricklerhandwerk: Oof, good catch!

PR #696, opened 2025-12-19
refactor: Migrate some tests to `pytest`

fricklerhandwerk:


src/shared/tests/conftest.py
florentc, line 1: Just to confirm, we move fixtures to fixtures.py so that we reference them in both `conftest.py` from shared and webview right? Could we reference conftest.py from shared in the conftest.py from webview (as "shared" may hint that the conf of webview would extend it)

src/shared/tests/fixtures.py
```diff
@@ -0,0 +1,143 @@
...
+    container = cve.container.create(provider=org, title="Dummy Title")
+    container.affected.add(affected)
+    container.descriptions.add(desc)
+    container.metrics.add(metric)
+
+    return container
+
+
+@pytest.fixture
+def evaluation(db: None) -> NixEvaluation:
```
florentc, line 52: This is not introduced in this PR but I forgot: why do we have a `db` parameter on some tests and fixtures? We don't seem to use it in the bodies. Is `db` a fixture defined somewhere? Its type is None anyway.

florentc (approved):
This is a great improvement of our tests! I can't wait to see them fully migrated to pytest and using fixtures!

What do you have in mind when we need for instance 2 derivations in a given test? Is there a way to use fixture params to use 2 different instances in a given test case, or is it only to generate more cases and we will have to define another fixture?

src/shared/tests/conftest.py
fricklerhandwerk: > Just to confirm, we move fixtures to fixtures.py so that we reference them in both conftest.py from shared and webview right?

Yes.

> Could we reference conftest.py from shared in the conftest.py from webview (as "shared" may hint that the conf of webview would extend it)

Not directly, because pytest operates on directories. Those `pytest_plugins` are the way to communicate fixtures between modules.

src/shared/tests/fixtures.py
fricklerhandwerk: It's a `pytest-django` builtin that initialises a temporary DB as a side effect, which is why its return type is `None` (fixtures are called by `pytest` when their names appear in an argument). The `db` fixtures allows doing queries in the first place, and we don't need to call it explicitly in dependent fixtures again.

PR #701, opened 2026-01-05
perf: use `CompositePrimaryKey` to avoid redundant indexes

fricklerhandwerk:
closes https://github.com/NixOS/nix-security-tracker/issues/661

- requires bumping to Django 5
- requires manual intervention on the DB to avoid stopping the world for ~20-30 min


Erethon (approved):
LGTM as far as I understand. I'm running the `create index concurrently` part on the production instance right now, this shouldn't be merged to the production branch until the index creation has finished running.

PR #712, opened 2026-01-12
feat: Set a default server block for nginx

Erethon:
This prevents connections from making it to the Django application without a proper host header.

You can test this on staging where it's already applied by running:
```
curl -k -v https://188.245.41.195
*   Trying 188.245.41.195:443...
* ALPN: curl offers h2,http/1.1
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* SSL Trust: peer verification disabled
* TLSv1.3 (IN), TLS alert, unrecognized name (624):
* TLS connect error: error:0A000458:SSL routines::tlsv1 unrecognized name
* closing connection #0
curl: (35) TLS connect error: error:0A000458:SSL routines::tlsv1 unrecognized name
```

Same thing on production succeeds and you get a reply from the Django application:
```
curl -k https://91.99.31.214

<!doctype html>
<html lang="en">
<head>
  <title>Bad Request (400)</title>
</head>
<body>
  <h1>Bad Request (400)</h1><p></p>
</body>
</html>
```

### DIFF
--- a/nix/web-security-tracker.nix
+++ b/nix/web-security-tracker.nix
@@ -182,6 +182,11 @@ in
 
       nginx.enable = true;
       nginx.virtualHosts = {
+        "_" = {
+          default = true;
+          rejectSSL = true;
+          locations."/".return = 200;
+        };
         ${cfg.domain} = {
           locations = {
             "/".proxyPass = "http://localhost:${toString cfg.wsgi-port}";


nix/web-security-tracker.nix
```diff
@@ -182,6 +182,10 @@ in
 
       nginx.enable = true;
       nginx.virtualHosts = {
+        "_" = {
+          default = true;
+          rejectSSL = true;
```
fricklerhandwerk, line 187: What if we drop the connection immediately instead of serving the dummy page?

```suggestion
          rejectSSL = true;
          locations."/".return = 444;
```

or

```suggestion
          rejectSSL = true;
          locations."/".return = 200;
```

https://nginx.org/en/docs/http/ngx_http_rewrite_module.html#return

fricklerhandwerk:
Please add the rationale to at least the commit message, otherwise it's hard to reverse why this is in place - or You Just Have Know.

nix/web-security-tracker.nix
Erethon: Both should work, but I don't see much difference between serving a default page or returning a 200/444/something else for http://ip. The important part is that http(s)://domain works in all cases and no random host headers make it to the backend. If you have a preference between 200/444 I'll commit the suggestion.

fricklerhandwerk: No particular preference... 200 I guess?

Erethon: 200 it is.

PR #714, opened 2026-01-12
Suggestions refactor

florentc:
Closes https://github.com/NixOS/nix-security-tracker/issues/674

### DIFF
--- a/src/webview/suggestions/urls.py
+++ b/src/webview/suggestions/urls.py
@@ -27,9 +27,7 @@ urlpatterns = [
         UntriagedSuggestionsView.as_view(),
         name="untriaged_suggestions",
     ),
-    path(
-        "accepted/", AcceptedSuggestionsView.as_view(), name="accepted_suggestions"
-    ),
+    path("accepted/", AcceptedSuggestionsView.as_view(), name="accepted_suggestions"),
     path(
         "dismissed/",
         RejectedSuggestionsView.as_view(),


src/shared/listeners/cache_suggestions.py
```diff
@@ -80,7 +80,7 @@ class Package(BaseModel):
     # Using lists of actual NixMaintainer model would be great
     # Alternatively use a typed dict like the Maintainer defined in shared.logs.events
     class CategorizedMaintainers(BaseModel):
-        original: list[dict]  # Maintainers of original packages
+        original: list[dict]  # Maintainers of original maintainers
```
fricklerhandwerk: Didn't this make it wronger?

fricklerhandwerk:
```suggestion
        original: list[dict]  # Maintainers of original packages
```

src/shared/listeners/cache_suggestions.py
```diff
@@ -414,3 +424,48 @@ def maintainers_list(packages: dict, edits: list[MaintainersEdit]) -> list[dict]
...
+
+    # Process edits to categorize maintainers
+    removed_github_ids = set()
+    added_maintainers = []
+
+    for edit in maintainers_edits:
+        if edit.edit_type == MaintainersEdit.EditType.REMOVE:
+            removed_github_ids.add(edit.maintainer.github_id)
+        elif edit.edit_type == MaintainersEdit.EditType.ADD:
+            added_maintainers.append(to_dict(edit.maintainer))
```
fricklerhandwerk, line 454: What happens to a maintainer that was removed and then added back? Won't it end up in both `ignored` and `active`?

src/shared/github.py
```diff
@@ -127,6 +130,7 @@ def additional_comment() -> str:
 {comment}
 {fence_backticks}"""
         else:
+            logger.error("####### 6 #######")
```
fricklerhandwerk:
```suggestion
```

Artefact?

src/webview/suggestions/context/builders.py
```diff
@@ -0,0 +1,93 @@
...
+    PackageListContext,
+)
+
+
+def is_suggestion_editable(suggestion: CVEDerivationClusterProposal) -> bool:
+    """Whether packages and maintainers can be edited depending on the suggestion status"""
+    return suggestion.status in [
+        CVEDerivationClusterProposal.Status.PENDING,
+        CVEDerivationClusterProposal.Status.ACCEPTED,
+    ]
```
fricklerhandwerk: This should be a [model method](https://docs.djangoproject.com/en/6.0/topics/db/models/#model-methods), it's part of the business logic, not presentation.

src/webview/suggestions/views/base.py
```diff
@@ -0,0 +1,137 @@
...
+            return self._redirect_to_origin(request)
+
+    def _redirect_to_origin(self, request: HttpRequest) -> HttpResponse:
+        """Redirect to the origin URL or fallback to suggestions list."""
+        # Get the current URL from HTMX header or referer
+        current_url = self._get_origin_url(request)
+        if not current_url:
+            # Fallback to suggestions list if no origin provided
+            return redirect("webview:suggestion:untriaged_suggestions")
+        return redirect(current_url)
```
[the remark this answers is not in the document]
fricklerhandwerk: This is a legitimate problem. It should be safe to use `HX-Current-URL` if we obtain it from a POST request with csrf token, but in the no-JS case we'd need to pass it explicitly.

src/webview/suggestions/views/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+        # Get suggestion context
+        suggestion = fetch_suggestion(suggestion_id)
+        suggestion_context = get_suggestion_context(suggestion)
+
+        # Validate that the suggestion status allows package editing
+        if suggestion.status not in [
+            CVEDerivationClusterProposal.Status.PENDING,
+            CVEDerivationClusterProposal.Status.ACCEPTED,
+        ]:
```
fricklerhandwerk: There would already be a method for this (see [the other comment](https://github.com/NixOS/nix-security-tracker/pull/714/changes#r2745429007))

src/webview/suggestions/views/maintainers.py
```diff
@@ -0,0 +1,366 @@
...
+
+    @abstractmethod
+    def _validate_operation(
+        self, suggestion: CVEDerivationClusterProposal, github_id: int
+    ) -> str | None:
+        """Validate if the operation can be performed. Return error message or None."""
+        pass
+
+    @abstractmethod
+    def _get_operation_name(self) -> str:
```
fricklerhandwerk, line 77:
```suggestion
    @property
    @abstractmethod
    def operation(self) -> str:
```

Then you can just use `self.operation`

src/webview/suggestions/views/base.py
```diff
@@ -0,0 +1,137 @@
...
+
+    template_name = "suggestions/components/suggestion.html"
+
+    class ForbiddenOperationError(Exception):
+        """Raised when access is denied for content editing."""
+
+        def __init__(self, response: HttpResponse) -> None:
+            self.error = response
+
+    def _check_access_rights_and_get_suggestion(
```
fricklerhandwerk, line 114: We should use Django's auth system here instead. Rough Idea:

- Create a mixin for [testing a condition](https://docs.djangoproject.com/en/6.0/topics/auth/default/#django.contrib.auth.mixins.UserPassesTestMixin)
  
  ```python
  class GroupRequiredMixin(UserPassesTestMixin):
      allowed_groups: list[str] = []
      
      def test_func(self) -> bool:
          return self.request.user.groups.filter(name__in=self.allowed_groups).exists()
  ```

- Use the mixin on the views (or base classes)
  ```python
  from django.conf import settings
 
  class MaintainerOperationBaseView(SuggestionContentEditBaseView, ABC):
    allowed_groups = [settings.DB_SECURITY_TEAM]
  ```

- Figure out what to do in case access is denied. From what I can tell, we just need to use `AccessMixin` with `raise_exception = True` ([docs](https://docs.djangoproject.com/en/6.0/topics/auth/default/#redirecting-unauthorized-requests-in-class-based-views))

Since you've split up the POST handling into separate views, no additional messing with edge cases should be needed.

src/webview/suggestions/views/lists.py
```diff
@@ -0,0 +1,69 @@
...
+from django.core.paginator import Paginator
+
+from shared.models.linkage import (
+    CVEDerivationClusterProposal,
+)
+
+from .base import SuggestionBaseView, get_suggestion_context
+
+
+class SuggestionListView(SuggestionBaseView, ABC):
```
fricklerhandwerk, line 13: Why is this not a `ListView`? Then we don't have to manually paginate

src/webview/suggestions/views/base.py
```diff
@@ -0,0 +1,137 @@
...
+        self, request: HttpRequest, suggestion_id: int
+    ) -> tuple[CVEDerivationClusterProposal, SuggestionContext]:
+        if not request.user or not can_publish_github_issue(request.user):
+            raise self.ForbiddenOperationError(HttpResponseForbidden())
+
+        # Get suggestion context
+        suggestion = fetch_suggestion(suggestion_id)
```
fricklerhandwerk, line 121: Those should not exist. There are [editing views](https://docs.djangoproject.com/en/6.0/ref/class-based-views/generic-editing/) we can inherit from. Specifically:

```python
from django.views.generic.edit import UpdateView

class SuggestionContentEditBaseView(UpdateView, SuggestionBaseView, ABC):
  http_method_names = ["post"]  # only accept POST requests for editing
```

This will provide you with `self.object` in your override of `post()`.

src/webview/suggestions/views/lists.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # Pagination first
+        paginator = Paginator(suggestions, self.paginate_by)
+        page_number = self.request.GET.get("page", 1)
+        page_obj = paginator.get_page(page_number)
+
+        # Convert suggestions to SuggestionContext objects for the current page
+        suggestion_contexts = []
+        for suggestion in page_obj.object_list:
+            suggestion_context = get_suggestion_context(suggestion)
```
fricklerhandwerk, line 38: This is substantially slower than before, because we're not batching all the related things. It just loops and loops over maintainers and events, due to the activity logs.

Image

Image

It was fetching all the maintainer names inefficiently before though, so the query time increase is "only" ~2x for this change.

src/webview/suggestions/urls.py
```diff
@@ -0,0 +1,81 @@
...
+app_name = "suggestion"
+
+urlpatterns = [
+    # Individual suggestion detail page
+    path("<int:suggestion_id>/", SuggestionDetailView.as_view(), name="detail"),
+    # Lists
+    path(
+        "list/untriaged/",
+        UntriagedSuggestionsView.as_view(),
+        name="untriaged_suggestions",
```
fricklerhandwerk:
```suggestion
    path("by-id/<int:suggestion_id>/", SuggestionDetailView.as_view(), name="detail"),
    # Lists
    path(
        "untriaged/",
        UntriagedSuggestionsView.as_view(),
        name="untriaged_suggestions",
```

How about this? Then we can later add a detail view by CVE.

src/webview/urls.py
```diff
@@ -22,6 +16,7 @@
     path("", HomeView.as_view(), name="home"),
     path("notifications/", include("webview.notifications.urls")),
     path("subscriptions/", include("webview.subscriptions.urls")),
+    path("suggestion/", include("webview.suggestions.urls")),
```
fricklerhandwerk:
```suggestion
    path("suggestions/", include("webview.suggestions.urls")),
```

See [related comment](https://github.com/NixOS/nix-security-tracker/pull/714/changes#r2745794916)

src/webview/suggestions/views/base.py
```diff
@@ -0,0 +1,137 @@
...
+        package_list_context=get_package_list_context(suggestion),
+        maintainer_list_context=get_maintainer_list_context(suggestion),
+        activity_log=fetch_activity_log(suggestion.pk),
+    )
+
+
+class SuggestionBaseView(TemplateView, ABC):
+    """Base view for all suggestion-related views with common functionality."""
+
+    def _handle_error(
```
fricklerhandwerk, line 47: I'm torn on this one. I understand that we need to render partials for HTMX errors, but really we shouldn't do explicit error responses. Ideally we'd just throw (default or custom) exceptions and handle them in a middleware.

For instance, user permissions (see [other comment](https://github.com/NixOS/nix-security-tracker/pull/714/changes#r2745571164)) are simply HTTP API hardening, so they can throw a dumb 403. The status change check can do the same -- there should be no UI to modify a frozen suggestion, and the only thing where that can happen under normal circumstances is a race condition between users or browser tabs. It would be nice to show a hint in context instead of an error page in that case, but it's not vital IMO.

Probably `SuggestionContentEditBaseView` should provide a `post()` base method along these lines:

```python
def post(self, request, *args, **kwargs):
    self.object = self.get_object()
    if not self.object.is_editable: # see other comment, that should be a model @property
        raise ForbiddenOperationError(
            message=f"Cannot edit {self.object.status} suggestion. The suggestion may have been {self.object.status} by a different user or in a different browser tab since you last viewed it.",
            suggestion=self.object,
            template_name=self.template_name,
        )
    return super().post(request, *args, **kwargs)
```

then have a middleware that will catch these

```python
# settings.py
MIDDLEWARE = [
    ...
    'django.contrib.messages.middleware.MessageMiddleware',
    'webview.middleware.ForbiddenOperationMiddleware',
    ...
]
```

```python
# shared/middleware.py
class ForbiddenOperationMiddleware:
    def __init__(self, get_response):
        self.get_response = get_response

    def __call__(self, request):
        return self.get_response(request)

    def process_exception(self, request, exception):
        if not isinstance(exception, ForbiddenOperationError):
            return None

        if request.headers.get("HX-Request") and exception.suggestion_id:
            context = get_suggestion_context(self.suggestion)
            context.error_message = exception.message
            return render(request, exception.template_name, {"data": context})
        else:
            messages.error(request, exception.message)
            return redirect(request.META.get("HTTP_REFERER", "/"))
```

We may have to look out for not fetching the related data twice depending on where the exception is thrown. So possibly the exception would have an optional `context` field that we check for here.

src/webview/suggestions/views/maintainers.py
```diff
@@ -0,0 +1,366 @@
...
+            return e.response
+
+        # Validate the requested operation
+        validation_error = self._validate_operation(suggestion, github_id)
+        if validation_error:
+            return self._handle_error(request, suggestion_context, validation_error)
+
+        # Perform the specific operation
+        try:
+            self._perform_operation(suggestion, github_id)
```
fricklerhandwerk, line 42: These should be methods on the (cached?) suggestion model, since this is business logic and shouldn't live in the views. Then the validation can happen there directly, and throw an appropriate exception. The exception would be handled by a middleware on the way out (see [related comment](https://github.com/NixOS/nix-security-tracker/pull/714/changes#r2745924815)).

src/shared/listeners/cache_suggestions.py
fricklerhandwerk: Ah okay, restoring a maintainer removes the `REMOVE` edit. The activity log is kept separately of that.

src/webview/suggestions/views/base.py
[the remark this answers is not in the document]
fricklerhandwerk: We'll leave it like this though, and consider dropping the no-JS workflow.

PR #728, opened 2026-01-15
fix: Typo on an f-string in tests

Erethon:
I believe the intention here was for this to be a f-string. Practically I don't think this changes much, since I don't see `derivation_path` being used elsewhere in tests, but let's fix it eitherway.


fricklerhandwerk:
Yes, thank you!

PR #746, opened 2026-01-21
feat: Email admins when an error occurs in Django

Erethon:
On the production instance we'll now be sending emails to `ADMINS` (which the secret value has my email and @fricklerhandwerk's) when a python error occurs. This is using the nixos.org email server with the account created [here](https://github.com/NixOS/infra/pull/919).


src/project/settings.py
```diff
@@ -149,6 +152,14 @@ class DjangoSettings(BaseModel):
...
             default=False,
         )
+        EMAIL_BACKEND: str = "django.core.mail.backends.console.EmailBackend"
+        EMAIL_HOST: str = "localhost"
+        EMAIL_PORT: int = 25
+        EMAIL_TIMEOUT: int = 10
+        EMAIL_HOST_USER: str = ""
+        EMAIL_USE_SSL: bool = False
+        DEFAULT_FROM_EMAIL: str = ""
+        SERVER_EMAIL: str = DEFAULT_FROM_EMAIL
```
fricklerhandwerk: I don't think you can access that value here though

fricklerhandwerk: You probably need to do something like

```python
SERVER_EMAIL: str = ""

@model_validator(mode='after')
def default_server_email(self):
    if not self.SERVER_EMAIL:
        self.SERVER_EMAIL = self.DEFAULT_FROM_EMAIL
    return self
```

Erethon: I believe you're correct. In my case it was working because the SERVER_EMAIL is set to an empty string this way and the Django code [fallsback automatically to `DEFAULT_FROM_EMAIL`](https://github.com/django/django/blob/3ea659dc1b8d72d3cab2de17d28cd9d2d06d6eed/django/core/mail/message.py#L246).

Erethon: Tested and works fine, nice catch, thank you.

PR #755, opened 2026-01-23
refactor: migrate suggestion editing tests to Playwright

fricklerhandwerk:
Towards https://github.com/NixOS/nix-security-tracker/pull/714


src/webview/tests/test_suggestions.py
```diff
@@ -31,6 +35,53 @@
...
+    drv1 = make_drv(pname="package1")
+    drv2 = make_drv(pname="package2")
+    suggestion = make_suggestion(
+        status=status,
+        drvs={
+            drv1: ProvenanceFlags.PACKAGE_NAME_MATCH,
+            drv2: ProvenanceFlags.PACKAGE_NAME_MATCH,
+        },
+    )
+    cache_new_suggestions(suggestion)
```
fricklerhandwerk: We could have a `make_cached_suggestion` fixture...

src/webview/tests/test_suggestions.py
```diff
@@ -31,6 +35,53 @@
...
+        as_staff.locator('input[value="package1"]').click()
+        if no_js:
+            purge = as_staff.get_by_role("button", name="Purge deleted packages")
+            purge.click()
+        else:
+            as_staff.reload()
+
+    expect(
+        as_staff.get_by_text("Matching in Nixpkgs").get_by_text("package1")
+    ).to_have_count(0)
```
fricklerhandwerk: We should check that package2 is still there

Issue #689, opened 2025-12-16
Remove maintainers of removed packages automatically


fricklerhandwerk:
As a security team member, I don't want to manually track which maintainers are obsolete after removing invalid packages.

## Acceptance criteria

Given an automatic suggestion with multiple packages that don't share maintainers,
When I remove one of the packages,
Then the maintainers that belong to that package (and don't belong to other, retained packages) are also removed from the notification list.

PR #766, opened 2026-01-30
fix: don't notify maintainers not belonging to active packages 

florentc:
This addresses part of #689 : it implements the backend part, but not the UX/view part. Instead, this shows a message in the maintainers lists.

Testing is blocked for now: we need to find a nice way to mock the github object because we want to check content of the generated github issues.

### DIFF
--- a/src/webview/templates/suggestions/components/maintainers_list.html
+++ b/src/webview/templates/suggestions/components/maintainers_list.html
@@ -6,10 +6,13 @@
     class="rounded-box column gap"
     {% if oob_update %}hx-swap-oob="true" open{% endif %}
   >
-    <h2 class="heading dimmed">Package maintainers</h2>
 
     <div class="column gap">
       <!-- Active maintainers -->
+      <div class="row gap spread wrap baseline">
+        <h2 class="heading dimmed">Package maintainers</h2>
+        {% if data.editable and user|is_maintainer_or_admin %}<p class="dimmed">Maintainers of ignored packages won't be notified.</p>{% endif %}
+      </div>
       {% if data.active %}
         <ul id="maintainers-list-{{ data.suggestion_id }}-active" class="column gap-small">
           {% for maintainer_context in data.active %}
--- a/src/webview/tests/test_issues.py
+++ b/src/webview/tests/test_issues.py
@@ -1,10 +1,14 @@
 from collections.abc import Callable
+from typing import Any
 from unittest.mock import patch
 
 import pytest
 from django.urls import reverse
+from github import Github
+from github.Issue import Issue as GithubIssue
 from playwright.sync_api import Page, expect
 from pytest_django.live_server_helper import LiveServer
+from pytest_mock import MockerFixture
 
 from shared.github import create_gh_issue
 from shared.listeners.cache_suggestions import cache_new_suggestions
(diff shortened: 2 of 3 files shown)


src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    no_js: bool,
+    mocker: MockerFixture,
+    ignore_package: bool,
+) -> None:
+    """Test that the body of a created issue mentions the maintainer, unless the package has been ignored."""
+
+    container = make_container(title="Dummy Title", description="Test description")
+    accepted_suggestion = make_suggestion(
+        container=container, status=CVEDerivationClusterProposal.Status.ACCEPTED
+    )
```
fricklerhandwerk: [simplification] We're not doing anything with the container and it doesn't need to be anything specific

```suggestion
    make_suggestion: Callable[..., CVEDerivationClusterProposal],
    live_server: LiveServer,
    as_staff: Page,
    no_js: bool,
    mocker: MockerFixture,
    ignore_package: bool,
) -> None:
    """Test that the body of a created issue mentions the maintainer, unless the package has been ignored."""

    accepted_suggestion = make_suggestion(
        status=CVEDerivationClusterProposal.Status.ACCEPTED
    )
```

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    cache_new_suggestions(accepted_suggestion)
+
+    as_staff.goto(live_server.url + reverse("webview:suggestion:accepted_suggestions"))
+    suggestion = as_staff.locator(f"#suggestion-{accepted_suggestion.cached.pk}")
```
fricklerhandwerk: [Correctness/consistency] Interesting, are we still indexing over the cached suggestion? I thought this should be


```suggestion
    suggestion = as_staff.locator(f"#suggestion-{accepted_suggestion.pk}")
```

Or does it only work by accident because we only have one suggestion and cached suggestion each?

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    publish = suggestion.get_by_role("button", name="Publish issue")
+    active_packages = as_staff.locator(
+        f"#suggestion-{accepted_suggestion.cached.pk}-active-packages"
+    )
+    ignore_package_button = active_packages.get_by_role("button", name="Ignore")
+
+    # Case where we ignore the package
+    if ignore_package:
+        expect(active_packages.get_by_text("foo")).to_be_visible()
```
fricklerhandwerk: [consistency/maintainability]

```suggestion
        drv = accepted_suggestion.derivations.first()
        assert drv
        expect(active_packages.get_by_text(drv.attribute)).to_be_visible()
```

Otherwise you have to Just Know™ the implementation details of the fixture

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+        # NOTE(@florentc): Necessary to avoid race condition on clicking "View" later
+        expect(active_packages.get_by_text("foo")).not_to_be_visible()
+
+    # Mocking GitHub
+    mock_repo = mocker.Mock()
+    mock_issue = mocker.Mock()
+    mock_issue.html_url = "https://fake.url"  # NOTE(@florentc): We need to define it because the view expects it
+    mock_repo.create_issue.return_value = mock_issue
+    mock_github = mocker.Mock()
+    mock_github.get_repo.return_value = mock_repo
```
fricklerhandwerk, line 139: [optional] Maybe all inline?

```suggestion
    mock_github = mocker.Mock(
        get_repo=mocker.Mock(
            return_value=mocker.Mock(
                create_issue=mocker.Mock(
                    return_value=mocker.Mock(html_url="https://fake.url"),
                ),
            ),
        ),
    )
```

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+
+    # NOTE(@florentc): We can't mock get_gh with patch because it is evaluated
+    # in module (as a default param) so we mock the entire create_gh_issue to
+    # which we pass our mock github object
```
fricklerhandwerk, line 143:
```suggestion
```

It's pretty clean/obvious that way, no need to apologize :)

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    def mock_create_gh_issue(*args: Any, **kwargs: Any) -> GithubIssue:
+        kwargs["github"] = mock_github
+        return create_gh_issue(*args, **kwargs)
```
fricklerhandwerk: Why not

```suggestion
        return create_gh_issue(*args, github=mock_github, **kwargs)
```

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+
+    # NOTE(@florentc): During issue creation, info of maintainers is refreshed
+    # by querying the GitHub API, we mock it to use what we have already. Since
+    # we want to check who ends up in the generated list of maintainers in the
+    # issue body, we can't just mock a constant return value for everyone
+    def mock_get_maintainer_username(
+        maintainer: dict, github: Github = mock_github
```
fricklerhandwerk, line 152:
```suggestion
        maintainer: dict, **kwargs: Any
```

You're not using it anyway -- unless you plan to?

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    ) -> str:
+        return maintainer["github"]
+
+    mocker.patch(
+        "webview.suggestions.views.status.create_gh_issue", mock_create_gh_issue
+    )
+    mocker.patch("shared.github.get_maintainer_username", mock_get_maintainer_username)
+
+    # Publish the issue
```
fricklerhandwerk, line 161:
```suggestion
```

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    publish.click()
+
+    if no_js:
+        as_staff.goto(live_server.url + reverse("webview:issue_list"))
+    else:
+        link = suggestion.get_by_role("link", name="View")
+        link.click()
+    # NOTE(@florentc): We wait to be sure publication has happened
```
fricklerhandwerk, line 169:
```suggestion
```

This is obvious IMO

src/webview/tests/test_issues.py
```diff
@@ -88,3 +92,88 @@ def test_publish_gh_issue_empty_title(
...
+    expect(suggestion).to_be_visible()
+
+    mock_repo.create_issue.assert_called_once()
+    issue_body = mock_repo.create_issue.call_args[1]["body"]
+
+    if ignore_package:
+        assert "@maintainer" not in issue_body
```
fricklerhandwerk: That's really a neat test!

src/webview/tests/test_issues.py
fricklerhandwerk: Ah I see, later you use `mock_repo` for assertions. So maybe leave that outside or something. That suggestion was optional anyway

src/webview/templates/suggestions/components/maintainers_list.html
```diff
@@ -6,10 +6,13 @@
     class="rounded-box column gap"
     {% if oob_update %}hx-swap-oob="true" open{% endif %}
   >
-    <h2 class="heading dimmed">Package maintainers</h2>
 
     <div class="column gap">
       <!-- Active maintainers -->
+      <div class="row gap spread wrap baseline">
+        <h2 class="heading dimmed">Package maintainers</h2>
+        <p class="dimmed">Package maintainers whose packages above are all ignored won't be notified.</p>
```
fricklerhandwerk: [optional] I know yours is more precise...

```suggestion
        <p class="dimmed">Maintainers of ignored packages won't be notified.</p>
```

fricklerhandwerk (approved):
Does the job to fix the issue. A few small things for readability/consistency, and a tangential semantics/correctness question

src/webview/templates/suggestions/components/maintainers_list.html
florentc: Mine is more exact but yours is enough to convey the meaning through common sense. I rephrased it and, more importantly, made it only visible when maintainers can be edited.

src/webview/tests/test_issues.py
florentc: This is a remnant from the other test we have in the module, that I used as a starting point to this one.

src/webview/tests/test_issues.py
florentc: Indeed. Gonna update that. Same for maintainer handle.

src/webview/tests/test_issues.py
florentc: Haha not an apology, just a note in case I wonder if the future why we don't mock `get_gh()` directly

src/webview/tests/test_issues.py
florentc: Why not indeed

src/webview/tests/test_issues.py
florentc: I don't plan to, this was to match the same visible interface as the original function. Happy to remove the unused argument

src/webview/tests/test_issues.py
florentc: Why not? I added it because it was not immediate to find where exactly we publish (hence where the actual interesting assertions start) in the other test and this one

src/webview/tests/test_issues.py
florentc: This was needed to avoid race conditions. I thought it's useful to tag those important "expects" for the execution

src/webview/tests/test_issues.py
florentc: In fact it's needed. Got a crash after trying to remove it.

src/webview/tests/test_issues.py
florentc: Done, remnant from other test which served as a basis

src/webview/tests/test_issues.py
fricklerhandwerk: It does say `publish.click()` :shrug:

Issue #332, opened 2024-11-11
Only display root of dependency graph in suggestions


fricklerhandwerk:
As a security team member, in order to minimize cognitive load, I want to see matching suggestions only for directly affected packages (and at most the number of reverse dependencies, or maybe a list/tree of them if I really care). Fixing the root fixes the others.

Currently we show all packages that have a name matching, but many of them turn out to be dependants of the actual thing in question.

![image](https://github.com/user-attachments/assets/428289b9-ddd3-4e3e-8873-4dab5a91a65b)

PR #787, opened 2026-02-06
perf: use zram for tmpfs

fricklerhandwerk:
With the pervious setup evaluations ran into both full tmpfs and OOM.
Chances are, this setup will give us more RAM to work with and more space to keep store derivatons in memory.

### DIFF
--- a/infra/staging.nix
+++ b/infra/staging.nix
@@ -1,7 +1,6 @@
 {
   config,
   pkgs,
-  lib,
   ...
 }:
 let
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Not writing anything to disk at the moment. Depending on what may get done first, ideally we'd track drv dependencies without hitting disk (upstream Nix change) or obtain evaluations from Hydra (substantial Hydra change). But in any case we only need that when we start looking at dependencies as in #332.

Issue #242, opened 2024-10-02
Use consistent name for PostgreSQL database across project


erictapen:
Found values so far:

- web-security-tracker
- nix-security-tracker

I'd vote for nix-security-tracker, as it's also the repository name.

fricklerhandwerk:
Yes, let's use `nix-security-tracker` everywhere.

RaitoBezarius:
This needs a rename of our staging deployment's DB, I can take care of it.

erictapen:
Yeah that would be great, because the dump has the name of the role all over the place. I used sed in the past to be able to restore a dump to my local dev database.

PR #806, opened 2026-02-16
docs: update instructions to obtain a database dump

fricklerhandwerk:
Align database names between remote and local databases.
Add `pv` to development shell so one doesn't have to wrap everything in an extra shell for the occasion.
Add instructions for getting SSH access to staging and production, make access explicit.

> [!WARNING]
>   This requires downtime and manual steps for deployment!
>   - Move the state directory with the Nixpkgs checkout
>   - Rename the database

Closes https://github.com/NixOS/nix-security-tracker/issues/242

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -142,13 +142,13 @@ Or set up [`nix-direnv`](https://github.com/nix-community/nix-direnv) on your sy
 ### Set up a local database
 
 Currently only [PostgreSQL](https://www.postgresql.org/) is supported as a database.
-You can set up a database on NixOS like this:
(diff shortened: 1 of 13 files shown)


Erethon:
According to #242 we want to move to `nix-security-tracker` (and I agree that that name makes more sense), so maybe we should go the other way and finally rename the database and users in staging/prod instead of changing to `web-security-tracker` in the dev setup?

fricklerhandwerk:
I was thinking that but then backed off because I'd need to shut down production for a moment. Maybe we should just do that...

fricklerhandwerk:
Did the following on staging before merging:
```
systemctl stop web-security-tracker-{server,worker,evaluator}
psql -U web-security-tracker
web-security-tracker=> ALTER DATABASE web-security-tracker RENAME TO nix-security-tracker
```

fricklerhandwerk:
Phew, that went through.

PR #818, opened 2026-02-22
fix: deduplicate Description and Reference records on CVE ingestion

### LINKED ISSUE #710: Deduplicate CVE metadata
We're creating supposedly small pieces of data indiscriminately:

https://github.com/NixOS/nix-security-tracker/blob/aca950262fed92687e90b26ea32d8d7c2819024a/src/shared/fetchers.py#L46-L52

https://github.com/NixOS/nix-security-tracker/blob/aca950262fed92687e90b26ea32d8d7c2819024a/src/shared/fetchers.py#L63-L71

This leads to a waste of some hundreds MB in storage for descriptions and  (likely a bit less due to internal compression):

```python
In [4]: duplicates = Description.objects.values('lang', 'value').annotate(count=Count('id'), value_length=Length('value')).filter(count__gt=1).order_by('-count')

In [5]: wasted_bytes = duplicates.aggregate(total=Sum((F('count') - 1) * F('value_length')))['total']

In [6]: wasted_bytes
Out[6]: 216105026

In [7]: wasted_bytes / 1024 / 1024
Out[7]: 206.09381294250488
```

```python
In [8]: duplicates = Reference.objects.values('url', 'name').annotate(count=Count('id'), value_length=Length('url')).filter(count__gt=1).order_by('-count')

In [9]: wasted_bytes = duplicates.aggregate(total=Sum((F('count') - 1) * F('value_length')))['total']

In [10]: wasted_bytes / 1024 / 1024
Out[10]: 241.48347282409668
```

A bit worse, due to millions of such duplicates we have ~2GB of useless indexes:

```log
nix-security-tracker=# SELECT pg_size_pretty(pg_total_relation_size('shared_description_media'));
 pg_size_pretty 
----------------
 19 MB
(1 row)

nix-security-tracker=# SELECT pg_size_pretty(pg_total_relation_size('shared_description'));
 pg_size_pretty 
----------------
 909 MB
(1 row)

nix-security-tracker=# SELECT pg_size_pretty(pg_total_relation_size('shared_reference'));
 pg_size_pretty 
----------------
 648 MB
(1 row)

nix-security-tracker=# SELECT pg_size_pretty(pg_total_relation_size('shared_reference_tags'));
 pg_size_pretty 
----------------
 1042 MB
(1 row)
```

Bbn08:
Fixes #710

`make_description` and `make_reference` were calling `objects.create()` unconditionally, generating millions of duplicate rows on every ingestion run.

Switched both to `get_or_create()` keyed on the natural identity fields (`lang`+`value` and `url`+`name` respectively). M2M relations are only populated for newly created objects.

Also adds tests covering the deduplication behaviour.

### DIFF
--- a/src/shared/tests/test_fetchers.py
+++ b/src/shared/tests/test_fetchers.py
@@ -1,0 +1,26 @@
+import pytest
+
+from shared.fetchers import make_description, make_reference
+from shared.models.cve import Description, Reference
+
+
+@pytest.mark.django_db
+def test_make_description_deduplicates() -> None:
+    """Calling make_description twice with identical lang+value returns the same row."""
+    data = {"lang": "en", "value": "A test vulnerability description."}
+    first = make_description(data)
+    second = make_description(data)
+
+    assert first.pk == second.pk
+    assert Description.objects.filter(lang="en", value=data["value"]).count() == 1
+
+
+@pytest.mark.django_db
+def test_make_reference_deduplicates() -> None:
+    """Calling make_reference twice with identical url+name returns the same row."""
+    data = {"url": "https://example.com/advisory", "name": "Advisory"}
+    first = make_reference(data)
+    second = make_reference(data)
+
+    assert first.pk == second.pk
+    assert Reference.objects.filter(url=data["url"], name=data["name"]).count() == 1
--- a/src/shared/fetchers.py
+++ b/src/shared/fetchers.py
@@ -44,12 +44,12 @@ def make_media(data: dict[str, str]) -> models.SupportingMedia:
 
 
 def make_description(data: dict[str, Any]) -> models.Description:
-    ctx: dict[str, Any] = dict()
-    ctx["lang"] = data["lang"]
-    ctx["value"] = data["value"]
-
-    obj = models.Description.objects.create(**ctx)
-    obj.media.set(map(make_media, data.get("supportingMedia", [])))
+    obj, created = models.Description.objects.get_or_create(
+        lang=data["lang"],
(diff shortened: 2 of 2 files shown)


src/shared/tests/test_fetchers.py
```diff
@@ -0,0 +1,60 @@
+import pytest
+
+from shared.fetchers import make_cve, make_description, make_reference
+from shared.models.cve import CveRecord, Description, Reference
+
+
+def _cve_data(cve_id: str = "CVE-2024-9999", cna_tags: list[str] | None = None) -> dict:
```
fricklerhandwerk: Not critical, but for increased long-term maintainability:

```suggestion
@pytest.fixture
def make_cve_data() -> Callable[...,dict]:
    def wrapper(
        cve_id: str = "CVE-2024-9999", cna_tags: list[str] | None = None
     ) -> dict:
         return { ...
    return wrapper
```

This would be in line with existing fixtures, keeping conceptual variability down. Also, using fixtures (even fixture factories) instead of just calling procedures provides for slightly better test readability, since the test signature then essentially enumerates prerequisites. It's not a big deal with such a small test file, but uniformity will make a difference as the test suite grows.

Bbn08: Thanks for the suggestion and for taking the time to review! I've applied the fixture factory pattern in #819 where it was needed. For this PR the tests only use simple inline dicts that don't need to vary across test cases, so I kept them as it is to avoid unnecessary indirection. Happy to update if you feel strongly about it :)

PR #819, opened 2026-02-23
fix: skip CVEs tagged with exclusively-hosted-service during ingestion

### LINKED ISSUE #804: Skip CVE tagged with `exclusively-hosted-service`
These issues are never going to be something we can match, it is likely safe to just skip them.

Example: https://tracker.security.nixos.org/suggestions/by-id/2358/

Bbn08:
Fixes #804

CVEs tagged `exclusively-hosted-service` on their CNA container arenever applicable to Nixpkgs packages. Added a module level `_EXCLUDED_CNA_TAGS` frozenset and an early return in `make_cve()` to skip ingestion of such records entirely no DB rows created.

Adds tests covering both the skip and the normal ingest paths.

### DIFF
--- a/src/shared/listeners/automatic_linkage.py
+++ b/src/shared/listeners/automatic_linkage.py
@@ -105,6 +105,17 @@ def build_new_links(container: Container) -> bool:
         logger.info("Suggestion already exists for '%s', skipping", container.cve)
         return False
 
+    if container.tags.filter(value="exclusively-hosted-service").exists():
+        logger.info(
+            "Container for '%s' is exclusively-hosted-service, rejecting without match.",
+            container.cve,
+        )
+        CVEDerivationClusterProposal.objects.create(
+            cve=container.cve,
+            status=CVEDerivationClusterProposal.Status.REJECTED,
+        )
+        return True
+
     drvs = produce_linkage_candidates(container)
     if not drvs:
         logger.info("No derivations matching '%s', ignoring", container.cve)
--- a/src/shared/tests/test_linkage.py
+++ b/src/shared/tests/test_linkage.py
@@ -5,7 +5,7 @@ import pytest
 
 from shared.listeners.automatic_linkage import build_new_links
 from shared.listeners.cache_suggestions import cache_new_suggestions
-from shared.models.cve import Container
+from shared.models.cve import Container, Tag
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
     DerivationClusterProposalLink,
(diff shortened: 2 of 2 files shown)


src/shared/fetchers.py
```diff
@@ -275,11 +275,18 @@ def make_container(
...
+
 def make_cve(
     data: dict[str, Any],
     record: models.CveRecord | None = None,
     triaged: bool = False,
-) -> models.CveRecord:
+) -> models.CveRecord | None:
+    cna_tags = set(data["containers"]["cna"].get("tags", []))
+    if cna_tags & _EXCLUDED_CNA_TAGS:
+        return None
```
fricklerhandwerk: I'm not entirely convinced we should skip ingestion rather than skip presenting a suggestion. Of course it makes sense not to store the data if we're not planning to do anything with it. But previous design discussions were always leaning towards aggregating and annotating data rather than throwing it away, so we can inspect it after the fact and make more informed decisions.

From that angle, it seems more natural to me to somehow mark a CVE as processed and add a corresponding result flag. One way to do it is to just create a linkage record, set a "non-matching" flag and put the record in a "dismissed" state.

https://github.com/NixOS/nix-security-tracker/blob/a35e45b284b521be9629abce5cacdcd99ed0a1ad/src/shared/models/linkage.py#L204-L212

Also, architecturally, skipping ingestion is strictly worse because it distributes matching logic into unrelated aspects of the code.

Bbn08: Thanks for the explanation, I understand the concern skipping at ingestion violates the "aggregate and annotate" principle and leaks matching logic into the wrong layer. I'll rework this to instead create a rejected proposal in build_new_links() when the container is tagged exclusively-hosted-service, so the CVE is still stored and the decision is recorded explicitly. Please feel freee to correct if there any holes in my understanding.

fricklerhandwerk: You seem to have gotten it right, thanks for making sure!

Bbn08:
Reworked the fix based on your feedback. Instead of skipping ingestion in `make_cve()`, the `exclusively-hosted-service` check now lives in `build_new_links()` where it belongs the CVE is still ingested normally, but a REJECTED proposal with no derivations is immediately created to record the decision explicitly. Also reverted the `fetchers.py` change and moved the test to `test_linkage.py` to match where the logic now lives.

src/shared/listeners/automatic_linkage.py
```diff
@@ -105,6 +105,17 @@ def build_new_links(container: Container) -> bool:
...
         return False
 
+    if container.tags.filter(value="exclusively-hosted-service").exists():
+        logger.info(
+            "Container for '%s' is exclusively-hosted-service, rejecting without match.",
+            container.cve,
+        )
+        CVEDerivationClusterProposal.objects.create(
+            cve=container.cve,
+            status=CVEDerivationClusterProposal.Status.REJECTED,
```
fricklerhandwerk, line 115: Ah damn, we don't have the equivalent of `ProvenanceFlags` for the CVE end of matches, yet. We need to attach the reason to the rejected suggestion somehow.

Bbn08: Oh I can see the comment field exists for human notes but there's no structured equivalent to `ProvenanceFlags` for recording machine-generated rejection reasons on the CVE side. Would it make sense to track this as a separate issue? The scope feels broader than just this PR any auto-rejection in `build_new_links()` would benefit from a structured reason field. Happy to open one and implement it, just want to make sure we agree on the right approach before touching the model.

fricklerhandwerk: Exactly. As you wish, for me it's fine to discuss it here in comments, because we're staying close to the code. But an issue with acceptance criteria is just as fine.

Bbn08: Sounds good, I'll open a separate issue for the structured rejection reason field and reference it from a new PR shortly.

PR #823, opened 2026-02-24
refactor: simplify application logic internals

fricklerhandwerk:


src/webview/suggestions/urls.py
```diff
@@ -30,23 +26,22 @@
     # Lists
     path(
         "untriaged/",
-        UntriagedSuggestionsView.as_view(),
+        SuggestionListView.as_view(status_filter="pending"),
```
fricklerhandwerk, line 29: @florentc We should use the enum here

PR #825, opened 2026-02-24
feat: rejection reason field

### LINKED ISSUE #824: Add structured rejection reason to CVEDerivationClusterProposal
Currently `CVEDerivationClusterProposal` has no structured way to record 
why a proposal was auto-rejected. The `comment` field exists for human 
notes, but machine-generated rejections (e.g. from `build_new_links()`) 
have no equivalent to `ProvenanceFlags` on the derivation side.

## Proposed solution

Add a `rejection_reason` field using `TextChoices` since rejection reasons 
are mutually exclusive (unlike provenance flags which can combine).

## Acceptance criteria

- `CVEDerivationClusterProposal` has a nullable `rejection_reason` field 
  with at least `EXCLUSIVELY_HOSTED_SERVICE` as a choice
- `build_new_links()` populates this field when creating a REJECTED proposal
- Migration included

## Related

Came up during review of #819

Bbn08:
Closes #824

Adds a `RejectionReason` TextChoices field to `CVEDerivationClusterProposal`
so that machine-generated rejections carry a structured reason, equivalent
to how `ProvenanceFlags` records provenance on the derivation side.

- Adds `RejectionReason` inner class with `EXCLUSIVELY_HOSTED_SERVICE` choice
- Adds nullable `rejection_reason` field to the model
- Populates it in `build_new_links()` when rejecting hosted-service CVEs
- Updates test to assert the reason is set correctly
- Includes migration

Depends on #819.

### DIFF
--- a/src/shared/models/linkage.py
+++ b/src/shared/models/linkage.py
@@ -28,6 +28,12 @@ class CVEDerivationClusterProposal(TimeStampMixin):
         ACCEPTED = "accepted", _("accepted")
         PUBLISHED = "published", _("published")
 
+    class RejectionReason(models.TextChoices):
+        EXCLUSIVELY_HOSTED_SERVICE = (
+            "exclusively_hosted_service",
+            _("exclusively hosted service"),
+        )
+
     cached: "shared.models.cached.CachedSuggestions"
 
     cve = models.ForeignKey(
@@ -50,6 +56,14 @@ class CVEDerivationClusterProposal(TimeStampMixin):
         help_text=_(
             "Optional free text comment for additional notes, context, dismissal reason"
         ),
+    )
+
+    rejection_reason = models.CharField(
+        max_length=126,
+        choices=RejectionReason.choices,
+        null=True,
+        blank=True,
+        help_text=_("Machine-generated reason for automatic rejection"),
     )
 
     @property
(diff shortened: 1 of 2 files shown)


src/shared/tests/test_fetchers.py
fricklerhandwerk, line 1: Please rebase, this one is merged in #818

src/shared/models/linkage.py
```diff
@@ -52,6 +57,14 @@ class Status(models.TextChoices):
         ),
     )
 
+    rejection_reason = models.CharField(
+        max_length=text_length(RejectionReason),
```
fricklerhandwerk: This is reasonable but turns out to be brittle. While in practice this particular string is rather long, what would typically happen is that one can add a longer string and then we suddenly need a migration.

```suggestion
        max_length=126,
```

https://www.postgresql.org/docs/current/datatype-character.html

> The storage requirement for a short string (up to 126 bytes) is 1 byte plus the actual string, which includes the space padding in the case of character. Longer strings have 4 bytes of overhead instead of 1.

So this is actually rather efficient.

Bbn08: Noted, thanks for the change

src/shared/tests/test_fetchers.py
Bbn08: Rebase done, removed the commit from #818.

PR #826, opened 2026-02-25
fix: show issue code inline with CVE metadata in published issues

### LINKED ISSUE #616: Task: Show CVE description and as in draft issues
Right now we're showing the CVE description but under our own issue ID:

Image

We want this to look the same even if the underlying data model is different (for historic reasons, we'll clean this up as we go).

Image

Then the issue ID needs a different place, but we can still face-lift this down the line.

Bbn08:
Closes #616

Published issues showed the internal issue code (NST-XXXX) as a prominent standalone header above the CVE description, making the layout inconsistent with draft issues (accepted suggestions) where the description appears immediately.

- Moves the issue code into the CVE metadata row in `suggestion.html` so both published and draft views look the same
- Removes the separate issue code header from `issue.html`, keeping the GitHub link and published date
- Adds optional `issue_code` field to `SuggestionContext`, populated by the `issue()` template tag

### DIFF
--- a/src/webview/suggestions/context/types.py
+++ b/src/webview/suggestions/context/types.py
@@ -82,6 +82,7 @@ class SuggestionContext:
         self.can_edit: bool = can_edit
         self.suggestion: CVEDerivationClusterProposal = suggestion
         self.suggestion_stub_context: SuggestionStubContext | None = None
+        self.issue_code: str | None = None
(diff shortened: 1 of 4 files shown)


fricklerhandwerk:
I'll close it as the original issue happened to be not well-maintained and solved by a previous change.

PR #843, opened 2026-03-02
Add florentc key for staging access

florentc:
This adds my key to authorized keys. I need access to get a fresh db dumps for my local development environment.

### DIFF
--- a/infra/common.nix
+++ b/infra/common.nix
@@ -53,7 +53,6 @@ in
       openssh.authorizedKeys.keyFiles = with keys; [
         fricklerhandwerk
         erethon
-        florentc
         security-tracker-gh-actions
       ];
       # We're using both keys and keyFiles here in order to keep some alignment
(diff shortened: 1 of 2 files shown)


infra/common.nix
```diff
@@ -53,6 +53,7 @@ in
       openssh.authorizedKeys.keyFiles = with keys; [
         fricklerhandwerk
         erethon
+        florentc
```
fricklerhandwerk: This is for both staging and production. Based on the PR title this seems like not intended. You'd need to add the entry to `staging.nix`

Erethon: After https://github.com/NixOS/nix-security-tracker/commit/5638f9f339f809004cc7d52b66c7bd80a2b00a32 the docs say you should make this change in `infra/staging.nix` to get access to staging only. ~~However, I don't think you can currently add this in staging only without overwriting the values in common as it currently stands.~~

I've applied this patch manually on staging so you can get a dump until we cleanly separate staging and production access.

Issue #840, opened 2026-03-01
Compact untriaged list


fricklerhandwerk:
As a security team member, I want a quick overview of untriaged suggestions.

## Acceptance criteria (TODO)

The envisioned workflow is to move good candidates to "accepted" after a quick glance, and continue editing there.

In the list view:
- Should only show the affected products
- Uncollapsed description; alternatively no description but only references (typically more informative)
- No packages or maintainers; alternatively collapse if more than some fixed N

We probably want all the details in the "accepted" list though, but this needs more elaboration in practice.

PR #848, opened 2026-03-05
Toggleable compact view for suggestions lists

florentc:
This proposes to address #840 

This introduces an optional query parameter `compact` on all list views: untriaged, dismissed, accepted, and even package search results.

And also links on list pages to switch between compact mode and normal classic mode:

By default, now the navigation bar and the home page link to untriaged and dismissed with `compact` enabled. Accepted suggestions and search result keep showing expanded lists by default. In any case, switching from one to the other is one click away.

## What compact view entails on suggestions:

- The description is removed
- The packages list is removed
- The maintainers list is removed
- Only the "affected products" are shown after the top line and title (which remain similar to classic view)
- The "affected products" title and rounded-box are hidden as it's the only info present
- The version list for each affected product is displayed in a row
- The gaps separating the various elements in the suggestions are set to `gap-small` instead of `gap`
- The status change buttons and comment section are hidden. **I think this is a topic for discussion:** we could display them to allow users to quickly dismiss or accept suggestions from the compact list without the need to switch to detail page or expanded list view. If so, I suggest that the comment section be compacted to a 1 line height and placed between "Dismiss" and "Accept" buttons to stay compact. Whether to display the buttons or not in compact view could also be a parameter like `compact`.

## Screenshots

<details>
<summary>Expanded view</summary>

</details>

<details>
<summary>Compact view</summary>

</details>

<details>
<summary>Compact view on search results (not by default)</summary>

</details>

### DIFF
--- a/src/webview/static/colors.css
+++ b/src/webview/static/colors.css
@@ -12,7 +12,4 @@
   --light-green: #dbfcdb;
   --light-yellow: #fcfaba;
   --light-gray: #eee;
-
-  --foreground: black;
-  --background: white;
 }
--- a/src/webview/static/utility.css
+++ b/src/webview/static/utility.css
@@ -24,24 +24,6 @@
   margin: 1em 0;
 }
 
-.small {
-  font-size: 0.75rem;
-}
-
-.tag {
-  border: solid 2px var(--background);
-  border-radius: 0.4em;
-  padding: 0.2rem 0.4rem;
-  outline-width: 2px;
-  word-break: normal;
-  display: inline-block;
-}
-
-.fill-gray {
-  color: var(--background);
-  background-color: var(--gray);
-}
-
 /* Buttons
  *
  * .btn - Base button style to be used on every standard button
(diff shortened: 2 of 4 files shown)


src/webview/templates/suggestions/suggestion_list.html
```diff
@@ -82,6 +82,17 @@ <h1 class="page-title row gap-small centered">{% status_icon "published" %}Publi
...
   {% endif %}
 
+  <div class="inline-row gap centered justify-right">
+    <div class="inline-row centered gap-small"><i class="icon-eye"></i>View:</div>
+    {% if is_compact %}
+      <div class="bold">Compact suggestions</div>
+      <a href="{% toggle_param 'compact' %}">Detailed suggestions</a>
+    {% else %}
+      <a href="{% toggle_param 'compact' %}">Compact suggestions</a>
+      <div class="bold">Detailed suggestions</div>
```
fricklerhandwerk:
```suggestion
      <div class="bold">Compact</div>
      <a href="{% toggle_param 'compact' %}">Detailed</a>
    {% else %}
      <a href="{% toggle_param 'compact' %}">Compact</a>
      <div class="bold">Detailed</div>
```

No need to be overly specific here

florentc: Done

PR #849, opened 2026-03-05
fix: only show cached suggestions in list, cache on demand in detail

### LINKED ISSUE #829: Race condition between suggestion creation and caching
We occasionally get an error of the sort

```
RelatedObjectDoesNotExist at /suggestions/untriaged/
CVEDerivationClusterProposal has no cached.
```

This seems to happen because suggestions are listed as soon as they're created, but the cached value is only created with some delay. So when someone accesses a list view with an uncached suggestion in it, they get a 500. This is a minor annoyance, but we just shouldn't have that sort of random outage.

The immediate fix is straightforward: Make the view list cached values only.
It gets a bit trickier when taking notifications into account. Those are created right after the suggestions, and notifications link to detail views on the suggestion. Probably the notification should only trigger when the cached value arrives, because only then it's safe to link to a view. We may even end up requiring the cached value to display the notification itself (e.g. to already show the subscribed packages affected).

Kishan0703:
Fixes #829

Suggestions are listed as soon as they're created, but the cached value
is only created with some delay. This causes a `RelatedObjectDoesNotExist`
crash when someone accesses the list view during that window.

Similarly, notifications are sent at the same time as the cache is built
(via separate pgpubsub channels), so users can receive notification links
to pages that aren't ready yet.

**Guard the views:**
- List view queryset now filters out suggestions without a cached value
- Detail view returns 404 instead of 500 for uncached suggestions

**Fix notification ordering:**
- Moved `create_package_subscription_notifications()` call to the end of
  `cache_new_suggestions()`, only on first cache creation
- Removed the separate `CVEDerivationClusterProposalNotificationChannel`
  listener and channel since it's no longer needed
- Removed the channel from the worker config in `configuration.nix`

Adds tests for both the view guards and the notification ordering.

### DIFF
--- a/src/webview/suggestions/views/detail.py
+++ b/src/webview/suggestions/views/detail.py
@@ -6,7 +6,9 @@ from django.urls import reverse
 from django.views.generic import DetailView, View
 
 from shared.auth import user_can_edit_suggestion
+from shared.listeners.cache_suggestions import cache_new_suggestions
 from shared.logs.fetchers import fetch_suggestion_events
+from shared.models.cached import CachedSuggestions
 from shared.models.issue import NixpkgsIssue
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
@@ -37,6 +39,8 @@ class SuggestionDetailView(DetailView):
 
     def get(self, request: HttpRequest, suggestion_id: int) -> HttpResponse:
         self.object = self.get_object()
+        if not CachedSuggestions.objects.filter(proposal=self.object).exists():
+            cache_new_suggestions(self.object)  # type: ignore
         if self.object.status == CVEDerivationClusterProposal.Status.PUBLISHED:
             issue = NixpkgsIssue.objects.get(suggestion=self.object)
             return redirect(
--- a/src/webview/tests/test_suggestions_cached_race.py
+++ b/src/webview/tests/test_suggestions_cached_race.py
@@ -1,0 +1,54 @@
+from django.urls import reverse
+from playwright.sync_api import Page, expect
+from pytest_django.live_server_helper import LiveServer
+
+from shared.listeners.cache_suggestions import cache_new_suggestions
+from shared.models.linkage import (
+    CVEDerivationClusterProposal,
+)
+
+
+def test_uncached_suggestion_not_visible(
+    live_server: LiveServer,
+    as_staff: Page,
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    as_staff.goto(live_server.url + reverse("webview:suggestion:untriaged_suggestions"))
+
+    # 500 will also not have the element visible
+    assert as_staff.response.status == 200
+    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).not_to_be_visible()
+
+    cache_new_suggestions(suggestion)
+
+    as_staff.goto(live_server.url + reverse("webview:suggestion:untriaged_suggestions"))
+    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).to_be_visible()
+
+
+def test_uncached_suggestion_detail_cached_on_demand(
+    live_server: LiveServer,
+    as_staff: Page,
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    as_staff.goto(
+        live_server.url
+        + reverse("webview:suggestion:detail", kwargs={"suggestion_id": suggestion.pk})
+    )
+
+    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).not_to_be_visible()
+
+
+def test_cached_suggestion_detail_returns_200(
+    live_server: LiveServer,
+    as_staff: Page,
+    cached_suggestion: CVEDerivationClusterProposal,
+) -> None:
+    as_staff.goto(
+        live_server.url
+        + reverse(
+            "webview:suggestion:detail",
+            kwargs={"suggestion_id": cached_suggestion.pk},
+        )
+    )
+
+    expect(as_staff.locator(f"#suggestion-{cached_suggestion.pk}")).to_be_visible()
--- a/src/webview/suggestions/views/lists.py
+++ b/src/webview/suggestions/views/lists.py
@@ -45,7 +45,7 @@ class SuggestionListView(ListView, ABC):
                 )
             except ValueError:
                 raise Http404
-        query_filters = Q()
+        query_filters = Q(cached__isnull=False)
         if self.status_filter is not None:
             query_filters &= Q(status=self.status_filter)
         if self.package_filter is not None:


src/webview/suggestions/views/detail.py
```diff
@@ -34,6 +35,8 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
 
     def get(self, request: HttpRequest, suggestion_id: int) -> HttpResponse:
         self.object = self.get_object()
+        if not CachedSuggestions.objects.filter(proposal=self.object).exists():
```
fricklerhandwerk, line 42: A caching delay in the detail view (especially if we optimize the data model further) could be small enough to allow on-demand caching instead

src/shared/listeners/cache_suggestions.py
```diff
@@ -219,6 +220,12 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
         logger.info(
             "CVE '%s' suggestion cached for the first time", suggestion.cve.cve_id
         )
+        try:
+            create_package_subscription_notifications(suggestion)
```
fricklerhandwerk: As noted in the comment on the related issue, we'd be dropping information without recovery options in the sad case here. IMO that's too risky, we want notifications to absolutely reach subscribers eventually. (In practice we can expect untriaged suggestions to be of less interest than accepted or published ones, but generally notifications are at the heart of this system.)

src/shared/tests/test_notification_timing.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        suggestion = make_suggestion()
+
+        # Precondition: no cached suggestion and no notifications yet
+        assert not CachedSuggestions.objects.filter(proposal=suggestion).exists()
+        assert Notification.objects.filter(user=user).count() == 0
+
+        # Act: cache the suggestion for the first time
+        cache_new_suggestions(suggestion)
```
fricklerhandwerk: Note for testing the sad case (see other review comment): mock that procedure with something that crashes.

src/webview/suggestions/views/detail.py
Kishan0703: You mentioned that the caching delay might be small enough to allow on-demand caching instead of returning a 404. I updated the detail view so that if the cache doesn’t exist, it now calls cache_new_suggestions() and builds it on the spot, allowing the page to load normally.

src/shared/listeners/cache_suggestions.py
Kishan0703: wrapping create_package_subscription_notifications() in a try/except could drop notifications without recovery. I removed the try/except so any failure propagates to the pgpubsub worker, allowing the operation to be retried and ensuring subscribers eventually receive notifications

src/shared/tests/test_notification_timing.py
Kishan0703: As suggested, I added a test that mocks create_package_subscription_notifications() to crash. The test verifies that the exception propagates and is not swallowed, while the cache is still created first.

src/shared/tests/test_notification_timing.py
```diff
@@ -0,0 +1,131 @@
...
+        make_drv: Callable[..., NixDerivation],
+    ) -> None:
+        """
+        If create_package_subscription_notifications() crashes, the exception
+        must propagate out of cache_new_suggestions() so the pgpubsub worker
+        can retry. Notifications must never be silently dropped.
+        """
+        suggestion = make_suggestion()
+
+        with patch(
```
fricklerhandwerk: This one may have been a misunderstanding. What happens if the *caching* fails? Right now we'd have no way to recover notifications.

src/shared/tests/test_notification_timing.py
```diff
@@ -0,0 +1,131 @@
...
+
+from shared.listeners.cache_suggestions import cache_new_suggestions
+from shared.models.cached import CachedSuggestions
+from shared.models.linkage import CVEDerivationClusterProposal, ProvenanceFlags
+from shared.models.nix_evaluation import NixDerivation, NixMaintainer
+from webview.models import SuggestionNotification as Notification
+
+
+@pytest.mark.django_db
+class TestCacheCreatesNotifications:
```
fricklerhandwerk: Style: We're using pytest-style tests everywhere else. Wrapping everything in a class makes it Django-style, and requires adding that annotation. Let's not do that. Adding any of the fixtures that access the db makes adding the `db` fixture or the `django_db` annotation unnecessary.

src/shared/tests/test_notification_timing.py
```diff
@@ -0,0 +1,131 @@
...
+    """Verify that cache_new_suggestions() triggers notifications on first cache creation."""
+
+    def test_notifications_created_on_first_cache(
+        self,
+        make_suggestion: Callable[..., CVEDerivationClusterProposal],
+        make_drv: Callable[..., NixDerivation],
+        make_maintainer: Callable[..., NixMaintainer],
```
fricklerhandwerk:
```suggestion
```

These seem to be unused. Similar in the other tests

src/shared/tests/test_notification_timing.py
```diff
@@ -0,0 +1,131 @@
...
+        make_user: Callable[..., User],
+        maintainer: NixMaintainer,
```
fricklerhandwerk: This one only incidentally happens to be the same as `suggestion.derivations.first().metadata.maintainer`. IMO it's cleaner to access it from the suggestion. (Pretty sure there's precedent in the code already. You'll need to assert `first()` is not `None` etc..., but it's still more evident we mean that very maintainer explicitly rather than accidentally.)

src/shared/listeners/cache_suggestions.py
```diff
@@ -214,13 +215,41 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
...
+    return created
 
-    if created:
-        logger.info(
-            "CVE '%s' suggestion cached for the first time", suggestion.cve.cve_id
-        )
-    else:
-        logger.info("CVE '%s' suggestion cache updated", suggestion.cve.cve_id)
+
+def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
```
fricklerhandwerk: Okay, seems we're still in misunderstanding mode and need more shared context.

This code doesn't make sense operationally. We can assume the caching logic is correct, it won't magically fail while the service runs. So there's no point to simply try twice.

A failure mode I expect is that we deploy a new version of the code while new entries are still being processed. A deployment will stop the service mid-way, and without somehow recording that work is in progress, there's no way to pick it up when the service starts up again. The reason I brought this up in the first place is that simply sequencing notifications after caching increases the amount of work per `Container` to a point where we may actually run into the situation that caching still runs into early working hours and then someone pushes a button.

We're currently not doing that recording, and figuring out how exactly it should work needs some thinking. This is an unhandled problem all over the code, since we're doing a bunch of things that simply must not fail, otherwise information is lost forever. Arguably most of the time it's not the biggest of deals (for either technical or practical reasons), but definitely we should not drop notifications.

Another few possibilities how to approach that, but I didn't put a lot of focus time into it yet:
- Actually record in the database when an item is picked up, and add recovery logic at `preStart` (as noted in a previous comment). For instance adopt [`django-tasks`](https://pypi.org/project/django-tasks/)
- Only create notifications when a suggestion is changed to "accepted" -- for anyone to press that button that there must be a cached value. This would also reduce noise for maintainers and make https://github.com/NixOS/nix-security-tracker/issues/836 less urgent.

src/webview/suggestions/views/detail.py
fricklerhandwerk: Oh but why did you remove the on-demand caching?

Kishan0703: I removed the on-demand caching from the detail view when splitting  the PR, because the on-demand caching approach in detail.py was  importing and calling cache_new_suggestions() from the listeners layer, 
which pulled in the same notification concerns we were trying to separate.

To keep this PR strictly focused on the view guards only and avoid  touching any listener/caching logic, I went with the simpler 404 approach.

If you'd prefer the on-demand caching approach for the detail view,  I'm happy to add it back, but wanted to confirm before doing so since  it would require importing cache_new_suggestions() into the view layer.

src/webview/tests/test_suggestions_cached_race.py
```diff
@@ -0,0 +1,62 @@
+import pytest
+from django.test import Client
+from django.urls import reverse
+
+from shared.models.linkage import CVEDerivationClusterProposal
+
+
+@pytest.fixture
+def staff_client(client: Client, user) -> Client:
```
fricklerhandwerk: Oh no, I meant actually using the playwright-based `as_staff` fixtures. Otherwise we'd be testing implementation details we really don't want to pin down in tests. (The response context is not observable for users and may change shape, but this should not require also changing tests.)

src/webview/suggestions/views/detail.py
fricklerhandwerk: Yes, let's actually bring back `cache_new_suggestions` even if it's dirty. https://github.com/NixOS/nix-security-tracker/pull/871 will cache in the detail view anyway, so we may as well test for it already.

src/webview/suggestions/views/detail.py
```diff
@@ -37,6 +38,9 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
...
+            from django.http import Http404
+            raise Http404
```
fricklerhandwerk:
```suggestion
            cache_new_suggestions(self.object)  # type: ignore
```

src/webview/tests/test_suggestions_cached_race.py
```diff
@@ -0,0 +1,38 @@
...
+    page = as_staff
+    page.goto(
+        live_server.url
+        + reverse(
+            "webview:suggestion:detail",
+            kwargs={"suggestion_id": cached_suggestion.pk},
+        )
+    )
+
+    assert page.locator(f"#suggestion-{cached_suggestion.pk}").count() == 1
```
fricklerhandwerk:
```suggestion
from django.urls import reverse
from playwright.sync_api import Page, expect
from pytest_django.live_server_helper import LiveServer

from shared.listeners.cache_suggestions import cache_new_suggestions
from shared.models.linkage import (
    CVEDerivationClusterProposal,
)


def test_uncached_suggestion_not_visible(
    live_server: LiveServer,
    as_staff: Page,
    suggestion: CVEDerivationClusterProposal,
) -> None:
    as_staff.goto(live_server.url + reverse("webview:suggestion:untriaged_suggestions"))

    # 500 will also not have the element visible
    assert as_staff.response.status == 200
    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).not_to_be_visible()

    cache_new_suggestions(suggestion)

    as_staff.goto(live_server.url + reverse("webview:suggestion:untriaged_suggestions"))
    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).to_be_visible()


def test_uncached_suggestion_detail_cached_on_demand(
    live_server: LiveServer,
    as_staff: Page,
    suggestion: CVEDerivationClusterProposal,
) -> None:
    as_staff.goto(
        live_server.url
        + reverse("webview:suggestion:detail", kwargs={"suggestion_id": suggestion.pk})
    )

    expect(as_staff.locator(f"#suggestion-{suggestion.pk}")).to_be_visible()
```

I can't push to your fork... please run `format`!

src/webview/suggestions/views/detail.py
```diff
@@ -5,7 +5,8 @@
 from django.urls import reverse
 from django.views.generic import DetailView, View
 
-from shared.auth import user_can_edit_suggestion
+from shared.auth import can_edit_suggestion, user_can_edit_suggestion
+from shared.models.cached import CachedSuggestions
 from shared.logs.fetchers import fetch_suggestion_events
 from shared.models.issue import NixpkgsIssue
```
fricklerhandwerk:
```suggestion
from shared.auth import user_can_edit_suggestion
from shared.listeners.cache_suggestions import cache_new_suggestions
from shared.logs.fetchers import fetch_suggestion_events
from shared.models.cached import CachedSuggestions
```

PR #857, opened 2026-03-07
Ingest CVSS 4.0 metrics 

### LINKED ISSUE #839: Ingest CVSS 4.0 metrics
We're currently only ingesting CVSS 3.1 data, but most new CVEs don't even ship that any more. This results in metrics not being shown in the overivew.

https://github.com/NixOS/nix-security-tracker/blob/8f4f5d745db7905b9a71de638595a62f9c8da35e/src/shared/fetchers.py#L102

DarshanCode2005:
Resolves #839 

Summary:

Implemented CVSS ingestion and display compatibility for newer CVE records that no longer include cvssV3_1.
- In fetchers.py metric parsing now falls back across versions in order: cvssV4_0 -> cvssV3_1 -> cvssV3_0, instead of only reading cvssV3_1.
- In the same path, base score parsing is now guarded (no crash on missing/invalid values), and base_severity is populated from raw CVSS data.
- In viewutils.py, severity badge generation no longer assumes every vector is CVSS3; CVSS4 vectors are handled safely without breaking overview rendering.
- Added tests in test_fetchers.py and test_viewutils.py to cover CVSS fallback ingestion and badge behavior for both CVSS4 and CVSS3 vectors.

- [x] Manually verified
- [x] New Tests created

Screenshot from 2026-03-08 04-03-59


src/webview/templatetags/viewutils.py
```diff
@@ -127,18 +127,36 @@ def severity_badge(metrics: list[dict]) -> dict:
...
-                "vectorString": m["raw_cvss_json"]["vectorString"],
-                "version": m["raw_cvss_json"]["version"],
-                "metrics": {
-                    # XXX(@fricklerhandwerk): Yes, the *value* description is also indexed by *key*!
-                    f"{METRICS_ABBREVIATIONS[k]} ({k})": f"{parsed.get_value_description(k)} ({v})"
-                    for k, v in parsed.metrics.items()
-                    if not k.startswith("M")  # Don't display modified metrics
-                },
+        raw_cvss = m.get("raw_cvss_json", {})
+        if "vectorString" in raw_cvss:
```
fricklerhandwerk: Why not select the right CVSS parser based on `m["format"]`?

src/shared/fetchers.py
```diff
@@ -97,15 +97,31 @@ def to_camel_case(name: str) -> str:
...
-    raw_cvss = data.get("cvssV3_1", {})
+    ctx["format"] = selected_format
     ctx["raw_cvss_json"] = raw_cvss
 
     if raw_cvss:
         ctx["scope"] = raw_cvss.get("scope")
         ctx["vector_string"] = raw_cvss.get("vectorString")
-        ctx["base_score"] = float(raw_cvss.get("baseScore"))
+        base_score = raw_cvss.get("baseScore")
+        if base_score is not None:
```
fricklerhandwerk: Why are we checking for None here? The spec doesn't allow that anyway. Since we're consuming data from the authoritative source, we can reasonably expect it to follow the spec (and so far it did).

src/shared/fetchers.py
```diff
@@ -97,15 +97,31 @@ def to_camel_case(name: str) -> str:
 
 
 def make_metric(data: dict[str, Any]) -> models.Metric:
+    cvss_preference = ("cvssV4_0", "cvssV3_1", "cvssV3_0")
+
+    selected_format = "cvssV3_1"
```
fricklerhandwerk: If the first item in the preference list is v4, why do we default to 3.1? 

We could use the opportunity to retain more information here. Before any change, we're storing `cvssV3_1` unconditionally, even if there's no data there, so there's no way to identify a gap after the fact.

For example, how about something along these lines:

```python
ctx["format"] = data.get("format", "")
if ctx["format"] in supported_cvss_versions:
  raw_cvss = data.get(ctx["format"])
elif not ctx["format"]:
  # loop over `supported_cvss_versions`
else:
  raw_cvss = data.get("other", {"content": {}})["content"]
```

src/shared/fetchers.py
```diff
@@ -97,15 +97,31 @@ def to_camel_case(name: str) -> str:
...
+    raw_cvss: dict[str, Any] = {}
+    for cvss_format in cvss_preference:
+        candidate = data.get(cvss_format)
+        if isinstance(candidate, dict) and candidate:
```
fricklerhandwerk: Isn't this always a dict by spec?

src/webview/tests/test_viewutils.py
```diff
@@ -0,0 +1,40 @@
+from webview.templatetags.viewutils import severity_badge
+
+
+def test_severity_badge_handles_cvss_v4_vectors() -> None:
```
fricklerhandwerk: It's surely good to exercise the code paths that produce user-visible results, but I'm not fully convinced we need to test what amounts to calling into the `cvss` library.

What may be worth testing though is that the same (cross-version compatible) `vectorString`, `baseScore`, and `baseSeverity` will produce the same output dict. The other fields are rather cosmetic (only for better human-readability) and computed by the library anyway.

In any case, if you're re-using input data, how about putting them into `conftest.py` as fixtures?

fricklerhandwerk (dismissed):
Thanks a lot for the stab at it. I think some things can be simplified, and some of your decisions are not entirely obvious to me. Would appreciate clarification here (if I'm just missing the point) or as in-code comments (if you're doing something clever).

src/shared/fetchers.py
DarshanCode2005: My intention was to keep the existing default behavior while adding v4 support, but I agree that preserving the actual format is better and avoids masking missing data. I'll update it to store the format explicitly and select the metric based on the provided format as suggested by u.

src/shared/fetchers.py
DarshanCode2005: That was mostly a defensive check out of habit. Since the data follows the spec, it makes sense to rely on it here. I'll remove the check and update the patch.

src/webview/templatetags/viewutils.py
DarshanCode2005: Yes, makes sense. I have stored it earlier. I guess I should follow this if m["format"] == "cvssV4_0",

src/webview/tests/test_viewutils.py
DarshanCode2005: Thanks for the feedback! I agree that testing the internal logic of the cvss library is redundant in this context. I'll refactor the tests to focus on verifying that the common fields stated by you are correctly extracted across different CVSS versions. I'll also move the test data into conftest.py as fixtures as suggested.

src/webview/templatetags/viewutils.py
```diff
@@ -127,18 +127,35 @@ def severity_badge(metrics: list[dict]) -> dict:
...
-            return {
-                "vectorString": m["raw_cvss_json"]["vectorString"],
-                "version": m["raw_cvss_json"]["version"],
-                "metrics": {
-                    # XXX(@fricklerhandwerk): Yes, the *value* description is also indexed by *key*!
+        raw_cvss = m.get("raw_cvss_json", {})
+        fmt = m.get("format", "")
+        if fmt.startswith(("cvssV3", "cvssV4")):
+            result = {
+                "vectorString": raw_cvss.get("vectorString"),
```
fricklerhandwerk: But why aren't you using `CVSS4()` here?

DarshanCode2005: Sir I looked into the cvss (https://github.com/RedHatProductSecurity/cvss/blob/master/cvss/cvss4.py) just to confirm that CVSS4 exposes the same interface like as that of CVSS3, namely metrics etc...
For exmaple in cvss4.py the parser populates the self.metrics during init:
```
class CVSS4(object):
    """
    Class to hold CVSS4 vector, parsed values, and all scores.
    """

    def __init__(self, vector):
        """
        Args:
            vector (str): string specifying CVSS4 vector, fields may be out of order, fields which
                          are not mandatory may be missing
        """
        self.vector = vector
        self.metrics = {}
        self.missing_metrics = []

        self.base_score = None
        self.severity = None

        self.parse_vector()
        self.check_mandatory()
        self.add_missing_optional()
        self.compute_base_score()
        self.compute_severity()
 ```
 This is similar to that interface used by CVSS3, I am thinking of updating the implementation to select the parser dynamically based on the metric format and use the corresponding abbr. tabels (constants3 and constants4) [ref:(https://github.com/RedHatProductSecurity/cvss/blob/master/cvss/constants3.py) and (https://github.com/RedHatProductSecurity/cvss/blob/master/cvss/constants4.py)]
 
 So I am roughly thinking now to do iit this way:
```
# I will import METRICS_ABBREVIATION as ...3 and ...4

if fmt.startswith("cvssV3"):
          parser_class = CVSS3
          abbrevs = METRICS_ABBREVIATIONS3
elif fmt.startswith("cvssV4"):
          parser_class = CVSS4
          abbrevs = METRICS_ABBREVIATIONS4
....
``` 
somthing like this. So does this approach looks correct?

fricklerhandwerk: Yes, exactly. Maybe for even more compact presentation one could have a little dict that maps from supported format to the right library items.

fricklerhandwerk:
This turned out to be really not that simple, because it required a bunch of non-trivial decisions. Also it doesn't look like they could have been pulled apart into smaller changes. But at least we have the recaching trigger, which allows deploying such a change without hassle.

fricklerhandwerk:
Briefly looked at it together with @adekoder and @florentc

PR #861, opened 2026-03-08
perf: batch activity log queries in list views

### LINKED ISSUE #847: Reduce activity log database queries in list views
## Problem

Loading a suggestion list page calls `fetch_suggestion_events()` once per suggestion, firing 3 database queries per suggestion (status events, package events, maintainer events). For a page of N suggestions this results in 3N queries just for activity logs.

This was noted as known remaining overhead in f87f563:
> "The remaining overhead is from inefficient querying of the activity log. But this need to be thought through properly. We may not need the edit types and could possibly rely on pghistory exclusively."

### Option 1 Batch queries (immediate fix)
Add `fetch_suggestion_events_batch()` that fetches all events for all suggestions on a page in 3 total queries using `filter(suggestion_id__in=suggestion_ids)`. Reduces 3N → 3 queries with no schema changes.

Tradeoff: doesn't address the underlying structural question, just queries the existing tables more efficiently.

### Option 2 Rely on pghistory directly
The custom event tables (`MaintainersEditEvent`, `PackageEditEvent`, `CVEDerivationClusterProposalStatusEvent`) duplicate what pghistory already tracks. Querying pghistory's unified event stream directly would reduce this to a single query and remove redundant code.

Tradeoff: pghistory's generic event format would need to be mapped back into the typed event model the activity log currently uses.

### Option 3 Denormalize into the suggestion model
Store a precomputed activity log snapshot directly on `CVEDerivationClusterProposal`, updated on each write. Zero extra queries on page load.

Tradeoff: every write (maintainer add, package edit, status change) must also update the precomputed field, adding write complexity and cache invalidation concerns.

## Current status

Option 1 is implemented and tested in branch `perf/batch-activity-log-queries`. Happy to open a PR, but wanted to raise the deeper question first since fricklerhandwerk flagged that this needs proper thought.

All three options are viable, but Option 1 is the seemed like the right starting point. Option 2 requires understanding exactly what pghistory tracks vs what the custom event tables add removing those tables without that clarity risks losing data or breaking the activity log format. Option 3 introduces write-side complexity that needs careful design to avoid stale data. Option 1 delivers the query reduction immediately with zero schema changes and zero risk, while leaving the door open for the deeper restructuring Options 2 and 3 represent. It also gives us a working baseline to benchmark against if we pursue those later.

Bbn08:
Closes #847

Reduces activity log queries in suggestion list views from 3N to 3 per page load by introducing `fetch_suggestion_events_batch()`.

Previously `SuggestionContext` called `fetch_suggestion_events()` for each suggestion individually, issuing 3 queries per suggestion. The new batch fetcher filters all three event types by `suggestion_ids` in one go and wires results back per suggestion in Python.

Applies to both the suggestion list views and the issue list view.

### DIFF
--- a/src/webview/suggestions/views/detail.py
+++ b/src/webview/suggestions/views/detail.py
@@ -23,13 +23,13 @@ class SuggestionDetailView(DetailView):
     def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
         can_edit = can_edit_suggestion(self.request.user)
-        events = fetch_suggestion_events([self.object.pk])  # type: ignore
+        events = fetch_suggestion_events([self.object.pk])
         context.update(
             {
                 "suggestion_context": get_suggestion_context(
                     self.object,  # type: ignore
                     can_edit=can_edit,
-                    pre_fetched_events=events[self.object.pk],  # type: ignore
+                    pre_fetched_events=events[self.object.pk],
                 )
             }
         )


src/webview/suggestions/views/base.py
```diff
@@ -22,11 +23,13 @@ def get_suggestion_context(
     suggestion: CVEDerivationClusterProposal,
     can_edit: bool,
     is_compact: bool = False,
+    pre_fetched_events: list[RawEventType] | None = None,
```
fricklerhandwerk: Is there any place left where we're not passing this? If no, there shouldn't be a default value and no conditional fetching.

Bbn08: There are still a few call sites not passing it, so I'll update all of them and make the parameter required. Working on it now.

Bbn08:
Removed the singleton and promoted the batch function as the sole interface. Also ran `nix-shell --run format` to fix the linter errors.

src/webview/suggestions/views/detail.py
```diff
@@ -22,11 +23,13 @@ class SuggestionDetailView(DetailView):
     def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
         can_edit = can_edit_suggestion(self.request.user)
+        events = fetch_suggestion_events([self.object.pk])  # type: ignore
```
fricklerhandwerk:
```suggestion
        events = fetch_suggestion_events([self.object.pk])
```

src/webview/suggestions/views/detail.py
```diff
@@ -22,11 +23,13 @@ class SuggestionDetailView(DetailView):
...
         context.update(
             {
                 "suggestion_context": get_suggestion_context(
                     self.object,  # type: ignore
                     can_edit=can_edit,
+                    pre_fetched_events=events[self.object.pk],  # type: ignore
```
fricklerhandwerk:
```suggestion
                    pre_fetched_events=events[self.object.pk],
```

PR #863, opened 2026-03-09
Fix: Show CVSS base score in GitHub issue

### LINKED ISSUE #859: Show CVSS base score in GitHub issue
As a Nixpkgs user, I want to see the overall severity of a security issue on GitHub.

Right now we're only rendering the vector string and an extended view inside the details.

## Acceptance Criteria

Given an accepted suggestion,
When I publish a GitHub issue,
Then the severity score is displayed (and ideally, appropriately highlighted) in the issue body.

gabrnavarro:
## What changed

The CVSS base score and severity label are now shown in published GitHub issues and in the web UI severity badge.

**GitHub issue body**: The `<details>` summary line now reads e.g. `CVSS 7.5 HIGH | CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:N/A:N`, making the severity immediately visible without expanding the details block.

**Web UI severity badge fix**: The `severity_badge()` inclusion tag was returning a flat dict (e.g. `{"vectorString": ..., "version": ...}`), but the template `severity_badge.html` expected a `metric` context variable (`{% if metric.vectorString %}`). This meant the CVSS badge never rendered in the web UI. Fixed by wrapping the return value as `{"metric": {...}}` and passing through all `raw_cvss_json` fields so the template can access `metric.attackVector`, `metric.scope`, etc. directly.

## Why

Fixes #859

Users couldn't see the overall severity score of a CVE at a glance — either on GitHub or in the web UI. The base score is the most actionable piece of CVSS information.

## Testing

- `test_cvss_base_score_in_issue_body`: creates a suggestion with a CVSS metric (`baseScore: 7.5`, `baseSeverity: HIGH`), publishes via `create_gh_issue`, and asserts both values appear in the issue body.
- `test_cvss_base_score_visible_in_web_ui`: Playwright test that loads the accepted suggestions page with CVSS data and asserts the base score `7.5` is visible without expanding any details.

### DIFF
--- a/src/shared/github.py
+++ b/src/shared/github.py
@@ -60,7 +60,11 @@ def create_gh_issue(
             metrics = "\n".join([f"- {k}: {v}" for k, v in metric["metrics"].items()])
             score = metric.get("baseScore")
             severity = metric.get("baseSeverity", "")
-            score_label = f"<strong>{score:.1f} {severity}</strong> | " if score is not None else ""
+            score_label = (
+                f"<strong>{score:.1f} {severity}</strong> | "
+                if score is not None
+                else ""
+            )
             return f"""
 <details>
 <summary>{score_label}{metric["vectorString"]}</summary>
--- a/src/webview/tests/test_issues.py
+++ b/src/webview/tests/test_issues.py
@@ -14,7 +14,6 @@ from shared.github import create_gh_issue
 from shared.listeners.cache_suggestions import cache_new_suggestions
 from shared.models.cve import (
     Container,
-    Metric,
 )
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
@@ -207,7 +206,6 @@ def test_cvss_base_score_visible_in_web_ui(
     as_staff.goto(live_server.url + reverse("webview:suggestion:accepted_suggestions"))
     suggestion = as_staff.locator(f"#suggestion-{accepted_suggestion.cached.pk}")
 
-
     # The base score should be visible without expanding the CVSS details
     expect(suggestion.get_by_text("7.5 HIGH")).to_be_visible()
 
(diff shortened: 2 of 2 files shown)


src/webview/tests/test_issues.py
```diff
@@ -177,3 +178,48 @@ def mock_get_maintainer_username(
...
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+    mocker: MockerFixture,
+) -> None:
+    """Test that the CVSS base score is displayed in the GitHub issue body."""
+    # Create a container with a proper CVSS metric that includes a base score
+    container = make_container()
+    # Update the metric to have a real CVSS vector with base score
+    metric = container.metrics.first()
+    assert metric is not None
+    metric.raw_cvss_json = {
```
fricklerhandwerk, line 194: There's a case where that JSON is empty (see issue about investing CVSSv4), so a test for that will reveal failure modes in this change that I suspect are unhandled. Please exercise that code path as well :)

src/shared/github.py
```diff
@@ -57,9 +57,12 @@ def cvss_details() -> str:
         metric = severity_badge(cached_suggestion.payload["metrics"])
         if metric:
             metrics = "\n".join([f"- {k}: {v}" for k, v in metric["metrics"].items()])
+            score = metric.get("baseScore")
+            severity = metric.get("baseSeverity", "")
+            score_label = f"{score:.1f} {severity} | " if score is not None else ""
```
fricklerhandwerk: Do you think we can make it bold and maybe colored according to the severity (there already exists a snippet that handles relevant cases in a template)? I don't know off the top of my head what GitHub markup supports here. A screenshot with the result would be great!

DarshanCode2005: Suggestion: what if according to the severity, we can show different colored dots like 🔴 (critical severity), 🟠 (high) , 🟡 (medium), 🟢(low) ... I guess could convey more information in a single go...

DarshanCode2005:
image

Run `nix-shell --run format` , I guess the red-build is because of that.

PR #865, opened 2026-03-09
fix: unsafe social account lookup in Github webhook

### LINKED ISSUE #858: Webhook team sync can grant wrong user admin/committer groups (missing provider filter) and crash on unknown users
**Description**
`src/shared/auth/github_state.py` has a vulnerable lookup in `sync_team_membership_from_webhook()`:

```py
user = User.objects.get(socialaccount__uid=github_user_id)
```

This lookup is unsafe because it does not filter `socialaccount__provider="github"` and does not handle missing users.

**Impact**
- Privilege misassignment: a non-GitHub social account with the same `uid` can receive GitHub-derived group permissions (`security` / `committers`).
- Production stability: webhook handling can return 500 when the GitHub user is not registered locally (`User.DoesNotExist`).

**Repro**
1. Create a local user with `SocialAccount(provider="google", uid="99999")`.
2. Send a GitHub team membership webhook for GitHub member id `99999`.
3. The Google-linked user may be matched and added to privileged Django groups.

**Expected**
- Only users with `SocialAccount(provider="github", uid=<github_user_id>)` are eligible.
- Unknown GitHub users should be ignored (no crash, no group change).

**Suggested fix**
```py
try:
    user = User.objects.get(
        socialaccount__provider="github",
        socialaccount__uid=github_user_id,
    )
except User.DoesNotExist:
    return
```

Optionally also catch `User.MultipleObjectsReturned` and log+return defensively.

DarshanCode2005:
Resolves #858 

Summary is already provided in the issue.

Implementation was approved by maintainer @fricklerhandwerk

### DIFF
--- a/src/shared/tests/test_github_sync.py
+++ b/src/shared/tests/test_github_sync.py
@@ -361,13 +361,11 @@ class GithubSyncTests(TestCase):
         apps.get_app_config("shared").github_state = gh_state  # type: ignore
 
         # Create a malicious Google account using the same UID as an intended victim
-        victim_github_uid = 88888 
-        
+        victim_github_uid = 88888
+
         malicious_user = User.objects.create_user(username="malicious-google-user")
         SocialAccount.objects.create(
-            user=malicious_user,
-            provider="google",
-            uid=str(victim_github_uid)
+            user=malicious_user, provider="google", uid=str(victim_github_uid)
         )
 
         payload = {


src/shared/auth/github_state.py
```diff
@@ -99,7 +99,15 @@ def sync_team_membership_from_webhook(
         Update group membership from the payload received via GitHub webhook.
         """
 
-        user: User = User.objects.get(socialaccount__uid=github_user_id)
+        try:
+            user: User = User.objects.get(
+                socialaccount__provider="github", socialaccount__uid=github_user_id
```
fricklerhandwerk: Please do add a test. IIRC `socialaccount__uid` is a string; in Python there's no way to know this will work without exercising the code path.

src/shared/auth/github_state.py
```diff
@@ -99,7 +99,15 @@ def sync_team_membership_from_webhook(
...
+            )
+        except (User.DoesNotExist, User.MultipleObjectsReturned):
```
fricklerhandwerk: These are two different error conditions and should therefore be handled in two separate cases. It's even kind of fine -- possibly even *better* -- to not catch anything here. We get notifications on 500s and I'd investigate what's going on. If this ever happens there will be something funky with that particular user.

Issue #836, opened 2026-02-28
Remove notifications for packages that are ignored


fricklerhandwerk:
As a user of the security tracker, I don't want to get notified about suggestions where my package was removed in triaging.

## Acceptance criteria

Given I'm subscribed to a set of packages,
When a subset of those is linked in a suggestion and another subset of them later ignored,
Then the second subset does not show up in the corresponding notification.

Given I'm subscribed to a set of packages,
When a subset of those is linked in a suggestion and all of it later ignored,
Then the notification with that suggestion is removed entirely.

Depends on:
- https://github.com/NixOS/nix-security-tracker/issues/835

PR #867, opened 2026-03-11
Show packages matched for a user (subscribes or maintains) in notifications

florentc:
This shows packages matching what a user subscribes to in a section, and what a user maintains in another.

## Current implementation

Since notifications cannot be automatically deleted in response to some events yet (for instance when packages are ignored as described in #836), the current implementation has 2 unavoidable quirks:
                                                                      
1. The user can receive a notification because they subscribed to a package. If they then unsubscribe, the notification will still be here but the matching package will no longer be displayed as such in the notification.
                                                                      
2. If at some point the package is ignored within the suggestion, then it won't appear in the list of matched packages in the notification. Once again, we may end up with notification that seemingly match nothing of interest. Another implementation choice is to check "original_packages" instead of "packages", which is the list of packages auto-matched initially, before anybody might have ignored some packages.

 
## Minor details

I removed "Affected packages" from the `suggestion_stub` component. Instead, I put this section directly in the notification template, along with the sections this PR introduces. Either I was to do this, or I would have had to put the new sections in stub too out of consistency and this `suggestion_stub` component was starting to no longer be a stub but something closer to the actual suggestion component (or its compact variant). It would have been confusing to see both suggestion and suggestion_stub coexist, being similar yet different.

So I decided to make stub only the very basic one link to the suggestion, and have the other sections in the notification itself. Besides, the new sections (matched subscribed/maintained packages) is notification specific and not suggestion-specific so it made sense IMO to present it that way.

## Tests

My local test environment is still broken so I'll add them as soon as I sort this out.

### DIFF
--- a/src/shared/tests/conftest.py
+++ b/src/shared/tests/conftest.py
@@ -372,22 +372,3 @@ def make_package_notification(
         return create_package_subscription_notifications(suggestion)
 
     return wrapped
-
-
-@pytest.fixture
-def make_package_subscription(
-    user: User,
-    drv: NixDerivation,
-) -> Callable[..., User]:
-    def wrapped(
-        user: User = user,
-        package_name: str = drv.attribute,
-    ) -> User:
-        profile = user.profile
-        if package_name not in profile.package_subscriptions:
-            profile.package_subscriptions.append(package_name)
-            profile.save(update_fields=["package_subscriptions"])
-
-        return user
-
-    return wrapped
--- a/src/webview/models.py
+++ b/src/webview/models.py
@@ -73,7 +73,7 @@ class SuggestionNotification(Notification):
         # We should then also show (maybe in the equivalent of the text message) what state that match is now in (e.g. ingored, ideally for which reason).
         # Maybe even garbage-collect the notification if it got obsolete and wasn't yet served or otherwise exposed to the user.
         # FIXME(@fricklerhandwerk): User-facing text should be generated from structured data in templates.
-        return f"{self.suggestion.cve.cve_id} was automatically matched to packages of interest for you"
+        return f"{self.suggestion.cve.cve_id} was automatically matched to packages you subscribed to"
 
 
 class Profile(models.Model):
@@ -131,6 +131,18 @@ class Profile(models.Model):
             read.delete()
 
         return count
+
+    def subscribe_to_package(self, package: str) -> None:
+        """Add a package to the subscribed packages."""
+        if package not in self.package_subscriptions:
+            self.package_subscriptions.append(package)
+            self.package_subscriptions.sort()
+            self.save(update_fields=["package_subscriptions"])
+
+    def unsubscribe_from_package(self, package: str) -> None:
+        """Remove a package from the subscribed packages."""
+        self.package_subscriptions.remove(package)
+        self.save(update_fields=["package_subscriptions"])
 
 
 @receiver(post_save, sender=User)
--- a/src/webview/notifications/context.py
+++ b/src/webview/notifications/context.py
@@ -77,8 +77,8 @@ class NotificationContext:
         # Filter packages where the user is a maintainer
         # NOTE(@florentc): We populate this list regardless of whether the user
         # has enabled "notify about packages I maintain". In case the user is
-        # notified about another package, it is still relevant info to show
-        # them they also maintain some related packages.
+        # notified about a package they otherwise subscribe to, it is still
+        # relevant to show them they also maintain some related packages.
         matching_packages = {
             attr: pdata
             for attr, pdata in suggestion_packages.items()
--- a/src/webview/notifications/views.py
+++ b/src/webview/notifications/views.py
@@ -65,7 +65,6 @@ class ToggleNotificationReadView(LoginRequiredMixin, TemplateView):
 
         if request.headers.get("HX-Request"):
             return self.render_to_response(
-                # {"notification": notification, "new_unread_count": new_unread_count}
                 {
                     "data": NotificationContext(
                         notification=notification,
--- a/src/webview/templates/notifications/components/notification.html
+++ b/src/webview/templates/notifications/components/notification.html
@@ -36,11 +36,14 @@
       </div>
     {% endif %}
     {% if not data.matching_maintained_packages and not data.matching_subscribed_packages %}
-    <div class="rounded-box {% if not data.notification.is_read %}hl-red{% endif %} prose">
-        <h2 class="heading">This notification may be obsolete</h2>
-        <p>It is about a suggestion that no longer matches packages you are subscribed to or maintain.</p>
-        <p>The suggestion was initially matched to packages you either follow or maintain but at some point somebody may have set these packages as irrelevant, or you unsubscribed. You may remove this notification directly or refer to the "Ignored packages" section of <a href="{% url 'webview:suggestion:detail' suggestion_id=data.notification.suggestion.pk %}">the suggestion</a> if you want to double check.</p>
-      </div>
+    <details class="rounded-box {% if not data.notification.is_read %}hl-red{% endif %} prose">
+        <summary class="bold">This notification is obsolete and can be deleted</summary>
+        <p>
+          The suggestion was initially matched to packages you either follow or maintain.
+          These packages were later set as irrelevant, or you unsubscribed.
+        </p>
+        <p>You may remove this notification directly or double check <a href="{% url 'webview:suggestion:detail' suggestion_id=data.notification.suggestion.pk %}#suggestion-{{ data.notification.suggestion.pk }}-ignored-packages">ignored packages of that suggestion</a>.</p>
+      </details>
     {% endif %}
   {% endif %}
   <form method="post" action="{% url 'webview:notifications:toggle_read' data.notification.id %}" class="row-reverse" hx-indicator="find button">
--- a/src/webview/subscriptions/views.py
+++ b/src/webview/subscriptions/views.py
@@ -63,9 +63,7 @@ class AddSubscriptionView(LoginRequiredMixin, TemplateView):
             )
 
         # Add subscription
-        profile.package_subscriptions.append(package_name)
-        profile.package_subscriptions.sort()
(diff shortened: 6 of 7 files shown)


fricklerhandwerk:
As a hack to squeeze the change in sooner, could we compute the notification list taking those cached `packages` into account? And as a very ugly workaround, recalculate the unread count when getting the list view... Otherwise we'd have to finish the ignored/additional set semantics first.

florentc:
> As a hack to squeeze the change in sooner, could we compute the notification list taking those cached packages into account? 

Not certain to understand how far you would like to go.
What I think I understood:
Instead of generating notifications once during ingestion, you want the list of notifications to be computed when the user browses to the notification center. This list would be obtained by querying the untriaged/accepted suggestions whose packages include a subscribed or maintained package.

If this is what you meant, to me it's a major change of the notifications paradigm that cannot be shoehorned in a PR like this one nor even started before we write very precise user stories covering the edge cases. In particular how to recalcultate the unread count: we can't do that for every request so we'd have to decide how to proceed, maybe a hook every time something modifies any suggestion.

> And as a very ugly workaround, recalculate the unread count when getting the list view

I don't think this would be useful. The unread count is useful when we are NOT currently browsing the notification center, that's when it's crucial that it's right, updating it last moment, when consulting the list of notifications, seems a bit too late. Besides, it would be weird as a user that the counter changes when you browse to the notifications without doing anything else. If I saw that as a user, I'd think it's a bug.

> Otherwise we'd have to finish the ignored/additional set semantics first.

You are talking about maintainers right? There is no "additional" for packages. For now, for the maintainers I don't even use those active/ignored/additional lists, but the data from the active package because the purpose of #836 is to display which packages are concerned.

## Opinion

I'd be in favor of going through with this in the current state. It does fit the acceptance criteria of #836. It is a user friendly step forward. It brings useful info. It doesn't introduce any regression from before. In fact it makes the situation better as now notifications without any "packages you subscribed" or "packages you maintain" will appear more clearly as outdated.

I'm thinking of a compromise for the time between now and when we come up with proper automatic notification cleanup: when a notification has no "packages you subscribed to" or "packages you maintain", we add a small info on top saying "This notification might have become obsolete", with a "Delete" button attached.

fricklerhandwerk:
> Instead of generating notifications once during ingestion

No, I meant in order to fix the current implementation's problem of ignored packages leaving empty notifications, filter them in the list view based on that cached list of active packages. That may interact with the notification count, but the hack would at worst overestimate the number until the list is accessed.

We could even delete obsolete ones when generating the list. It's ugly internally but can be fixed later and doesn't require rushing the data model change to have the feature implemented cleanly.

src/shared/tests/conftest.py
```diff
@@ -359,14 +359,35 @@ def wrapped(
...
+    drv: NixDerivation,
+) -> Callable[..., User]:
+    def wrapped(
+        user: User = user,
+        package_name: str = drv.attribute,
+    ) -> User:
+        profile = user.profile
+        if package_name not in profile.package_subscriptions:
+            profile.package_subscriptions.append(package_name)
+            profile.save(update_fields=["package_subscriptions"])
```
fricklerhandwerk: Hm, how about we make this a method on `user.profile`? Because really this is not a fixture but a state manipulation on the user.

src/webview/notifications/context.py
```diff
@@ -0,0 +1,91 @@
...
+        # seemingly match nothing of interest. Another implementation choice is
+        # to check "original_packages" instead of "packages", which is the list
+        # of packages auto-matched initially, before anybody might have ignored
+        # some packages.
+        suggestion_packages = self.notification.suggestion.cached.payload.get(
+            "packages", {}
+        )
+
+        # Filter by user's manual subscriptions
+        subscribed_attrs = set(user_profile.package_subscriptions)
```
fricklerhandwerk, line 60:
```suggestion
        subscribed_packages= set(user_profile.package_subscriptions)
```

Let's preemptively assume we'll work with packages as a first-class concept. Once that model is in, `attr` will be confusing.

src/webview/notifications/context.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # FIXME(@florentc): Not all users in db seem to have a github_id so
+        # matching is done by username here instead. It turns out it is also
+        # how matching is done to generate the notification in the first place.
+        username = user_profile.user.username
+
+        # Filter packages where the user is a maintainer
+        # NOTE(@florentc): We populate this list regardless of whether the user
+        # has enabled "notify about packages I maintain". In case the user is
+        # notified about another package, it is still relevant info to show
```
fricklerhandwerk:
```suggestion
        # has enabled "notify about packages I maintain". In case the user is
        # notified about a package they otherwise subscribe to, it is still relevant info to show
```
I had a hard time parsing that, but yes, the logic makes sense.

Side note on formatting: I may be a weirdo due to mental scars from past years, but I consider in-code comments documentation since they're prose. It does help having them one-sentence-per-line, but I won't insist on it.

src/webview/notifications/views.py
```diff
@@ -48,10 +61,18 @@ def post(self, request: HttpRequest, notification_id: int) -> HttpResponse:
             user=request.user,
         )
         new_unread_count = notification.toggle_read()
+        user_profile = self.request.user.profile
 
         if request.headers.get("HX-Request"):
             return self.render_to_response(
-                {"notification": notification, "new_unread_count": new_unread_count}
+                # {"notification": notification, "new_unread_count": new_unread_count}
```
fricklerhandwerk: Artefact?


```suggestion
```

src/webview/templates/notifications/components/notification.html
```diff
@@ -1,30 +1,61 @@
 {% load viewutils %}
-{% if new_unread_count is not None %}
-{% notifications_badge count=new_unread_count oob_update=True %}
+{% if data.new_unread_count is not None %}
```
fricklerhandwerk, line 2: I'm not convinced there's a point in nesting the `NotificationContext` under `data`. We already have everything we need in the type, no?

src/webview/templates/notifications/components/notification.html
```diff
@@ -1,30 +1,61 @@
...
+          {% nixpkgs_package attribute pdata%}
+          {% endfor %}
+        </div>
+      </div>
+    {% endif %}
+    {% if not data.matching_maintained_packages and not data.matching_subscribed_packages %}
+    <div class="rounded-box {% if not data.notification.is_read %}hl-red{% endif %} prose">
+        <h2 class="heading">This notification may be obsolete</h2>
+        <p>It is about a suggestion that no longer matches packages you are subscribed to or maintain.</p>
+        <p>The suggestion was initially matched to packages you either follow or maintain but at some point somebody may have set these packages as irrelevant, or you unsubscribed. You may remove this notification directly or refer to the "Ignored packages" section of <a href="{% url 'webview:suggestion:detail' suggestion_id=data.notification.suggestion.pk %}">the suggestion</a> if you want to double check.</p>
```
fricklerhandwerk: Didn't test what it looks like, but how about collapsing the hint by default? Also slightly more compact wording.

```suggestion
        <details><summary>This notification is obsolete and can be deleted</summary>
        <p>
          The suggestion was initially matched to packages you either follow or maintain.
          These packages were later set as irrelevant, or you unsubscribed.
          You may remove this notification directly or double check <a href="{% url 'webview:suggestion:detail' suggestion_id=data.notification.suggestion.pk %}#suggestion-{{ data.notification.suggestion.pk }}-ignored-packages">ignored packages on that suggestion</a>.
          </p>
         </details>
```

src/webview/models.py
```diff
@@ -73,7 +73,7 @@ def title(self) -> str:
         # We should then also show (maybe in the equivalent of the text message) what state that match is now in (e.g. ingored, ideally for which reason).
         # Maybe even garbage-collect the notification if it got obsolete and wasn't yet served or otherwise exposed to the user.
         # FIXME(@fricklerhandwerk): User-facing text should be generated from structured data in templates.
-        return f"{self.suggestion.cve.cve_id} was automatically matched to packages you subscribed to"
+        return f"{self.suggestion.cve.cve_id} was automatically matched to packages of interest for you"
```
fricklerhandwerk:
```suggestion
        return f"{self.suggestion.cve.cve_id} was automatically matched to packages you subscribed to"
```

The FIXME above explains why it already was the correct amount of wrong. I suggest we fix the underlying problem (in a different PR), then we can be precise and generate the text according to whether the user subscribed to or maintains the packages in question.

src/webview/notifications/context.py
florentc: This builds the data for the package list component (dictionary from attribute to package data) using the same naming conventions attr, pdata as where this component is used in the rest of the project. I think it's better to keep it consistent.

Once we introduce the package data model, this whole function will have to be redone (or even removed) anyway.

src/webview/models.py
florentc: I will restore the initial message because it's not a big deal but the new generic wording seemed adapted to wait for the fix (it encompasses both cases), whereas the previous one may state something that is wrong (in case the package is just a maintained one).

src/webview/templates/notifications/components/notification.html
florentc: It works well with some bold for the title :+1:

src/webview/notifications/views.py
florentc: Yes

src/webview/templates/notifications/components/notification.html
florentc: This is to be consistent with how it's done for suggestion templates. Now everything is in `NotificationContext` (including `new_unread_count` and the `NotificationContext` object is what's passed as the `data` key of the `notification` component. There is no redundancy.

Besides, if ultimately we redesign our templates workflow, at least the starting point will be more consistent and easier to refactor.

src/shared/tests/conftest.py
florentc: True! I introduced methods in profile and used them also in the subscription page views.

PR #871, opened 2026-03-13
Schema versioning for cached suggestions

### LINKED ISSUE #769: Schema versions for cached suggestions
Right now we're just regenerating all cached values when the schema changes. Ideally we'd either periodically or on demand redo the caching whenever we deploy a new schema.

Related (and probably depends on):
- https://github.com/NixOS/nix-security-tracker/issues/768

DarshanCode2005:
This introduces schema versioning for `CachedSuggestions` so cache entries can be invalidated incrementally instead of deleting and rebuilding the entire cache when the payload schema changes.

Each cached row now stores a `schema_version`. Rows with a version lower than `CURRENT_SCHEMA_VERSION` are treated as stale and regenerated.

Changes:

* add `schema_version` field and `is_stale` property to `CachedSuggestions`
* introduce `CURRENT_SCHEMA_VERSION`
* stamp new cache entries with the current schema version
* update `regenerate_cached_suggestions` to rebuild only stale or missing entries
* ensure cache validity in the view layer before rendering
* add tests for version stamping and staleness behavior

The migration defaults `schema_version` to `0`, so existing rows automatically become stale after migration.

Related Issue:

Resolves #769


src/shared/models/cached.py
```diff
@@ -19,3 +21,9 @@ class CachedSuggestions(TimeStampMixin):
 
     # The exact format of this payload will change until it's properly defined.
     payload = models.JSONField(encoder=DjangoJSONEncoder)
+
+    schema_version = models.PositiveIntegerField(default=0)
```
fricklerhandwerk, line 35:
```suggestion
    schema_version = models.PositiveIntegerField(default=CURRENT_SCHEMA_VERSION)
```

We could set this here and remove that in [the unrelated place](https://github.com/NixOS/nix-security-tracker/pull/871/changes#diff-590fdacaf25146d42224ce6e9b3208f11ad24285b25fa8eca148c55dc677187cR19)

src/shared/models/cached.py
```diff
@@ -3,6 +3,8 @@
 
 from shared.models.linkage import CVEDerivationClusterProposal
 from shared.models.nix_evaluation import TimeStampMixin
+# Version must be bumped whenever the CachedSuggestion Pydantic schema changes.
+CURRENT_SCHEMA_VERSION = 1
```
fricklerhandwerk: Maybe unnecessary, so very optional: have you considered making this a `@classproperty`? This would allow putting it right above `payload` in the model.

Also about comment placing: the goal is for contributors to not overlook this when changing the schema, so really the comment should live where the schema is touched.

src/webview/suggestions/views/base.py
```diff
@@ -18,11 +19,20 @@ def fetch_suggestion(suggestion_id: int) -> CVEDerivationClusterProposal:
...
+        cache_new_suggestions(suggestion)
+        suggestion.refresh_from_db()
+
+
 def get_suggestion_context(
     suggestion: CVEDerivationClusterProposal,
     can_edit: bool,
     is_compact: bool = False,
 ) -> SuggestionContext:
+    ensure_fresh_cache(suggestion)
```
fricklerhandwerk: @DarshanCode2005 you can't know this because it's hard to see from context (hah), but those classes named `*Context` are supposed to be pure wrappers that encode the user-visible data structures. So the transformations shouldn't have side effects. Therefore the call here is misplaced. I know, from a dataflow perspective this seems nice because all rendering is funneled here, but... So IMO this should go into the views, possibly `get_queryset`, not sure. My intuition is: where the DB operations happen.

(@florentc this needs a note in the contributing guide)

src/webview/suggestions/views/base.py
```diff
@@ -18,11 +19,20 @@ def fetch_suggestion(suggestion_id: int) -> CVEDerivationClusterProposal:
     return get_object_or_404(CVEDerivationClusterProposal, id=suggestion_id)
 
 
+def ensure_fresh_cache(suggestion: CVEDerivationClusterProposal) -> None:
+    """Regenerate stale or missing cache before the view renders."""
+    cached = getattr(suggestion, "cached", None)
+    if cached is None or cached.is_stale:
```
fricklerhandwerk: Hm, what speaks against putting this `or` into `is_stale`?

src/webview/suggestions/views/base.py
```diff
@@ -18,11 +19,20 @@ def fetch_suggestion(suggestion_id: int) -> CVEDerivationClusterProposal:
     return get_object_or_404(CVEDerivationClusterProposal, id=suggestion_id)
 
 
+def ensure_fresh_cache(suggestion: CVEDerivationClusterProposal) -> None:
```
fricklerhandwerk: Also, why not make this a method on the model? Anything speaking against that? (Not asking rethorically, possibly there's a good reason to do it that way and I'm not seeing that.)

src/shared/models/cached.py
DarshanCode2005: 1. Got it, I'll update the comment placement accordingly.
2. Scoping the version constant inside the class make sense. Also, a plain class const is probably the simplest path, no extra imports required. (Just share some final thoughts over this).

src/shared/models/cached.py
DarshanCode2005: * Default value decision: Using default=CURRENT_SCHEMA_VERSION, i guess in future this will create new migrations (which could be unnecessary). Keeping defualt = 0 avoids this and will ensure existing cached rows are automatically marked as stale without extra work.
* Encapsulation stuff: I can add a custom QuerySet manager to the model (example, CachedSuggestions.objects.stale()). That way, the management command can just call .stale() and we successfully encapsulate all schema version logic inside the model layer.
Does this approach sound good?

src/webview/suggestions/views/base.py
DarshanCode2005: Since `is_stale` is a property on the model, if the cache doesn't exist yet, it can cause error. However , I understand your concern for encapsulating this! I can move this logic to a `@property` named `cache_stale` directly on the `CVEDerivationCLusterProposal` model. That way , it can internally check if the cache is missing or if the version is outdated. Just let me know if it's alr?

src/webview/suggestions/views/base.py
DarshanCode2005: Just for keeping things separate, why not keep the core model focused strictly on the schema and let a util handle the side effects of rebuilding the caching layer. Just a POV, you are obviously more experienced than me, please correct me if I am wrong.

src/webview/suggestions/views/base.py
fricklerhandwerk: Ah, I read it too quickly and thought the `@property` already *is* on `CVEDerivationClusterProposal`. So yes, let's move it there!

src/webview/suggestions/views/base.py
fricklerhandwerk: My intuition would be to keep the data stuff together, and not have the view think about whether the data is any good. All the views should do is declare what data they want.

Also mid- to long-term we should get rid of the cache (by improving the data layout to reduce access times), it creates a lot of headache. So when ripping it out some internals about how data is stored shouldn't touch the views.

Another angle is that if we had a second app, say for a REST API, it would also want to use the cache in the same way.

So arguably the cache stuff should be part of the data model.

src/shared/models/cached.py
fricklerhandwerk: How do you imagine a class constant on Django model classes? ;)

src/shared/models/cached.py
fricklerhandwerk: Default value: Oh yes right, good catch.

QS manager: good idea!

src/shared/models/cached.py
DarshanCode2005: I'll go with your approach :+1: , class constant might accidentally evolve into something , that django treats specially.

src/shared/management/commands/regenerate_cached_suggestions.py
```diff
@@ -14,12 +14,22 @@ class Command(BaseCommand):
...
+            f"Regenerating {len(to_regenerate)} entries "
+            f"({len(stale_ids)} stale, {len(missing_ids)} missing)"
+        )
 
         for suggestion in (
-            CVEDerivationClusterProposal.objects.all()
+            CVEDerivationClusterProposal.objects.filter(pk__in=to_regenerate)
             .order_by("-updated_at")
             .iterator()
         ):
```
florentc: If I understand this correctly, we only re-cache suggestions which either lack a cached suggestion, or have a cached suggestion belonging to an older schema version.

I think we should still offer an option through a flag to force regenerate everything, regardless of schema version.

**Regenerating in case we manually mess up with cached data:**
If we introduce a bug where views update cached suggestions in a way that is inconsistent with what should be in cached suggestion, then we'd like to regenerate everything although there's been no schema version bump.
For instance, we have `MaintainersEdit` in database which are used to compute lists of maintainers in cached suggestions. In a view that adds a `MaintainersEdit`, we would probably just remove/add the maintainer in the cached suggestion maintainers lists directly instead of regenerating everything. But imagine we do it wrong and deploy. Then we have a bunch of corrupted cached_suggestions that need to be regenerated after the fix, although the schema version did not change.

**Quality of life when developing:**
For development purposes, when working on a PR that modifies the schema, we will likely iterate through modifications and it would be inconvenient to change the version number all the time. Being able to force regenerate everything regardless of version number would be more convenient on our local dev environments.

**Regenerating in case of mismanagement of version numbers**
People may forget to update the version number or the version number might get messy after merging to concurrent PRs that update the schema with the same new version number. We could still push a PR to bump the version number which would trigger the regeneration but I feel like we may want to have the option to just force regenerate in those cases.

src/shared/listeners/cache_suggestions.py
```diff
@@ -24,6 +24,7 @@
 logger = logging.getLogger(__name__)
 
 
+# Version must be bumped whenever the CachedSuggestion Pydantic schema changes.
```
florentc: I think this should be more prominent. I don't know how though. It is likely we will forget to update the version number, especially since this module is big and we may land in it from a "go to definition" right in the middle.

Ideally in the future, we would have a fancy hook which detects changes to the schema and warns/errors when the version hasn't been a version bump.

src/shared/management/commands/regenerate_cached_suggestions.py
fricklerhandwerk: Agreed, an `--all` flag that will purge them first would be in order.

src/shared/listeners/cache_suggestions.py
DarshanCode2005: @florentc Sir and @fricklerhandwerk Sir , do you have any suggestion on this?

Adding the hook , will be a little more involved right now. How about a warning or important comment with some :warning: emoji would do.

```
# ============================================================================
# ⚠️ IMPORTANT: CACHE SCHEMA VERSIONING
#
# When modifying this schema, you MUST bump
# CachedSuggestions.CURRENT_SCHEMA_VERSION.
#
# Otherwise cached data may become inconsistent.
# ============================================================================
class CachedSuggestion(BaseModel):
```

src/shared/models/linkage.py
```diff
@@ -93,6 +106,11 @@ def restore_package(self, package: str) -> None:
             edit_type=PackageEdit.EditType.REMOVE,
         ).delete()
 
+    class Meta:  # type: ignore[override]
+        indexes = [
+            BTreeIndex(fields=["status"]),
+        ]
+
```
```suggestion
```

Unrelated change?

src/shared/tests/test_suggestion_caching.py
```diff
@@ -70,3 +70,31 @@ def test_caching_newest_package(
...
 
     assert package["major_version"] == "2.0"
+
+
+def test_cache_stamped_with_current_version(
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    cache_new_suggestions(suggestion)
+    cached = CachedSuggestions.objects.get(proposal=suggestion)
+    assert cached.schema_version == CachedSuggestions.CURRENT_SCHEMA_VERSION  # type: ignore
```
fricklerhandwerk:
```suggestion
    assert cached.schema_version == CachedSuggestions.CURRENT_SCHEMA_VERSION
```

This particular one seems not to be needed.

fricklerhandwerk:
```suggestion
    assert cached.schema_version == CachedSuggestions.CURRENT_SCHEMA_VERSION  # type: ignore
    assert cached.is_stale is False
```

Then the third test isn't needed.

src/webview/suggestions/views/lists.py
```diff
@@ -77,6 +77,7 @@ def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         suggestion_ids = [s.pk for s in page_obj.object_list]
         events_by_suggestion = fetch_suggestion_events(suggestion_ids)
         for suggestion in page_obj.object_list:
+            suggestion.ensure_fresh_cache()
```
fricklerhandwerk:
```suggestion
```

This will be too slow. We're already not even loading uncached ones. But now you can expand the filter to skip stale cache values as well.

src/webview/suggestions/views/detail.py
```diff
@@ -23,6 +23,7 @@ class SuggestionDetailView(DetailView):
     def get_context_data(self, **kwargs: Any) -> dict[str, Any]:
         context = super().get_context_data(**kwargs)
         user_can_edit = user_can_edit_suggestion(self.request.user)
+        self.object.ensure_fresh_cache()  # type: ignore
```
fricklerhandwerk: This would be better suited in an override of `self.get_object()`. `get_context_data` is for preparing the fetched data to be used in templating, not fetching stuff.

src/shared/listeners/cache_suggestions.py
florentc: The hook would be icing on the cake but not a requirement for this PR. The more visible warning you propose is fine IMO

src/shared/tests/test_suggestion_caching.py
DarshanCode2005: yeah , I removed it and ran pyright and it worked.

DarshanCode2005: Yupp! removed the extra test. Thanks!

PR #873, opened 2026-03-14
feat: explain pgpubsub listener registration pattern in CONTRIBUTING.md

DarshanCode2005:
Summary:
This PR adds a new section to new section to Contributing guideline that documents the required pattern for registering pgpubsub listeners.

Thought Process:
In this project, pgpubsub listeners must be explicitly imported during the appliication's startup phase to be correctly registered. Currently, this is handled by:
1. `src/shared/apps.py` imports `shared.listeners`.
2. `src/shared/listeners/__init__.py` importing individual listener modules.

New contributors adding listeners in new files might overlook this requirement, leading to listeners that silently fail to trigger. Explicitly documenting this step helps prevent debugging frustration and ensures consistent implementation across th code.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -232,7 +232,7 @@ This is the default Django workflow.
 
 ## `pgpubsub` listener registration pattern
 
-The application uses `django-pgpubsub` to react to database changes asynchronously.
+The application uses [`django-pgpubsub`](https://github.com/PaulGilmartin/django-pgpubsub) to react to database changes asynchronously.
 Listeners are defined as functions decorated with `@pgpubsub.post_insert_listener`, `@pgpubsub.post_update_listener` etc., and are primarily located in the [`src/shared/listeners/`](src/shared/listeners/) directory.
 
 To ensure your listener is proactively registered when the Django application starts, its containing module must be imported.


CONTRIBUTING.md
```diff
@@ -230,6 +230,24 @@ manage migrate
 
 This is the default Django workflow.
 
+## `pgpubsub` listener registration pattern
+
+The application uses `django-pgpubsub` to react to database changes asynchronously. Listeners are defined as functions decorated with `@pgpubsub.post_insert_listener`, `@pgpubsub.post_update_listener` etc., and are primarily located in the `src/shared/listeners/` directory.
```
fricklerhandwerk: In documentation, please always link names of technical components to authoritative sources. Searching for what you may mean (especially if it's an everyday term, doesn't apply here but happens more often than you'd think) needlessly adds friction.

Especially link to local files.  Not applicable here, and more of a side note: Ideally we should somehow avoid having to reference files altogether as long as there's no link checking (since refactoring will break those links without notice to the person making the change).

Also please reformat to one sentence = one line, makes a lot easier to make targeted suggestions in reviews.

CONTRIBUTING.md
```diff
@@ -230,6 +230,24 @@ manage migrate
...
+2. Import the module inside `src/shared/listeners/__init__.py` so it's loaded as part of the package:
+
+   ```python
+   # inside src/shared/listeners/__init__.py
+   import shared.listeners.my_new_listener  # noqa
+   ```
+
+3. `src/shared/apps.py` triggers these imports in its `ready()` method by importing `shared.listeners`, registering all listeners upon app initialization.
+
+**Warning:** If you create a new listener module but forget to add its import to `src/shared/listeners/__init__.py`, your listener will fail to run silently!
```
fricklerhandwerk: Use GitHub callout syntax here (note how annoying it is to figure out what I mean without a link, especially if you haven't encountered that particular concept before)

CONTRIBUTING.md
```diff
@@ -230,6 +230,27 @@ manage migrate
 
 This is the default Django workflow.
 
+## `pgpubsub` listener registration pattern
+
+The application uses `django-pgpubsub` to react to database changes asynchronously.
```
fricklerhandwerk:
```suggestion
The application uses [`django-pgpubsub`](https://github.com/PaulGilmartin/django-pgpubsub) to react to database changes asynchronously.
```

PR #880, opened 2026-03-19
chore: Include git in list of shell packages

adekoder:
Description
Explicitly add `pkgs.git` to the dev shell packages in `default.nix`.

## Why

The `fetch_all_channels` management command and the Nix evaluation pipeline
(`shared/git.py`) both rely on git to clone and fetch Nixpkgs commits.
Without this change, the dev shell falls back to the system git which may
be too old `git fetch --porcelain` requires git 2.41+ and silently breaks
on older versions (e.g. macOS ships git 2.39.2 which does not support this flag).

Pinning git through nixpkgs ensures every developer uses a known-good version
regardless of what is installed on their machine.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -94,6 +94,12 @@ rec {
       packages = [
         manage
         package
+        # Explicitly pin git from nixpkgs to ensure the `fetch_all_channels` management command
+        # and the Nix evaluation pipeline (which clones and fetches Nixpkgs commits via shared/git.py)
+        # always use a known-good version. Without this, the shell falls back to the system git,
+        # which may be too old to support required flags (e.g. `git fetch --porcelain` requires git 2.41+)
+        # and can cause silent failures or broken behaviour in development.
+        pkgs.git
         pkgs.nix-eval-jobs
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -23,3 +23,4 @@ nixpkgs-gc-roots/
 .envrc-local
 # user settings file
 .settings.py
+.DS_Store
```
fricklerhandwerk: Don't add files specific to your system. Either [configure it globally](https://docs.github.com/en/get-started/git-basics/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer) for your Git, or per project with https://git-scm.com/docs/gitignore#_description

adekoder: That is why i added this line in the `.gitignore` file , but i have remove it using system wide config.

fricklerhandwerk: Oh sorry I was too coarse with the Git docs link. I meant `$GIT_DIR/info/exclude`, which is not tracked

adekoder: i already resolved this btw.

PR #887, opened 2026-03-19
docs: add guideline on contribution culture

fricklerhandwerk:
After many years in software maintenance, I realized that most of the time reviewing is spent deciding whether doing the things proposed in a change are a good idea. And that accepting a change means taking responsibility for the consequences. The more decisions, the longer it takes. Hard decisions block easy decisions, especially if they're unrelated.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ nix-shell --run format
 
 ## Contribution culture
 
-A pull request asks a maintainer to accept responsibility for a decision.
+A pull request asks maintainers to accept responsibility for a decision.
 Help them understand what they're agreeing to.
 
 To minimise turnaround time for getting your contribution merged:


CONTRIBUTING.md
```diff
@@ -20,6 +20,24 @@ Run the formatter manually with:
 nix-shell --run format
 ```
 
+## Contribution culture
+
+Please bear in mind that reviewing contributions primarily means validating (and often discussing) the decisions they contain.
+Every pull request is saying "we should do _this_ like _that_ from now on".
```
florentc: I don't understand this sentence "Every pull request is saying..." nor what to take from it

Is it something you criticize? You mean that usually PRs are too assertive?

I feel it gives some kind of "upset" tone.

CONTRIBUTING.md
```diff
@@ -20,6 +20,24 @@ Run the formatter manually with:
...
+
+Therefore:
+
+- Make as few decisions as possible in each pull request, ideally exactly one.
```
florentc: I think you should clarify what you mean by decision. An example PR with a few commits and a PR title maybe?

CONTRIBUTING.md
```diff
@@ -20,6 +20,24 @@ Run the formatter manually with:
...
+- Describe the decision in the commit message title.
```
florentc: You mean the PR title or commit title? You would like to aim for 1 decision per PR so is it like saying 1 commit per PR or not a lot more?

CONTRIBUTING.md
```diff
@@ -20,6 +20,24 @@ Run the formatter manually with:
...
+- If the reason is not trivial, explain _why_ the change is made in the commit message and the pull request description.
+- Don't describe _what_ the change is, this is evident from the change itself.
+- Strive to keep the diff small, ideally less than 100 lines.
+  Larger changes typically mean that you made too many decisions at once.
```
florentc: As you imagine, I'm not 100% aligned with you on this one. If I take my last PR for instance https://github.com/NixOS/nix-security-tracker/pull/888 there is a ~300loc diff. Yet I can't see how I would bring this down to below 100, unless I remove comments, or split into PRs such as:

1. data model change
2. view change + basic test
3. proper UI elements

As for 1:
- it would add unsused bits in the data model, probably with a TODO in code. If I have to work on things deemed more urgent, it might end up being dead code / noise
- maybe when implementing 2 I realize that in fact the data model choice was not the best choice and I want to change my mind so in the end pre-merging 1 wasn't so useful

As for 2:
- I'm not sure 2 would fit below 100 and the feature needs to be 100% implemented because we deploy every time we merge

As for 3:
- Before it's merged, users may have to deal with prototype UI which may be unintuitive or glitchy.
- Once 1 and 2 are merged, it is likely that 3 will never be considered important enough to get attention and might never happen.


I'm not advocating for 1000loc PRs. I fully agree with the general idea, having 1 purpose PRs and aiming to divide into the smaller possible chunks. But sometimes 1 purpose takes more than 100loc, and forcing to split it to fit a rule, IMO, may be counter productive in time and quality.

I don't think tiny PR policy reduces the amount of conflicts to resolve. PRs are smaller but there are more.

I think a strict tiny PR policy, if pushed as a rule or dogma, can bring a tendency for short-sighted design choices, noisy inconsistent code, dead code, half baked features, and a deployed application (since we deploy on the go without releases)  that feels broken/unfinished.

I think we should ask contributors to split their contributions into the tiniest possible chunks, but not give them a target or limit. The 100loc target and the "if it's above you made too many decisions" seems arbitrary/dogmatic (or not, maybe there is research about this I don't know)

florentc (commented):
I think we should add an item about not shoving unrelated things in a commit or a PR. For instance while developing a feature you notice a typo or simple thing to refactor in a module that has nothing to do with your PR and you feel the urge to implement it.
I think we should stress that 1 commit = 1 unit of semantic change, not 1 commit = a few things I did in 1 file.
Maybe this is what you meant by "decision"?

I'd also like to add an item about aiming to maintain the codebase as consistent as possible. Avoiding pattern breaks when they don't bring that much.


Is there already an official "contribution culture" for the nix community projects? Is this inspired by it?

CONTRIBUTING.md
fricklerhandwerk: No, I want to say that this is what a PR amounts to, semantically. One could phrase it more politely, but it takes more words: Here is a suggestion for how to do things from now on, what do you think?

This is my take on how to think about the whole contribution cycle, or how to think about what we're actually doing here. That we're really just encoding decisions into a machine so it can do them for us repeatedly from now on.

The rest of the idea behind this PR is to establish that we should not lump such suggestions (the "decisions" made in a PR) together, because practical experience shows that what takes most time is deliberating whether to accept them; having to make many decisions about one packaged proposal makes the package land slower. And easy decisions get blocked by hard decisions (that need more time to think about or need discussion).

CONTRIBUTING.md
fricklerhandwerk: Oh, some earlier version had both PR and commit title... not sure why that disappeared, will fix.

CONTRIBUTING.md
fricklerhandwerk: Yes I expected this kind of reply, and I fully agree with you on all points. The idea behind this phrase is to be a guideline for when it's likely that a change tries to do too many things (or when a contributor made too many decisions) at once. But it's true that such as in your example, one idea takes more lines to express.

It's of course not a hard limit. And of course it doesn't make sense to add dead code. (Although that has nuance too: I would for instance make exceptions for tests: if we know we want a feature, adding a failing test that describes it precisely seems productive to me. It's a way of encoding a decision into the code.) Should I rephrase it to say something like "try to minimize the number of lines changes"? There's more detail, such as a mechanical change like string replacement that is huge but has very little information content...

Maybe I'll do this: remove the contentious parts here because they add yet more ideas to an already new one.

Issue #856, opened 2026-03-07
Dismiss as "not in Nixpkgs"


fricklerhandwerk:
As a security team member, I want to dismiss a suggestion that doesn't match anything in Nixpkgs with one click.

## Acceptance criteria

Given an automatic suggestion,
When I dismiss it as "not in Nixpkgs",
Then
- I don't have to supply a freeform comment
- All matched packages are ignored automatically
- The suggestion is set to state "dismissed"
- The dismissal reason is recorded in the database

Bonus points if the dismissal reason is displayed on the suggestion.

PR #888, opened 2026-03-20
 feat: dismiss with rejection reason (for now not-in-nixpkgs)

florentc:
This implements #856.
**Warning** This introduces database migrations

The "Dismiss" button now has an attached dropdown menu to choose a rejection reason. The default is comment and behaves like before.

-------------------

The "Not in nixpkgs" reason allows the user to dismiss directly without a comment. This reason will be saved in the object in database.

-------------------

The rejection reason is displayed next to the "Dismiss" status. When usually this status is not displayed (for instance in the "Dismissed suggestions" list, if there is a reason, it will be displayed anyway. Otherwise, like before, it's possible to refer to the comment if no builtin reason has been chosen.

-------------------

The rejection reason is also saved in the activity log and displayed there when applicable.

-------------------

New CSS utility classes are introduced for grouping buttons with other widgets horizontally. These have been used in other parts of the UI to improve experience and consistency.

Add a maintainer:

Subscribe to a package:

### TODO

- [x] Tests

### DIFF
--- a/src/webview/templates/subscriptions/components/packages.html
+++ b/src/webview/templates/subscriptions/components/packages.html
@@ -7,7 +7,7 @@
   {% endif %}
 
   <!-- Add new subscription form -->
-  <form method="post" action="{% url 'webview:subscriptions:add' %}" hx-post="{% url 'webview:subscriptions:add' %}" hx-target="find .join" hx-swap="outerHTML" hx-indicator="find button" class="row">
+  <form method="post" action="{% url 'webview:subscriptions:add' %}" hx-post="{% url 'webview:subscriptions:add' %}" hx-target="closest .package-subscriptions-container" hx-swap="outerHTML" hx-indicator="find button" class="row">
     <div class="row join join-green">
       <input type="text" name="package_name" placeholder="Package name" required class="package-input rounded-box-compact join-item">
       {% csrf_token %}


src/shared/models/linkage.py
```diff
@@ -33,6 +36,10 @@ class RejectionReason(models.TextChoices):
             "exclusively_hosted_service",
             _("exclusively hosted service"),
         )
+        NOT_IN_NIXPKGS = (
+            "not_in_nixpkgs",
+            _("not in nixpkgs"),
```
```suggestion
            _("not in Nixpkgs"),
```

src/webview/suggestions/views/status.py
```diff
@@ -60,19 +65,41 @@ def post(self, request: HttpRequest, suggestion_id: int) -> HttpResponse:
...
-            if not new_comment and not undo_status_change:
+            if not rejection_reason:
+                # Without a rejection reason, we require a comment
+                if not new_comment and not undo_status_change:
+                    return self._handle_error(
+                        request,
+                        suggestion_context,
+                        "You must provide a dismissal comment",
+                    )
+            elif (
```
fricklerhandwerk: Can't we do like `if rejection_reason in C: suggestion.rejection_reason = C[rejection_reason]`?

fricklerhandwerk (commented):
Very nice! Waiting for tests

src/shared/models/linkage.py
florentc: Done

src/webview/suggestions/views/status.py
florentc: Done

PR #889, opened 2026-03-20
refactor: clarify distinction between edit permissions and frozen state

abdella-h:
Flip the boolean so `frozen=True` means the suggestion is locked (REJECTED or PUBLISHED). Add a comment clarifying that `frozen` reflects the suggestion's status while `can_edit` reflects the user's permission.


src/webview/suggestions/context/types.py
```diff
@@ -27,8 +27,9 @@ class Reference:
 class PackageListContext:
     active: dict
     ignored: dict
-    editable: bool
-    # FIXME(@fricklerhandwerk): Arguably the same thing as `editable` if both views and template handle it right.
+    frozen: bool 
+    # True when the suggestion's status no longer allows changes (REJECTED or PUBLISHED).
+    # Distinct from 'can_edit' which reflects whether the current user has permission to make edits.
```
fricklerhandwerk:
```suggestion
    # True when the suggestion's status no longer allows changes (REJECTED or PUBLISHED).
    frozen: bool
    # True when the current user has permission to make edits.
```

Convention is to put comments above parameters. Arguably inconsistent with function docstrings and against reading order, but due to deeply ingrained habits it's more readable for most in practice.

src/webview/suggestions/context/types.py
```diff
@@ -118,7 +121,7 @@ def update_package_list_context(
                 for k, v in self.suggestion.cached.payload["original_packages"].items()
                 if k not in active_packages
             },
-            editable=self.suggestion.is_editable,
+            frozen = not self.suggestion.is_editable,
```
fricklerhandwerk: You should also change the `@property` on the suggestion model; this is where the truth lives. Then no need to actually describe anything in the comments, since the code already says what happens, and the SuggestionContext uses the same naming as the model

src/webview/suggestions/context/types.py
```diff
@@ -27,8 +27,9 @@ class Reference:
 class PackageListContext:
     active: dict
     ignored: dict
-    editable: bool
-    # FIXME(@fricklerhandwerk): Arguably the same thing as `editable` if both views and template handle it right.
+    frozen: bool 
+    # True when the suggestion's status no longer allows changes (REJECTED or PUBLISHED).
+    # Distinct from 'can_edit' which reflects whether the current user has permission to make edits.
     can_edit: bool
```
fricklerhandwerk: If we rename this to `user_can_edit`, the comment also becomes superfluous.

fricklerhandwerk (commented):
This is on the way to becoming a strong readability improvement!

PR #890, opened 2026-03-20
feat: add lychee in pre-commit

DarshanCode2005:
Script for lychee setup , now lychee is the part of `nix-shell --run format`


nix/git-hooks.nix
```diff
@@ -61,5 +61,19 @@
...
         enable = true;
       };
+    }
+    // {
+      lychee = {
+        enable = true;
+        name = "lychee";
+        entry = "${pkgs.lib.getExe pkgs.lychee} --offline --no-progress";
+        files = "\\.md$";
+        pass_filenames = true;
```
fricklerhandwerk:
```suggestion
```

This is the default

Issue #617, opened 2025-07-17
Task: Package data model


fricklerhandwerk:
Right now we're aggregating derivations ad hoc into some ad hoc notion of "package". But in the UI we have a standalone idea of "package" which we should formalize.

The presentation of a package is currently done through a weird `channel_structure()`. We should probably replace this with something like

```
Package {
   attribute :: String # attribute path of the package
   drvs :: Set NixDerivation
}
```

This should live as a cached value (not be part of the database models) since the underlying raw data can be computed from `NixDerivation`. We will need to refine this as we go, see https://github.com/Nix-Security-WG/nix-security-tracker/issues/391 for gotchas.

For the UI we want more something like this:

```
Package {
   attribute :: String # attribute path of the package
   description :: String # description from the latest known version
   channels :: Map Channel NixDerivation # this is how we see this in the UI
}
```

We may end up having to move this to the database for https://github.com/Nix-Security-WG/nix-security-tracker/issues/618 though.

fricklerhandwerk:
Discussed the design a bit with @florentc today:
- Another observation is that packages part of package sets (e.g. `python312` or `linux_6_18`) will have the same source location per checkout. We may leverage this to form equivalence classes.
  - Not clear how to use this in the UI though. E.g. we may want to show for a given (prefix-free) package name which package sets it's in. Or, opposite, hide duplicates that are inside package sets, to reduce the clutter.
  - There are also *variants* (e.g. `qemu_xen`, `qemu_kvm`), which may still have the same source location -- those may have the same semantics as package sets but a different naming convention. As opposed to packages in package sets that collect packages for some platform, variants also belong to a platform, but that platform isn't represented as a package set, or one could say the platform -> package mapping is reversed.  
- We probably want to have the full attribute path as primary key, and then see from there

fricklerhandwerk:
Another update after staring at data for what's obviously too long:
- Some of the metadata is about packages (which are actually an external entity, hence our record linkage problem), some is about derivations (which are specific to a commit):
  - Package (where we're interested only in the latest value):
    - maintainers
    - homepage
    - description
    - name, attribute name (there may be multiple but they all refer to the same package)
    - platforms
  - Derivation (which are bound to a certain point in time)
    - version
    - system
    - licenses (a property of the source that goes into a derivation)
    - derivation path, dependencies, outputs
    - knownVulnerabilities
    - all computed flags: insecure, available, broken, unfree, unsupported
[...]

PR #892, opened 2026-03-20
refactor: rename automatic_linkage module and its internal functions

adekoder:
Description

Renames `automatic_linkage.py` to `cve_derivation_matcher.py` and updates internal function names to better reflect their intent: `build_new_links_following_new_containers` -> `match_derivations_on_container_insert`, `build_new_links` -> `create_derivation_proposal`, `produce_linkage_candidates` -> `find_linkage_candidates`. Updates all call sites in `__init__.py`, `propose_cve_links.py`, and `test_linkage.py`.

We are doing this to make the the module more clear and understandable in-terms of what it does.
No logic changes in the refactor, pure rename and restructure

### DIFF
--- a/src/shared/listeners/cve_derivation_matcher.py
+++ b/src/shared/listeners/cve_derivation_matcher.py
@@ -16,7 +16,7 @@ from shared.models.nix_evaluation import (
 logger = logging.getLogger(__name__)
 
 
-def produce_linkage_candidates(
+def find_linkage_candidates(
     container: Container,
 ) -> dict[NixDerivation, ProvenanceFlags]:
     latest_complete_channels = (
@@ -93,22 +93,20 @@ def produce_linkage_candidates(
     return candidates
 
 
-def build_new_links(container: Container) -> bool:
+def create_derivation_proposal(container: Container) -> bool:
     if container.cve.triaged:
         logger.info(
-            "Container received for '%s', but already triaged, skipping linkage.",
-            container.cve,
+            f"Container received for {container.cve}, but already triaged, skipping linkage.",
         )
         return False
 
     if CVEDerivationClusterProposal.objects.filter(cve=container.cve).exists():
-        logger.info("Suggestion already exists for '%s', skipping", container.cve)
+        logger.info(f"Suggestion already exists for {container.cve}, skipping")
         return False
 
     if container.tags.filter(value="exclusively-hosted-service").exists():
         logger.info(
-            "Container for '%s' is exclusively-hosted-service, rejecting without match.",
-            container.cve,
+            f"Container for {container.cve} is exclusively-hosted-service, rejecting without match.",
         )
         CVEDerivationClusterProposal.objects.create(
             cve=container.cve,
@@ -117,16 +115,14 @@ def build_new_links(container: Container) -> bool:
         )
         return True
 
-    drvs = produce_linkage_candidates(container)
+    drvs = find_linkage_candidates(container)
     if not drvs:
-        logger.info("No derivations matching '%s', ignoring", container.cve)
+        logger.info(f"No derivations matching {container.cve}, ignoring")
         return False
 
     if len(drvs) > settings.MAX_MATCHES:
         logger.warning(
-            "More than '%d' derivations matching '%s', ignoring",
-            settings.MAX_MATCHES,
-            container.cve,
+            f"More than {settings.MAX_MATCHES} derivations matching {container.cve}, ignoring",
         )
         return False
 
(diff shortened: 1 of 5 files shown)


src/shared/listeners/cve_derivation_matcher.py
```diff
@@ -93,22 +93,20 @@ def produce_linkage_candidates(
     return candidates
 
 
-def build_new_links(container: Container) -> bool:
+def create_derivation_proposal(container: Container) -> bool:
```
fricklerhandwerk, line 96:
```suggestion
def create_suggestion(container: Container) -> bool:
```

I'm so sorry, the naming is a mess in this code base. We're still cleaning up the prototype-era fallout. @florentc do we agree that we've firmly converged on calling the automatic matchings "suggestions"?

fricklerhandwerk (commented):
Thanks, the renames are helpful! If you want to make the logger calls uniformly use f-strings, please do all of them in one separate PR.

src/shared/listeners/cve_derivation_matcher.py
adekoder: but isn't suggestion in the context of how the frontend treat matches ?

fricklerhandwerk: A(n automatically generated) suggestion links CVEs and (currently) derivations (we want that to be packages #617), we also call these links matches. We call the process "linking" or "matching". Ideally the frontend would directly visualise that data model.

The code is still all over the place, sometimes they are called "proposals" or "matches" or "links". I think we should stick with "suggestions", but "matches" is not wrong either (but would require starting to think about redirects from deprecated URL patterns).

florentc: I'm biased as I spend most of my time in `webview` rather than `shared`. In `webview`, clearly, "suggestion" is the term that's used to refer to a `CVEDerivationClusterProposal` and its cached information. Stuff that is called suggestion usually has type `CVEDerivationClusterProposal`. It used to also refer to only the cached data. I tried to bring some consistency these last few months by always calling suggeston the `CVEDerivationClusterProposal` and access the cached field and the payload explicitly. There might still be some leftover.

So yeah, in `webview` at least, it would totally make sense to rename `CVEDerivationClusterProposal` to `Suggestion`.

In `shared` however I can't say with as much confidence. In the backend at the lower level, it's true that calling it "match" gives a better intuition about what it is. But this object has fields like "comment" and "status" which clearly go beyond the description of a match. Should we maybe have a `Match` table and and `Suggestion` table, each suggestion having a match (the CVE and Derivations) and adding fields describing its current state (status, comment, PackageEdits, MaintainersEdits, etc) in the suggestion workflow?

Sorry I'm not helping, just brainstorming too

fricklerhandwerk: Separating matches and "the thing that people edit" (I'm inclined to call that a "draft" as it was before in the UI, even at the risk of leaving the question "but draft of what?" or possibly "draft issue"). Then this draft would be part of a published issue, which tacks on even more data.

Most importantly this would be the first step towards having issues with multiple matches!

florentc: Indeed! The draft could be linked to several matches, we could have an "add another match to the draft" UI thing, and the issue would be linked to 1 draft.

In fact, maybe issue and draft could be merged. That would clear any naming weirdness.

Match: a CVE the system matched to some derivations (and the associated packages/severity/maintainers)

Issue: A thing the user edits. Associated to 1 or several matches. Which could be in various status: untriaged, dismissed, draft, and published. When in "dismissed" or "published" the issue is frozen like suggestions are now (no edits).

Publication: What becomes attached to an issue once in "published" status. Publication would have a github link, and additional data such as open/closed, tags, and whatever we want to display (on sec tracker) is happening on GitHub about this issue.

fricklerhandwerk: Sounds good. We can pull out `cve` and `derivations` into a `Match` model. We just need to think about what happens with the ignored/additional items, since right now they live on what would end up as `Issue` and thus apply to all of the matches. Is that what we want?

florentc: For the record, here is what would happen with a basic approach to the problem where `Match` are immutable and `Issue` proposes to apply `...Edits` to packages and maintainers (and references).

Each `Match` has a set of packages (in fact derivations but viewed grouped by attribute name as packages) and their associated maintainers (and references)

An `Issue` may be associated to several matches.

An `Issue` may have 0, 1 or several `PackageEdits`, `MaintainersEdit`, and `ReferencesEdit`

The current state (and what's displayed) of the lists of packages/maintainers/references of an `Issue` would be the union of all the packages/maintainers/references of all its associated matches, on which additions and removals are applied by the overlay `PackageEdit`/`MaintainersEdit`/`ReferencesEdit` of that `Issue`.


This is, IMO, the straightforward first approach we can take.

fricklerhandwerk:
This has diverged quite a lot. Let's take it step by step with the renames on the next opportunity.

Issue #391, opened 2024-11-29
Improve matching suggestions quality


fricklerhandwerk:
This needs refinement.

We discussed a few approaches to reduce the noise in suggestions

## Ideas

- Filter out `hardware` type CPEs
- Compile hand-curated (allow or deny) list of product names that can be used for filtering
   - e.g. `openshift4/ose-docker-builder` or `rhel{8,9,10}` will never concern us
- Accommodate for Nixpkgs-specifics
  - deduplicate wrapped/unwrapped but still encourage maintainers to fix both
  - drivers and kernel modules contain the kernel version in their attribute names
    - those CVEs will have thousands of matching derivations at the moment
  - handle aliases
  - packages in multiple versions
    - GCC, LLVM, Python are regular offenders
    - we can work with version constraints and help navigation with appropriate rendering
    - https://github.com/Nix-Security-WG/nix-security-tracker/issues/332
      - root dependencies will (rightfully) yield thousands of matches, but denoising that in UI would help
    - (advanced) analyze package variants such as overrides
 - compute Levenshtein distance on matches and cut off after some fixed distance
   - may want to combine it with stripping of package set and other Nixpkgs-specific pre/postfixes
- Leverage crowd-sourced data
  - Once we dismiss derivations from a suggestion, the tracker can suggest to open a pull request for creating a Nixpkgs-native CPE to avoid confusion in the future
   
 # Examples
 
 Substring matching is too dumb:
 
![image](https://github.com/user-attachments/assets/25152837-890c-42ee-9572-33dcf6e5a890)

Low Levensthein distance is too dumb:

![image](https://github.com/user-attachments/assets/c603d87e-35b5-4bca-8d64-1f5082045947)

Too many affected products that don't improve the matching quality but take a lot of screen real estate:

![image](https://github.com/user-attachments/assets/dc3b4d41-431b-4966-8d9a-b84663e025e4)

h0nIg:
@fricklerhandwerk what does Long-term-operation means in months / years? Is this something for the next 1-2 years or something which may land in 6 months?

fricklerhandwerk:
The long-term milestone is merely a collection for items we want to address before committing to a production deployment. This is somewhat detached from the actual development schedule, which will depend on available funding. I'll post updates on [the relevant Discourse thread](https://discourse.nixos.org/t/nixpkgs-supply-chain-security-project/34345) if something noteworthy happens in that regard.

PR #894, opened 2026-03-20
feat: filter out hardware-only CPEs

### LINKED ISSUE #920: Filter out hardware-only CPEs

ArkhamKnight25:
Closes #920

Part of #391

`produce_linkage_candidates` was proposing matches for hardware-only CVEs
just because the product name happened to match a derivation name.

That breaks down for CPEs with `part=h`: things like router or switch
firmware are not Nixpkgs packages, but they could still produce linkage
candidates through substring matching.

This change parses the CPE part and skips affected products whose CPEs
are hardware-only before building the query. Affected products without
any CPEs are left alone, so name-based matching still works in those
cases.

Also adds `prefetch_related("cpes")` while filtering, so this does not
introduce N+1 queries.

Tests added for:
- hardware CPEs not producing a match
- application CPEs still producing a match
- mixed inputs only skipping the hardware-only affected products

All 14 tests pass.

### DIFF
--- a/src/shared/tests/test_linkage.py
+++ b/src/shared/tests/test_linkage.py
@@ -145,7 +145,6 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
     assert proposal.derivations.count() == 0
 
 
-@pytest.mark.xfail(reason="Not implemented")
 def test_hardware_cpe_produces_no_match(
     make_container: Callable[..., Container],
     make_drv: Callable[..., NixDerivation],
@@ -174,7 +173,6 @@ def test_application_cpe_produces_match(
     assert build_new_links(container)
 
 
-@pytest.mark.xfail(reason="Not implemented")
 def test_mixed_cpe_parts_skips_hardware_only_affected_products(
     make_container: Callable[..., Container],
     make_drv: Callable[..., NixDerivation],


src/shared/listeners/automatic_linkage.py
```diff
@@ -16,6 +16,19 @@
 logger = logging.getLogger(__name__)
 
 
+def get_cpe_part(cpe_name: str) -> str | None:
```
fricklerhandwerk: The change has so much merit on its own that I'll let that hack pass, but it must be made explicit that this is an atrocity. 

```suggestion
# FIXME(@fricklerhandwerk): This only works because we're validating syntax on ingestion.
# Use a proper parsing library such as https://github.com/nilp0inter/cpe to work on structured data.
# That particular one looks like the best candidate, but appears unmaintained (or could just be very stable); needs thorough review before adopting it.
def get_cpe_part(cpe_name: str) -> str | None:
```

src/shared/tests/test_linkage.py
```diff
@@ -143,3 +143,56 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
...
+        cve_id="CVE-2025-0003",
+        package_name="myapp",
+        product="myapp",
+        cpes=["cpe:2.3:a:vendor:myapp:1.0:*:*:*:*:*:*:*"],
+    )
+    make_drv(pname="some_router", version="1.0")
+    make_drv(pname="myapp", version="1.0", attribute="myapp")
+
+    assert not build_new_links(hw_container)
+    assert build_new_links(app_container)
```
fricklerhandwerk, line 198: Since we're at it, also check that we're matching the right derivations.

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: > The change has so much merit on its own that I'll let that hack pass, but it must be made explicit that this is an atrocity.

Done, added the suggested FIXME comment

README.md
```diff
@@ -17,4 +17,4 @@ By discussing issues with security team members and other maintainers, they can
 
 ## Contributing
 
-Please see the [**Contributing Guide**](CONTRIBUTING.md) for more information on how to get started. 
+Please see the [**Contributing Guide**](CONTRIBUTING.md) for more information on how to get started.
```
fricklerhandwerk, line 20: Sorry, I had merged a fluke into main which is now fixed, please rebase.

ArkhamKnight25: Done

src/shared/listeners/automatic_linkage.py
```diff
@@ -33,16 +49,16 @@ def produce_linkage_candidates(
...
-        .distinct()
-    )
-    products = (
-        container.affected.exclude(product__isnull=True)
-        .values_list("product", flat=True)
-        .distinct()
+    affected_products = [
+        a
+        for a in container.affected.prefetch_related("cpes").all()
+        if not should_skip_affected_product(a)
```
fricklerhandwerk: This tight loop containing a query hidden in a procedure seems wrong, this will be very slow. That's the main thing that makes me very uneasy about the overall change.

Have you considered prefetching all of that very early on, such as at the call site of `build_new_links`?

src/shared/listeners/automatic_linkage.py
```diff
@@ -16,6 +16,22 @@
...
+    parts = cpe_name.split(":")
+    if len(parts) < 3 or parts[0] != "cpe" or parts[1] != "2.3":
+        return None
+    part = parts[2]
+    return part if part in ("a", "h", "o") else None
+
+
+def should_skip_affected_product(affected: AffectedProduct) -> bool:
+    cpes = list(affected.cpes.values_list("name", flat=True))
+    return bool(cpes) and all(get_cpe_part(c) == "h" for c in cpes)
```
fricklerhandwerk:
```suggestion
    return cpes != [] and all(get_cpe_part(c) == "h" for c in cpes)
```

The `not x` pattern for structured data that has falseness when empty is in itself contentious among Python people AFAIK. But casting to bool feels like mocking that ambiguous semantics and reads very weird, let's not do that.

src/shared/listeners/automatic_linkage.py
```diff
@@ -16,6 +16,22 @@
...
+def get_cpe_part(cpe_name: str) -> str | None:
+    parts = cpe_name.split(":")
+    if len(parts) < 3 or parts[0] != "cpe" or parts[1] != "2.3":
+        return None
+    part = parts[2]
+    return part if part in ("a", "h", "o") else None
+
+
+def should_skip_affected_product(affected: AffectedProduct) -> bool:
+    cpes = list(affected.cpes.values_list("name", flat=True))
```
fricklerhandwerk:
```suggestion
    cpes = affected.cpes.values_list("name", flat=True)
```


flat=True already makes it a list

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: Replaced with an ORM-level filter using Exists + OuterRef subqueries, no Python loop, no hidden queries. The filtering now happens in a single SQL WHERE clause.

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: the filtering is now done at the ORM level, so this is no longer applicable.

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: this is also no longer needed

src/shared/listeners/automatic_linkage.py
```diff
@@ -33,13 +54,24 @@ def produce_linkage_candidates(
         .filter(row_num=1)
     )
 
+    # Exclude hardware-only affected products: those with at least one CPE
+    # where all CPEs have part=h (cpe:2.3:h:...). Products with no CPEs are
+    # kept so name-based matching still works for entries without CPE data.
+    has_any_cpe = Exists(Cpe.objects.filter(affectedproduct=OuterRef("pk")))
+    has_non_hardware_cpe = Exists(
+        Cpe.objects.filter(affectedproduct=OuterRef("pk")).exclude(
+            name__startswith="cpe:2.3:h:"
```
fricklerhandwerk: The DB field accepts case insensitive strings. I don't have the CPE spec at hand to check whether we'd run into not matching upper case variants with this approach, but to be sure we could simply start storing them lowercase from now on.

fricklerhandwerk (commented):
Yes and now you don't need the original CPE part helper, and, most importantly, can enable the associated tests to prove this is indeed working.

A good follow-up PR would be to auto-reject and add an appropriate rejection reason if there are *only* hardware CPE packages in the CVE.

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: Switched to `istartswith `to handle this. For storing lowercase on ingestion, that sounds like a good separate cleanup, happy to do that as a follow-up if you want to file an issue.

src/shared/listeners/automatic_linkage.py
```diff
@@ -33,13 +43,27 @@ def produce_linkage_candidates(
         .filter(row_num=1)
     )
 
+    # Exclude hardware-only affected products: those with at least one CPE
+    # where all CPEs have part=h (cpe:2.3:h:...). Products with no CPEs are
+    # kept so name-based matching still works for entries without CPE data.
+    # FIXME(@fricklerhandwerk): This only works because we're validating syntax on ingestion.
+    # Use a proper parsing library such as https://github.com/nilp0inter/cpe to work on structured data.
```
fricklerhandwerk:
```suggestion
    # FIXME(@fricklerhandwerk): se a proper parsing library such as https://github.com/nilp0inter/cpe to work on structured data.
```

The code is self-explanatory, naming tells the story.

PR #895, opened 2026-03-20
test: add xfail tests for hardware CPE filtering

ArkhamKnight25:
Part of #391

Splitting out the tests from #894 as requested in review. 
`test_hardware_cpe_produces_no_match` — `part=h` CPEs should never match Nixpkgs derivations. Marked `xfail`.

`test_application_cpe_produces_match` — `part=a` CPEs should still match normally. No `xfail` since this already works.

`test_mixed_cpe_parts_skips_hardware_only_affected_products` — only application affected products should generate candidates when mixed with hardware-only ones. Marked `xfail`. Also checks the matched derivation is the correct one per review feedback.

Also extends `make_container` to accept a `cpes` list.

### DIFF
--- a/src/shared/tests/test_linkage.py
+++ b/src/shared/tests/test_linkage.py
@@ -145,7 +145,7 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
     assert proposal.derivations.count() == 0
 
 
-@pytest.mark.xfail(reason="Not implemented: hardware CPE filtering (#391)")
+@pytest.mark.xfail(reason="Not implemented")
 def test_hardware_cpe_produces_no_match(
     make_container: Callable[..., Container],
     make_drv: Callable[..., NixDerivation],
@@ -174,7 +174,7 @@ def test_application_cpe_produces_match(
     assert build_new_links(container)
 
 
-@pytest.mark.xfail(reason="Not implemented: hardware CPE filtering (#391)")
+@pytest.mark.xfail(reason="Not implemented")
 def test_mixed_cpe_parts_skips_hardware_only_affected_products(
     make_container: Callable[..., Container],
     make_drv: Callable[..., NixDerivation],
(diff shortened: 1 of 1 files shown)


src/shared/tests/test_linkage.py
```diff
@@ -143,3 +143,60 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
         == CVEDerivationClusterProposal.RejectionReason.EXCLUSIVELY_HOSTED_SERVICE
     )
     assert proposal.derivations.count() == 0
+
+
+@pytest.mark.xfail(reason="Not implemented: hardware CPE filtering (#391)")
```
fricklerhandwerk:
```suggestion
@pytest.mark.xfail(reason="Not implemented")
```

src/shared/tests/test_linkage.py
```diff
@@ -143,3 +143,60 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
...
+        package_name="myapp",
+        product="myapp",
+        cpes=["cpe:2.3:a:vendor:myapp:1.0:*:*:*:*:*:*:*"],
+    )
+    make_drv(pname="myapp")
+
+    assert build_new_links(container)
+
+
+@pytest.mark.xfail(reason="Not implemented: hardware CPE filtering (#391)")
```
fricklerhandwerk:
```suggestion
@pytest.mark.xfail(reason="Not implemented")
```

src/shared/tests/test_linkage.py
```diff
@@ -143,3 +143,60 @@ def test_exclusively_hosted_service_creates_rejected_proposal(
...
+        product="myapp",
+        cpes=["cpe:2.3:a:vendor:myapp:1.0:*:*:*:*:*:*:*"],
+    )
+    make_drv(pname="some_router", version="1.0")
+    make_drv(pname="myapp", version="1.0", attribute="myapp")
+
+    assert not build_new_links(hw_container)
+    assert build_new_links(app_container)
+    suggestion = CVEDerivationClusterProposal.objects.get(cve=app_container.cve)
+    assert suggestion.derivations.filter(name__startswith="myapp").exists()
```
fricklerhandwerk, line 202:
```suggestion
    assert suggestion.derivations.filter(name__startswith="myapp").exists()
    assert not suggestion.derivations.filter(name__startswith="some_router").exists()
```

PR #898, opened 2026-03-21
feat(tags): add unique constraint to value and implement __str__

DarshanCode2005:
This PR improves data integrity and observability for the `Tag` model.
 Changes:
`src/shared/models/cve.py` :
* Added unique=True to the Tag.value field to prevent duplicate data.
* Implemented `str` to return the tag's value, improving the representation in logs.

Rationale: Duplicate tags ("critical", "critical" etc.) were previously possible in the database, leading to inconsistent data. This change enforces a clean tag registry. The string representation also significantly improves the developer & admin experience by making objects identifiable by their actual text.


src/shared/models/cve.py
```diff
@@ -116,7 +116,10 @@ class Meta:  # type: ignore[override]
 class Tag(models.Model):
     """Class representing a tag related to a CVE record."""
 
-    value = models.CharField(max_length=128)
+    value = models.CharField(max_length=128, unique=True)
+
+    def __str__(self) -> str:
```
fricklerhandwerk, line 121: We can have this a separate PR if it also simplifies a call site that would otherwise have done it inline. But given my main comment this may be moot.

src/shared/models/cve.py
```diff
@@ -118,6 +118,9 @@ class Tag(models.Model):
 
     value = models.CharField(max_length=128)
 
+    def __str__(self) -> str:
```
fricklerhandwerk, line 121: Now, were we doing that by hand anywhere in the code? If so, please use this new thing there, if not, we can close the PR. :)

fricklerhandwerk:
I'll close this since the improvement of using `__str__` is marginal and would actually add more lines of code. Feel free to reopen if you think otherwise.

PR #899, opened 2026-03-21
docs: document different components of the project

DarshanCode2005:
As discussed with the maintainer, he suggested implementing this section.

Thought Process: New contributors often find it difficult to understand how the various services (Daphne, Workers, Cron) are defined and where the core business logic resides. This documentation provides the necessary "mental map" to navigate the project effectively.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -13,10 +13,6 @@ It contains general contribution information, and lists resources to help you ge
 
 Other directories in this repository have additional `README.md` files with more specific information relevant to their sibling files.
 
-## System Architecture and Project Structure
-
-The security tracker is a distributed system where Nix manages the infrastructure and services, while Django handles the application logic and web interface.
-
 Service definitions are in [`nix/configuration.nix`](nix/configuration.nix).
 
 Application logic lives in the [`src/`](src/) directory.


CONTRIBUTING.md
```diff
@@ -13,6 +13,27 @@ It contains general contribution information, and lists resources to help you ge
 
 Other directories in this repository have additional `README.md` files with more specific information relevant to their sibling files.
 
+## System Architecture and Project Structure
+
+The security tracker is a distributed system where Nix manages the infrastructure and services, while Django handles the application logic and web interface.
+
```
fricklerhandwerk:
```suggestion
```

We already say this here:

https://github.com/NixOS/nix-security-tracker/blob/0f2eb73dd2fd800b8539432064aca5f4e6127167/CONTRIBUTING.md#L18-L19

CONTRIBUTING.md
```diff
@@ -13,6 +13,27 @@ It contains general contribution information, and lists resources to help you ge
...
+- **Background Worker (`nix-security-tracker-worker`)**: A generic job processor that listens for database events (via `pgpubsub`) to handle asynchronous tasks like cache regeneration or notifications.
+- **Cron Jobs**: Scheduled tasks for fetching channel information and CVE deltas.
+
+### Codebase Organization
+
+The core business logic and web application reside in the [`src/`](src/) directory. From here, it follows standard Django patterns:
+
+- [`src/shared/`](src/shared/): Contains the core logic shared across all services, including database models, migrations, background listeners, and common utility functions.
+- [`src/webview/`](src/webview/): Contains the web interface implementation, including views, templates, and static assets (CSS/JS).
+- [`src/project/`](src/project/): Contains the Django project configuration (`settings.py`, `urls.py`, `asgi.py`).
```
fricklerhandwerk:
```suggestion
Service definitions are in [`nix/configuration.nix`](nix/configuration.nix).

Application logic lives in the [`src/`](src/) directory.
From here, it follows standard Django patterns:
- [`src/project/`](src/project/): global project configuration
- [`src/shared/`](src/shared/): [application](https://docs.djangoproject.com/en/6.0/ref/applications/) with data models and business logic
- [`src/webview/`](src/webview/): application for the web frontend
```

The details will change with the code, no point in having another place to check or go out of date when it happens. If there was anything non-obvious in that Nix code (or the application directories), I'd suggest adding comments in place (or the READMEs mentioned above in the file) that would summarize in very general terms what's happening, or why it is the way it is, but I don't think that's the case (if you know the components, but what those are can be looked up.)

fricklerhandwerk (approved):

PR #901, opened 2026-03-21
fix: atomic evaluation concurrency gate to prevent race condition

WHOIM1205:
### Summary
Fixes a critical race condition and off-by-one error in the evaluation concurrency limiter that could allow more than the configured number of `nix-eval-jobs` to run in parallel.

### Problem
- Off-by-one bug: `>` used instead of `>=`, allowing extra concurrent evaluations  
- TOCTOU race: non-atomic `count()` check and state update allowed multiple workers to pass simultaneously  

This could lead to unbounded parallel evaluations, high memory usage, OOM kills, and failed jobs.

### Fix
- Replaced check-then-update with atomic transaction using `select_for_update()`  
- Corrected condition to `>=`  
- Ensures safe slot acquisition across concurrent workers  

### Impact
- Enforces strict concurrency limits  
- Prevents OOM crashes  
- Improves evaluation reliability  
- Eliminates silent over-allocation  

### Reference
Detailed bug: :contentReference[oaicite:0]{index=0}

### DIFF
--- a/src/shared/listeners/nix_evaluation.py
+++ b/src/shared/listeners/nix_evaluation.py
@@ -190,13 +190,10 @@ async def evaluation_entrypoint(
             )
             return True
 
-    acquired = False
-    while not acquired:
-        acquired = await sync_to_async(_try_acquire_slot)()
-        if not acquired:
-            # Add in average 30s as a jitter to enable clear winners during the grab for the evaluation slot.
-            jitter = random.randint(1, 60)
-            await asyncio.sleep(avg_eval_time + jitter)
+    while not (await sync_to_async(_try_acquire_slot)()):
+        # Add in average 30s as a jitter to enable clear winners during the grab for the evaluation slot.
+        jitter = random.randint(1, 60)
+        await asyncio.sleep(avg_eval_time + jitter)
     repo = GitRepo(settings.LOCAL_NIXPKGS_CHECKOUT)
     start = time.time()
     try:


src/shared/listeners/nix_evaluation.py
```diff
@@ -172,20 +173,30 @@ class TooManyStuff(BaseException):
...
+            )
+            return True
+
+    acquired = False
+    while not acquired:
+        acquired = await sync_to_async(_try_acquire_slot)()
+        if not acquired:
+            # Add in average 30s as a jitter to enable clear winners during the grab for the evaluation slot.
+            jitter = random.randint(1, 60)
+            await asyncio.sleep(avg_eval_time + jitter)
```
```suggestion
    while not (await sync_to_async(_try_acquire_slot)()):
        jitter = random.randint(1, 60)
        await asyncio.sleep(avg_eval_time + jitter)
```

Why not as compact as before?

src/shared/listeners/nix_evaluation.py
```diff
@@ -172,20 +173,30 @@ class TooManyStuff(BaseException):
...
-        jitter = random.randint(1, 60)
-        await asyncio.sleep(avg_eval_time + jitter)
-    # Atomically update the state to prevent anyone going over the
-    # specified concurrency.
-    await NixEvaluation.objects.filter(id=evaluation.pk).aupdate(
-        state=NixEvaluation.EvaluationState.IN_PROGRESS,
-        updated_at=timezone.now(),
-    )
+    # Acquire an evaluation slot atomically: lock all IN_PROGRESS rows
+    # to prevent concurrent workers from racing past the count check.
```
```suggestion
```

No need to repeat what the code says very clearly.

PR #903, opened 2026-03-21
perf: stop evaluation ingestion immediately on EOF

WHOIM1205:
### Summary
Fixes a critical busy-loop issue where `drain_lines` would spin ~75,000 iterations after `nix-eval-jobs` closes stdout. This caused unnecessary CPU usage on every evaluation completion.

### Root Cause
`stream.readline()` returns `b""` at true EOF, but this condition was not handled immediately. Since no timeout occurred, the loop kept appending empty lines until the batch window limit triggered. :contentReference[oaicite:0]{index=0}

### Fix
- Detect `b""` (true EOF) immediately after `readline()`
- Break the loop early to avoid unnecessary iterations
- Yield any remaining buffered lines before exiting

### Impact
- Eliminates CPU waste due to tight loop at EOF
- Ensures efficient stream draining for every evaluation
- No change to existing timeout/batch behavior

### Notes
This fix preserves existing logic while addressing a high-frequency performance issue in the core evaluation flow.

### DIFF
--- a/src/shared/listeners/nix_evaluation.py
+++ b/src/shared/listeners/nix_evaluation.py
@@ -140,10 +140,8 @@ async def drain_lines(
             async with asyncio.timeout(timeout) as cm:
                 line = await stream.readline()
                 if line == b"":
-                    # True EOF: the stream is closed.  readline() returns
-                    # b"" (no trailing newline) only at end-of-stream.
-                    # Without this check the loop would spin ~75,000 times
-                    # appending b"" until max_batch_window triggers.
+                    # `readline()` returns `b""` only at end-of-stream:
+                    # https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readline
                     eof = True
                     break
                 lines.append(line)


src/shared/listeners/nix_evaluation.py
```diff
@@ -138,7 +138,15 @@ class TooManyStuff(BaseException):
     while not eof:
         try:
             async with asyncio.timeout(timeout) as cm:
-                lines.append(await stream.readline())
+                line = await stream.readline()
+                if line == b"":
+                    # True EOF: the stream is closed.  readline() returns
+                    # b"" (no trailing newline) only at end-of-stream.
+                    # Without this check the loop would spin ~75,000 times
+                    # appending b"" until max_batch_window triggers.
```
```suggestion
```

Comment is almost entirely redundant, the code already obviously says "this means EOF". It may be nonobvious when not knowing readline() behavior by heart, but there one line or a link to the respective part of documentation would be enough:


```suggestion
                    # `readline()` returns `b""` only at end-of-stream:
                    # https://docs.python.org/3/library/asyncio-stream.html#asyncio.StreamReader.readline
```

PR #904, opened 2026-03-21
fix: handle missing user in GitHub webhook handler

WHOIM1205:
Description

This PR addresses an existing FIXME by preventing webhook handler crashes caused by missing users and headers in GitHub webhook processing.

Changes:

• Wrapped User.objects.get(...) in a try/except block to handle User.DoesNotExist and avoid 500 errors when the GitHub user has not logged in yet.

• Added graceful logging and early return for unknown users instead of crashing the webhook handler.

• Added try/except for missing HTTP_X_GITHUB_EVENT header to prevent KeyError and return a 403 response when absent.

These changes improve webhook reliability and prevent repeated failures due to GitHub retries on 500 responses.


src/shared/auth/github_webhook.py
```diff
@@ -60,7 +60,10 @@ def handle_github_hook(request: HttpRequest) -> HttpResponse:
     else:
         payload = json.loads(request.body)
 
-    event = request.META["HTTP_X_GITHUB_EVENT"]
+    try:
+        event = request.META["HTTP_X_GITHUB_EVENT"]
+    except KeyError:
```
fricklerhandwerk: How likely is that though? Can't we expect GitHub to follow their own spec? And if they don't we probably want a crash here anyway that will show up in our notifications, because that would be an event worth investigating.

src/shared/auth/github_state.py
```diff
@@ -99,9 +99,18 @@ def sync_team_membership_from_webhook(
...
 
-        user: User = User.objects.get(
-            socialaccount__provider="github", socialaccount__uid=str(github_user_id)
-        )
+        try:
+            user: User = User.objects.get(
+                socialaccount__provider="github",
+                socialaccount__uid=str(github_user_id),
+            )
+        except User.DoesNotExist:
```
fricklerhandwerk: I don't have the entire code at the top of my mind right now, but can this even happen? Won't the GitHub sync at startup obtain these users, or do they really only show up on login?

src/shared/auth/github_webhook.py
WHOIM1205: It is unlikely but still possible in practice and if it happens we end up returning a 500 which GitHub will retry repeatedly turning it into a noisy failure loop instead of a single signal
Since this is an external webhook I think it is safer to fail gracefully and log it rather than rely on GitHub always behaving perfectly

src/shared/auth/github_state.py
WHOIM1205: Yes this can happen because users are only created after login so if someone is added to a team before ever logging in we will not have a corresponding user yet
In that case the current code raises DoesNotExist and returns a 500 which again triggers retries so handling it gracefully keeps the webhook stable while still logging the situation

src/shared/auth/github_webhook.py
fricklerhandwerk: Will GitHub retry though? Do you have proof?

The thing with upstream following their own spec is: if they don't and we just log a warning, nobody will even notice it happened. But maintainers will get an email on 500. Knowing we failed at implementing the spec and are losing information is worth deleting a dozen duplicate emails. Knowing upstream failed at implementing their spec is hilarious enough too.

src/shared/auth/github_webhook.py
```diff
@@ -60,7 +60,10 @@ def handle_github_hook(request: HttpRequest) -> HttpResponse:
...
+        return HttpResponseForbidden("Missing event header")
```
fricklerhandwerk:
```suggestion
```

In any case this is a decision to be made, and would be addressed in a separate PR. A commit title that contains "and" connecting two independent ideas is a strong smell!

fricklerhandwerk:
Oh no, you did catch genuine bug there, because surely a team could update a real user that just happens not to be in our database and then we'd get a crash. Your change was correct and one of the tests needs changing.

PR #906, opened 2026-03-21
fix: store unedited packages as original_packages in suggestion cache

WHOIM1205:
### Summary
Fixes an issue where `original_packages` was incorrectly set to the edited `packages` instead of the actual unmodified package data.

### Problem
The value computed from `channel_structure(...)` was never stored. Both `original_packages` and `packages` ended up referencing the edited data, breaking:
- Undo functionality
- Original vs edited comparisons in UI
- Maintainer categorization logic

### Fix
- Assign `original_packages` correctly to the unmodified data
- Remove outdated FIXME comment

### Changes
```diff
- original_packages=packages
+ original_packages=original_packages

### DIFF
--- a/src/webview/tests/test_packages.py
+++ b/src/webview/tests/test_packages.py
@@ -93,3 +93,134 @@ def test_ignore_restore_package(
     expect(active_packages.get_by_text("package1")).to_be_visible()
     expect(active_packages.get_by_text("package2")).to_be_visible()
     expect(ignored_packages).not_to_be_visible()
+
+
+def test_ignore_multiple_packages(
+    live_server: LiveServer,
+    as_staff: Page,
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    make_drv: Callable[..., NixDerivation],
+) -> None:
+    """Ignoring packages one by one keeps all of them in the ignored section."""
+    drv1 = make_drv(pname="alpha")
+    drv2 = make_drv(pname="bravo")
+    drv3 = make_drv(pname="charlie")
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/shared/tests/test_suggestion_caching.py [hunk 1]
```diff
@@ -70,3 +73,139 @@ def test_caching_newest_package(
     )
 
     assert package["major_version"] == "2.0"
+
+
+def test_original_packages_preserved_after_edit(
```
fricklerhandwerk: This tests the right things but pins down implementation details that will change, but when they do it shouldn't require changing any tests. Please move them to `webview` and follow the playwright pattern.

src/webview/tests/test_packages.py
```diff
@@ -93,3 +93,132 @@ def test_ignore_restore_package(
...
+    # Ignore alpha
+    active.locator(".package-alpha").get_by_role("button", name="Ignore").click()
+    container.get_by_text(re.compile("Ignored packages")).click()
+    expect(ignored.get_by_text("alpha")).to_be_visible()
+    expect(active.get_by_text("bravo")).to_be_visible()
+    expect(active.get_by_text("charlie")).to_be_visible()
+
+    # Ignore bravo
+    active.locator(".package-bravo").get_by_role("button", name="Ignore").click()
+    expect(ignored.get_by_text("alpha")).to_be_visible()
```
fricklerhandwerk, line 135: This can't work due to https://github.com/NixOS/nix-security-tracker/issues/913, you have to expand the ignored section first. Please only push commits that will pass all tests.

PR #908, opened 2026-03-22
fix: kill nix-eval-jobs subprocess on evaluation failure

WHOIM1205:
## Description

This PR fixes a critical subprocess lifecycle bug where `nix-eval-jobs` processes were left running as orphans whenever an exception occurred during evaluation ingestion.

Previously, failures inside the ingestion loop (e.g. DB errors, integrity issues, timeouts) would mark the evaluation as `CRASHED` but never terminate the spawned subprocess, leading to severe resource leaks and inconsistent system state.

## Changes

- Updated `evaluation_entrypoint()` in `nix_evaluation.py` to properly manage subprocess lifecycle
  - Wrapped subprocess execution in a `try/finally` block
  - Ensures `eval_process.kill()` and `await eval_process.wait()` are always called on failure paths
- Added safety check to only kill the process if it is still running (`returncode is None`)

## Impact

- Prevents orphaned `nix-eval-jobs` processes consuming CPU and memory
- Eliminates cascading failures caused by resource exhaustion under concurrent workloads
- Ensures evaluation slot tracking remains accurate (DB state matches actual running processes)
- Guarantees safe cleanup of worktrees and temporary directories
- No behavior change on successful execution path

Fixes a critical ingestion reliability issue and improves overall system stability under failure conditions.


src/shared/tests/test_nix_evaluation.py
```diff
@@ -0,0 +1,100 @@
...
+def _make_mock_process(*, returncode=None):
+    proc = AsyncMock()
+    proc.stdout = AsyncMock(spec=asyncio.StreamReader)
+    proc.returncode = returncode
+    proc.kill = MagicMock()
+    proc.wait = AsyncMock(return_value=0)
+    return proc
+
+
+def _setup_mocks():
```
fricklerhandwerk: All these helper procedures should be proper fixtures. This is both more readable (idiomatic, shows up in test arguments as dependencies) and easier to work with (naturally composes). For instance, returning a tuple is cumbersome to extend in the future. With fixtures we'd just add another fixture.

PR #909, opened 2026-03-22
fix: crash signal detection for `nix-eval-jobs`

WHOIM1205:
## Description

this fixes an issue where crashes from nix-eval-jobs were not being detected properly and were getting treated like normal failures

## What was wrong

the signal handling was off in two ways:

- the values used for SIGSEGV and SIGABRT were incorrect (they were shell exit codes, not actual signals)
- asyncio returns negative signal numbers (like -11 for SIGSEGV), so the check never actually matched real crashes

because of this, segfaults were silently ignored and no proper failure reason was recorded

## Changes

- replaced the hardcoded signal values with a proper set of crash signals
- added handling for both:
  - negative signal return codes (asyncio)
  - 128+signal style exit codes
- updated the check to use this new set

## Result

crashes are now correctly detected and won't be misclassified anymore


src/shared/tests/test_crash_signals.py
```diff
@@ -0,0 +1,20 @@
+import signal
+
+from shared.listeners.nix_evaluation import CRASH_SIGNALS
+
+
+def test_crash_signals_detect_asyncio_returncodes():
```
fricklerhandwerk: What should be tested instead is what happens if the evaluation process gets SIGSEGV/SIGABRT. Expected outcome is it would return 137/134. This would reveal that the change was unnecessary in the first place, but the test would be very valuable regardless as we'd have a scaffold for testing other failure modes.

src/shared/tests/test_crash_signals.py
```diff
@@ -0,0 +1,87 @@
...
+        patch("shared.listeners.nix_evaluation.aiofiles.open", new_callable=MagicMock),
+    ):
+        from shared.listeners.nix_evaluation import evaluation_entrypoint
+
+        asyncio.run(evaluation_entrypoint(0.0, evaluation))
+
+
+@pytest.mark.django_db
+def test_sigsegv_asyncio_marks_crashed(evaluation_for_crash_test):
+    _run_entrypoint(evaluation_for_crash_test, -signal.SIGSEGV)
```
fricklerhandwerk: Yes and now it should become evident that you(r LLM) just made up that return code. This is not what happens in reality. You must actually run `nix-eval-jobs` on something and send it one of those signals, and then observe that the return codes will be 137/134.

src/shared/listeners/nix_evaluation.py
```diff
@@ -25,8 +26,13 @@
...
 
-SIGSEGV = 137
-SIGABRT = 134
+CRASH_SIGNALS = (
+    -signal.SIGSEGV,
+    -signal.SIGABRT,
+)
+# asyncio.create_subprocess_exec returns negative signal numbers on Unix
+# (e.g. -11 for SIGSEGV, -6 for SIGABRT). The 128+signal convention only
+# applies when a shell wraps the process, which is not the case here.
```
```suggestion
# When terminated  by signal, `asyncio.create_subprocess_exec()` returns the negative signal number on Unix.
# https://docs.python.org/3/library/asyncio-subprocess.html#asyncio.subprocess.Process.returncode
CRASH_SIGNALS = (
    -signal.SIGSEGV,
    -signal.SIGABRT,
)
```

PR #923, opened 2026-03-24
feat: command for cleaning up old untriaged CVE matches

### LINKED ISSUE #922: Management command for deleting matches older than 1 year
As a security team member, I don't care about CVEs older than a year. But I would like to eventually arrive at an empty queue.

## Acceptance criteria

Given there are a lot of old matches,
When I run the new management command,
The matches older than 1 year will be gone and leave no data behind (i.e. also clean up the related fields).

DarshanCode2005:
Resolves #922 

This PR adds a new Django management command delete_old_matches to clean up outdated CVE-package matches and keep the queue manageable.
Matches older than one year are removed based on date_published, with a fallback to date_reserved for CVEs that were never published. Deletions are performed in batches (configurable via `--batch-size`) to avoid heavy database operations, and cascading ensures all related data (cached suggestions , edits, links etcetera.) is cleaned up automatically. A `--dry-run` option is included to preview deletions safely.

Comprehensive tests are there to verify deletion logic, fallback behavior, batching, cascading cleanup and dry-run safety.


src/shared/tests/test_delete_old_matches.py
```diff
@@ -0,0 +1,256 @@
...
+
+    container = make_container()
+    _set_cve_dates(
+        container.cve,
+        date_published=timezone.now() - timedelta(days=400),
+    )
+    suggestion = make_suggestion(container=container)
+
+    maintainer: NixMaintainer = make_maintainer()  # type: ignore[assignment]
+    MaintainersEdit.objects.create(
```
fricklerhandwerk, line 172: Ah, good catch on the gap in specification. We absolutely want to preserve everything that has data entered by humans attached to it, so actually we want to delete only untriaged suggestions. Those can be expected not to have seen any user interaction.

src/shared/tests/test_delete_old_matches.py
```diff
@@ -0,0 +1,256 @@
...
+        CveRecord.objects.filter(pk=cve.pk).update(**updates)
+
+
+# ---------------------------------------------------------------------------
+# Core deletion behaviour
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.django_db
+def test_deletes_old_match(
```
fricklerhandwerk: The tests should primarily check that all user data and derivations/CVEs stay intact. We want to garbage-collect them separately. I imagine this should carefully trace through all relations and demonstrate they're preserved. Likely there's a clever way of changing the relation directions and `on_delete` directives to make this elegant. Just make sure it's reasonably incremental (40M derivations in production, and growing...).

src/shared/management/commands/delete_old_matches.py
```diff
@@ -0,0 +1,84 @@
...
+
+        # We process in batches to avoid fetching all PKs into memory at once.
+        # Since we are deleting records, the records matching the filter will shrink,
+        # so we always just fetch the first batch_size records until none are left.
+        # This keeps the execution strictly incremental, scaling perfectly for large tables.
+        deleted_total = 0
+        batch_num = 1
+
+        while True:
+            batch_pks = list(old_proposals.values_list("pk", flat=True)[:batch_size])
```
fricklerhandwerk, line 66: Why not `iterator()`?

src/shared/tests/test_delete_old_matches.py
```diff
@@ -0,0 +1,295 @@
...
+        updates["date_published"] = date_published
+    if date_reserved is not None:
+        updates["date_reserved"] = date_reserved
+    if updates:
+        CveRecord.objects.filter(pk=cve.pk).update(**updates)
+
+
+# ---------------------------------------------------------------------------
+# Core deletion behaviour
+# ---------------------------------------------------------------------------
```
fricklerhandwerk:
```suggestion
```

This sort of comment tends not to age well, I recommend not having those. Includes those inline comments that read a bit like headings. You already have very readable naming and group code blocks that belong together, there's really not more one can do to maintain readability.

src/shared/tests/test_delete_old_matches.py
```diff
@@ -0,0 +1,295 @@
...
+    date_reserved: object = None,
+) -> None:
+    """Helper to set date fields without triggering model-level auto_now logic."""
+    updates: dict[str, object] = {}
+    if date_published is not None:
+        updates["date_published"] = date_published
+    if date_reserved is not None:
+        updates["date_reserved"] = date_reserved
+    if updates:
+        CveRecord.objects.filter(pk=cve.pk).update(**updates)
```
fricklerhandwerk: This be done inline as:

```
CveRecord.objects.filter(pk=cve.pk).update(date_published=...)
```

When you see this a couple of times you get used to this being the pattern to avoid triggering `auto_now`.

src/shared/tests/test_delete_old_matches.py
DarshanCode2005: Okay Sir! I will remove those.

src/shared/management/commands/delete_old_matches.py
DarshanCode2005: Good point. I avoided `iterator()` here because we need to delete in bounded batches. Using slicing with [:batch_size] ensures each iteration operates on a small, fixed set of PKs , and since deleted rows drop out of queryset, it naturally progresses without holding a long-lived cursor. `iterator()` would stream all matching rows, but we'd still need to manually chunk and it could hold the query open longer, which isn't ideal for large-scale deletions. 

Happy to switch if you think `iterator(chunk_size=..)` fits better here.

src/shared/tests/test_delete_old_matches.py
DarshanCode2005: Okay! I'll implement it.

src/shared/management/commands/delete_old_matches.py
fricklerhandwerk: Iterator will also give us controlled chunks, so there's no memory issue. But what's the problem with the cursor?

DarshanCode2005: So if we iterate with iterator() and delete during iteration, the cursor can still return already-deleted rows. That makes it inefficient.

With slicing ([:batch_size]), each batch is a fresh query on the current state of the table. Deleted rows no longer appear, and each batch runs as a short, independent transaction.

Ref: https://devcenter.heroku.com/articles/postgresql-concurrency [in the disadvantages of MVCC]

src/shared/tests/test_delete_old_matches.py
```diff
@@ -0,0 +1,246 @@
...
+) -> None:
+    """A PENDING proposal is deleted, but CveRecord, NixDerivation, and Container are preserved."""
+    # Use the injected `cve` fixture and modify its date
+    CveRecord.objects.filter(pk=cve.cve.pk).update(
+        date_published=timezone.now() - timedelta(days=366)
+    )
+    # create the proposal using the injected fixtures
+    make_suggestion(container=cve, drvs={drv: ProvenanceFlags.PACKAGE_NAME_MATCH})
+
+    assert CVEDerivationClusterProposal.objects.count() == 1
```
fricklerhandwerk, line 35: Aha, there could also be notifications around. What should we do about those?

DarshanCode2005: We can make notifications follow  the proposal lifecycle:
* Change `SuggestionNotification.suggestion` to `on_delete=CASCADE` so notifications tied to deleted (untriaged) proposals are cleaned up automatically.
* Add a `post_delete` signal on `Notification` to keep `unread_notifications_count` consistent.
* Keep core/important models protected as-is.
* Add tests to verify notifications are deleted with proposals and counters remain correct.

fricklerhandwerk: SGTM!

src/webview/models.py
```diff
@@ -61,7 +61,7 @@ class SuggestionNotification(Notification):
         CVEDerivationClusterProposal,
         # XXX(@fricklerhandwerk): It's unlikely we'll delete suggestions any time soon.
         # But when we do, we don't want to mess up the notification counter. [ref:count-notifications]
```
fricklerhandwerk:
```suggestion
```

Fixed!

PR #930, opened 2026-03-25
fix: deduplicate CVE references in suggestions context

### LINKED ISSUE #876: Deduplicate CVE references
It can happen that for some reason a CVE has many duplicate references attached to it. Deduplicate them in the view, such that #775 (ignoring items) will work correctly.

Image

devprashant19:
## Description

This PR resolves issue #876 where CVE references derived from multiple containers (like CNA and ADP) were blindly appended to the UI context, causing identical reference links to be displayed multiple times with identical or split tags.

## Changes
- Modified `SuggestionContext.update_references` inside [src/webview/suggestions/context/types.py](cci:7://file:///media/prashant1901/New%20Volume/GSOC/nix-security-tracker/src/webview/suggestions/context/types.py:0:0-0:0) to efficiently aggregate references using a URL-keyed dictionary.
- Seamlessly merges [Tag](cci:2://file:///media/prashant1901/New%20Volume/GSOC/nix-security-tracker/src/shared/models/cve.py:113:0-116:44) objects from identical references across different containers so that no metadata is lost (e.g. merging `vdb-entry` and `x_refsource_REDHAT` dynamically).
- Adjusted the [Reference](cci:2://file:///media/prashant1901/New%20Volume/GSOC/nix-security-tracker/src/webview/suggestions/context/types.py:16:0-20:19) dataclass `tags` type hint to correctly accept standard Django model instances, appeasing strict static analysis tools.

Fixes #876

### DIFF
--- a/src/webview/suggestions/context/types.py
+++ b/src/webview/suggestions/context/types.py
@@ -1,7 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
-from typing import Any
-
+from shared.models import Tag
 from shared.logs.batches import batch_events
 from shared.logs.events import (
     Maintainer,  # FIXME(@florent): This is to import it from that module
@@ -18,7 +17,7 @@ from shared.models.linkage import CVEDerivationClusterProposal
 class Reference:
     url: str
     name: str
-    tags: list[Any]
+    tags: set[Tag]
 
 
 # Packages
@@ -197,15 +196,11 @@ class SuggestionContext:
                     refs_by_url[url] = Reference(
                         url=url,
                         name=ref.name,
-                        tags=list(ref.tags.all()),
+                        tags=set(ref.tags.all()),
                     )
                 else:
                     existing_ref = refs_by_url[url]
-                    existing_tag_values = {tag.value for tag in existing_ref.tags}
-                    for tag in ref.tags.all():
-                        if tag.value not in existing_tag_values:
-                            existing_ref.tags.append(tag)
-                            existing_tag_values.add(tag.value)
+                    existing_ref.tags.update(ref.tags.all())
                     
                     if not existing_ref.name and ref.name:
                         existing_ref.name = ref.name
--- a/src/webview/suggestions/context/types.py
+++ b/src/webview/suggestions/context/types.py
@@ -1,5 +1,6 @@
 from dataclasses import dataclass
 from enum import Enum
+from typing import Any
 
 from shared.logs.batches import batch_events
 from shared.logs.events import (
@@ -17,7 +18,7 @@ from shared.models.linkage import CVEDerivationClusterProposal
 class Reference:
     url: str
     name: str
-    tags: list[str]
+    tags: list[Any]
 
 
 # Packages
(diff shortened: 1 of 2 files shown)


fricklerhandwerk (changes requested):
https://github.com/NixOS/nix-security-tracker/blob/9214c169728c79aa96d41efdad18628b15dcb349/CONTRIBUTING.md#L85

src/webview/suggestions/context/types.py
```diff
@@ -188,17 +189,27 @@ def update_maintainer_list_context(
...
-                refs.append(
-                    Reference(
-                        url=ref.url,
+                url = ref.url
+                if url not in refs_by_url:
+                    refs_by_url[url] = Reference(
+                        url=url,
                         name=ref.name,
-                        tags=[tag for tag in ref.tags.all()],
+                        tags=list(ref.tags.all()),
```
fricklerhandwerk: This is clever but loses us type safety, which in turn affects readability. Anyone used to strong typing will go "huh >_>" seeing `tags: list[Any]`.

(WTFs/min is an actual measure of code quality.)

src/webview/suggestions/context/types.py
```diff
@@ -17,7 +18,7 @@
 class Reference:
     url: str
     name: str
-    tags: list[str]
+    tags: list[Any]
```
fricklerhandwerk: How about making that a set instead, then no need to manually loop?

devprashant19: I've updated the PR to securely type the references using the explicit [Tag](cci:2://file:///media/prashant1901/New%20Volume/GSOC/nix-security-tracker/src/shared/models/cve.py:113:0-116:44) model (`set[Tag]`).

As you cleverly suggested, I also refactored the deduplication to gracefully orchestrate `set.update()` on native Python sets, which cleanly eliminates the clunky nested `.append()` loops from my previous commit.

The updated `Reference` dataclass now natively imports the Django model for strict typing:

```python
from shared.models import Tag

@dataclass
class Reference:
    url: str
    name: str
    tags: set[Tag]
```
And the logic has been simplified:
```python
    def update_references(self) -> None:
        refs_by_url = {}
        for container in self.suggestion.cve.container.all():
            for ref in container.references.all():
                url = ref.url
                if url not in refs_by_url:
                    refs_by_url[url] = Reference(
                        url=url,
                        name=ref.name,
                        tags=set(ref.tags.all()),
                    )
                else:
                    existing_ref = refs_by_url[url]
                    existing_ref.tags.update(ref.tags.all())
                    
                    if not existing_ref.name and ref.name:
                        existing_ref.name = ref.name
        self.references = list(refs_by_url.values())
```

fricklerhandwerk (approved):

src/webview/suggestions/context/types.py
fricklerhandwerk: Good. No need to repeat the contents of the commits in comments though, I really have enough stuff to read through every day :)

src/webview/tests/test_references.py
```diff
@@ -0,0 +1,40 @@
+from collections.abc import Callable
+import pytest
+
+from shared.listeners.cache_suggestions import cache_new_suggestions
+from shared.models.linkage import CVEDerivationClusterProposal
+from shared.models.cve import Container, Tag, Reference
+from webview.suggestions.context.types import SuggestionContext
+
+@pytest.mark.django_db
+def test_suggestion_context_reference_deduplication(
```
fricklerhandwerk, line 10: This test is semantically alright, but please make it a playwright test, following the pattern of the others. That runs a lot slower, I know, but I want to avoid pinning implementation details that will change anyway.

fricklerhandwerk (changes requested):
Sorry for the premature approval, I need to slow down a bit with the reviews.

src/webview/suggestions/context/types.py
devprashant19: sry :)

src/webview/tests/test_references.py
fricklerhandwerk: And here's the reason why testing implementation details is not a good idea: https://github.com/NixOS/nix-security-tracker/issues/876#issuecomment-4133064339

Whichever implementation we end up adopting, a test of the outward interface will just stay the way it is.

fricklerhandwerk:
This is now fully superseded by https://github.com/NixOS/nix-security-tracker/pull/994. Sorry @devprashant19 for the detour. Your work for sure helped thinking through the problem more clearly, so thank you very much for the contribution in any case, it is highly appreciated!

devprashant19:
Not a problem :) I'll work on something else

PR #936, opened 2026-03-25
docs: add quick start helper command and instructions

fricklerhandwerk:
This allows spinning up the service quickly.
Details on GitHub intergration are deferred to later.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -163,16 +163,51 @@ To replicate this on a traditional Unix-like system:
 
 ### Start the service
 
-The service is comprised of the Django server and workers for ingesting CVEs and derivations.
-What needs to be run is defined in the [`Procfile`](./Procfile) managed by [hivemind](https://github.com/DarthSim/hivemind).
+> ![NOTE]
+> For a quick start, create dummy credentials:
+>
+> ```console
+> dummy-credentials
+> ```
+>
+> Logging in and publishing issues requires [setting up credentials](#setting-up-credentials).
 
-Run everything with:
+Run the server:
 
-```bash
-hivemind
+```console
+manage runserver
 ```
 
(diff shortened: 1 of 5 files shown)


CONTRIBUTING.md
```diff
@@ -319,20 +353,7 @@ We use the following pattern:
...
-
-Add 100 CVE entries to the database:
-
-```console
-manage ingest_bulk_cve --subset 100
-```
-
-This will take a few minutes on an average machine.
-Not passing `--subset N` will take about an hour and produce ~500 MB of data.
-
```
florentc, line 352: Why do we remove this?

CONTRIBUTING.md
```diff
@@ -227,16 +227,50 @@ To replicate this on a traditional Unix-like system:
...
+manage fetch_all_channels
 ```
 
-<!-- FIXME(@fricklerhandwerk): Add instructions for manually obtaining CVEs and derivations. -->
+Finally, ingest some CVEs:
+
+<!-- FIXME(@fricklerhandwerk):
+Change the command to specify a date bracket.
+Otherwise, if that particular subset doesn't produce matches, using a larger subset will re-ingest the existing ones.
+-->
```
florentc: Oh good one!

CONTRIBUTING.md
fricklerhandwerk: Because mentioning that particular command s is moved to the quickstart instructions, which I deem enough to get exposed to it. (I also removed the note on duration and storage because that doesn't age well and I don't even know how true it still is.)

contrib/mkcredentials.sh
florentc, line 1: By the way why don't we delete this altogether

fricklerhandwerk: Oh I didn't... oops.

PR #943, opened 2026-03-25
refactor: added overlay base model for all overlay model to inherit.

adekoder:
Description:

This PR introduce a Base Abstract Class  `Overlay` for Overlay models to inherit from.


src/shared/models/base.py
fricklerhandwerk, line 1: Maybe just put it into `models/linkage.py` so it stays together.

src/shared/models/base.py
```diff
@@ -0,0 +1,17 @@
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+
+class BaseAbstractOverlay(models.Model):
```
fricklerhandwerk:
```suggestion
class Overlay(models.Model):
```

fricklerhandwerk (approved):

PR #945, opened 2026-03-25
refactor: update overlay choice field option values

adekoder:
Description:

This PR is a follow up to this one [944](https://github.com/NixOS/nix-security-tracker/pull/944)

- It update the `Type` Field Choices Values from, `added` to `additional` and `remove` to `ignored`
- upated the activity log event label and ensure the correct event are recording `add`, `delete`, `ignore`, `restore`

- This PR also include the command to handle the data migration 
`migrate_overlay_data`

Note:
>We need to prepare a script to update the old record for the MaintainerOverlay and PackageOverlay Model and corresponding events models, during deployment.

### DIFF
--- a/src/shared/listeners/cache_suggestions.py
+++ b/src/shared/listeners/cache_suggestions.py
@@ -497,13 +497,13 @@ def categorize_maintainers(
 
     # Process edits to categorize maintainers
     ignored_github_ids = set()
-    added_maintainers = []
+    additional_maintainers = []
 
     for overlay in maintainer_overlays:
         if overlay.edit_type == MaintainerOverlay.Type.IGNORED:
             ignored_github_ids.add(overlay.maintainer.github_id)
         elif overlay.edit_type == MaintainerOverlay.Type.ADDITIONAL:
-            added_maintainers.append(
+            additional_maintainers.append(
                 CachedSuggestion.Maintainer.model_validate(to_dict(overlay.maintainer))
             )
 
@@ -521,5 +521,5 @@ def categorize_maintainers(
         original=original_maintainers,
         active=active_maintainers,
         ignored=ignored_maintainers,
-        added=added_maintainers,
+        added=additional_maintainers,
     )
--- a/src/shared/management/commands/migrate_overlay_data.py
+++ b/src/shared/management/commands/migrate_overlay_data.py
@@ -1,0 +1,124 @@
+import logging
+from typing import Any
+
+from django.core.management.base import BaseCommand
+from django.db import connection
+
+from shared.models import (
+    MaintainerOverlay,
+    MaintainerOverlayEvent,
+    PackageOverlay,
+    PackageOverlayEvent,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def migrate_records(migration: dict) -> None:
+    model = migration["class"]
+    data_mapping = migration["data_mapping"]
+    model_id = migration["model_id"]
+    field_to_update = migration["field_to_update"]
+    table_name = model._meta.db_table
+
+    print(f"Running data migration for {table_name}")
+
+    if migration["is_event_table"]:
+        with connection.cursor() as cursor:
+            # Disable the append-only trigger as we have "append-only" trigger
+            # that prevents updates and deletes.
+            # let temporarily disable it for the migration.
+            cursor.execute(f"ALTER TABLE {table_name} DISABLE TRIGGER ALL")
+
+    try:
+        for old_data, new_data in data_mapping.items():
+            print(f"Migration {old_data} -> {new_data}")
+            total_updated = 0
+            batch_size = 1000
+
+            while True:
+                batch_ids = list(
+                    model.objects.filter(**{field_to_update: old_data}).values_list(
+                        model_id, flat=True
+                    )[:batch_size]
+                )
+                if not batch_ids:
+                    break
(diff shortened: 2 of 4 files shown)


src/shared/models/linkage.py
```diff
@@ -79,39 +79,41 @@ def is_frozen(self) -> bool:
...
     """
-    A single manual edit of the list of maintainers of a suggestion.
+    An element in the overlay set of maintainers of a suggestion.
     """
 
-    class EditType(models.TextChoices):
-        ADD = "add", _("add")
-        REMOVE = "remove", _("remove")
+    # TODO (@adekoder): Preprare and run a script to update older records
+    class OverlayType(models.TextChoices):
```
```suggestion
    class Type(models.TextChoices):
```

From overall reading it seems this would be easier on the eyes

src/shared/models/linkage.py
```diff
@@ -156,59 +160,59 @@ class Meta:
...
-    sender: type[PackageEdit],
-    instance: PackageEdit,
+    sender: type[PackageOverlay],
+    instance: PackageOverlay,
     created: bool,
     **kwargs: Any,
 ) -> None:
     if created:
-        # TODO Adapt when PackageEdit supports more than REMOVE
+        # TODO Adapt when PackageOverlay supports more than REMOVE
```
```suggestion
        # Adapt when PackageOverlay supports more than IGNORED
```

Or remove the comment altogether, it should actually be obvious from context

src/shared/listeners/cache_suggestions.py
```diff
@@ -402,7 +402,7 @@ def maintainers_list(packages: dict, edits: list[MaintainersEdit]) -> list[dict]
 
 def categorize_maintainers(
     packages: dict[str, CachedSuggestion.Package],
-    maintainers_edits: list[MaintainersEdit],
+    maintainers_edits: list[MaintainerOverlay],
```
fricklerhandwerk:
```suggestion
    maintainer_overlays: list[MaintainerOverlay],
```

Not sure whether this should be named by the whole set, i.e. `maintainers_overlay`... But at least let's make it consistent with the one for packages.

adekoder: The model name is singular, so the set is plural , so when you read you will always read it as `maintainer_overlays` and not `maintainers_overlay`

src/shared/models/linkage.py
adekoder: done

src/shared/listeners/cache_suggestions.py
```diff
@@ -496,23 +496,23 @@ def categorize_maintainers(
     original_maintainers = list(original_maintainers_dict.values())
 
     # Process edits to categorize maintainers
-    removed_github_ids = set()
+    ignored_github_ids = set()
     added_maintainers = []
```
florentc:
```suggestion
    additional_maintainers = []
```

To keep the same name as the overlay type

src/shared/migrations/0079_alter_maintaineroverlay_edit_type_and_more.py
florentc, line 1: Shouldn't there be something to migrate the existing events in the activity log in the following way:

For packages
add -> restore
remove -> ignore

For maintainers
add -> add
remove -> delete

Since now we have also renamed the labels of these events and added 2 new ones in the pghistory.ManualEvent annotations

src/webview/tests/test_packages.py
```diff
@@ -179,9 +179,10 @@ def test_restore_one_of_multiple_ignored_packages(
     ignored.click()
     ignored.locator(".package-alpha").get_by_role("button", name="Restore").click()
 
-    ignored.click()
     expect(active.get_by_text("alpha")).to_be_visible()
     expect(active.get_by_text("charlie")).to_be_visible()
+    # Expand the ignored packages section before asserting since it will close due to re-render
+    container.get_by_text(re.compile("Ignored packages")).click()
```
florentc:
```suggestion
    container.get_by_text("Ignored packages").click()
```

What is this about?
An unreliable test?

For clarity maybe it should have its own tiny PR fix rather than be part of this PR.

adekoder: this part of my PR started failing because of this, but i can create another PR for though.

src/shared/migrations/0079_alter_maintaineroverlay_edit_type_and_more.py
adekoder: i have a script PR that handles the data migration [(PR)](https://github.com/NixOS/nix-security-tracker/pull/979)

src/shared/listeners/cache_suggestions.py
adekoder: i did not want to touch the rename of some code variables too much, so as not to make this PR too big, i plan on create another PR for this, what do you think ?

src/shared/migrations/0079_alter_maintaineroverlay_edit_type_and_more.py
fricklerhandwerk: I suppose the change and the migration should be the same commit, logically. Yes, in practice we'll deploy them all at once anyway, but effectively this commit breaks the activity log by making it undisplayable. 

(Also I already made review comments on the other one, and propagating changes back and forth is clearly a hassle...)

fricklerhandwerk (approved):

PR #946, opened 2026-03-26
feat: auto-reject CVEs with hardware-only CPE affected products

ArkhamKnight25:
Follow-up to #894 as suggested in review.

When all affected products in a CVE have hardware-only CPEs (`part=h`),
`build_new_links` was silently returning `False` with no record of why.

This adds `HARDWARE_ONLY_CPE` to `RejectionReason` and creates a rejected
proposal when all affected products are filtered out by the hardware CPE
check, the same pattern used for `exclusively-hosted-service`.

Products with no CPEs are unaffected.


src/shared/tests/test_linkage.py
```diff
@@ -194,8 +197,35 @@ def test_mixed_cpe_parts_skips_hardware_only_affected_products(
...
+
+    result = build_new_links(container)
+
+    assert result is True
+    proposal = CVEDerivationClusterProposal.objects.get(cve=container.cve)
+    assert proposal.status == CVEDerivationClusterProposal.Status.REJECTED
+    assert (
+        proposal.rejection_reason
+        == CVEDerivationClusterProposal.RejectionReason.HARDWARE_ONLY_CPE
+    )
```
fricklerhandwerk: Those checks can be moved into the other only-hardware test above, no need to make an extra test

src/shared/listeners/automatic_linkage.py
```diff
@@ -137,7 +129,27 @@ def build_new_links(container: Container) -> bool:
         )
         return True
 
-    drvs = produce_linkage_candidates(container)
+    has_any_cpe = Exists(Cpe.objects.filter(affectedproduct=OuterRef("pk")))
```
fricklerhandwerk, line 135: Please leave the FIXME comment, the smell isn't gone

src/shared/models/linkage.py
```diff
@@ -36,6 +36,10 @@ class RejectionReason(models.TextChoices):
             "not_in_nixpkgs",
             _("not in Nixpkgs"),
         )
+        HARDWARE_ONLY_CPE = (
+            "hardware_only_cpe",
+            _("hardware only CPE"),
```
```suggestion
            _("CPEs only affect hardware"),
```

```suggestion
            _("hardware only"),
```


This will be user-facing for the foreseeable future, it should be grammatically correct

src/shared/tests/test_linkage.py
ArkhamKnight25: Merged the assertions into `test_hardware_cpe_produces_no_match` and removed the separate test.

src/shared/listeners/automatic_linkage.py
ArkhamKnight25: Added back.

src/shared/models/linkage.py
ArkhamKnight25: Changed to `"hardware only"`

PR #956, opened 2026-03-28
refactor: parse CPEs with a dedicated library

DarshanCode2005:
Description:
This PR replaces manual string-based CPE parsing with the standardized `cpe` version 1.3.1 library. 

Added cpe python library (v 1.3.1) library to overlay.nix, using a secure hash for reproducibility.


fricklerhandwerk:
Finally got around to inspecting that library. I'm not excited about depending on a repo that hasn't been updated in two years, but it looks rather robust and very close to the spec, so it's better than anything we'd come up with any time soon. Rebased on top of `main` and did some lateral cleanups. @DarshanCode2005 please take another look.

src/shared/models/cve.py
```diff
@@ -247,37 +249,34 @@ def affects(self, version: str | None) -> str:
...
-        ]
-        triggers = [
-            # Add a trigger to maintain the search vector updated with row changes
-            UpdateSearchVector(
-                name="cpe_search_vector_idx",
-                vector_field="search_vector",
-                document_fields=["name"],
-            )
-        ]
+    def is_hardware(self) -> bool:
```
fricklerhandwerk: The reason I didn't keep it is that we'd be doing the validation/fallback every single time rather than just on ingestion. Also it doesn't scale nicely: now you'd have to do the same dance for every method you want to call on `CPE()`. My omission was to not add (some equivalent of) `.full_clean()` to the ingestion code path, where we should log an error and throw away that particular CPE.

If something crashes (such as pre-existing malformed data) we'll get notified and clean it up. Not impossible, but unlikely, and worth our attention in that case and only then.

DarshanCode2005:
LGTM, I actually didn't read the commit description properly and restored some changes. I'll move the `full_clean` in make_cpe(fetchers.py) in a follow-up PR.

PR #971, opened 2026-04-01
fix: pass issue context through template context instead of ad-hoc model attributes

abdella-h:
## Problem 
`get_context_data()` attached `suggestion_context` and `github_issue` directly onto `NixpkgsIssue` model instances — fields that don't exist on the model. This only worked accidentally because Django templates use `object.__dict__` internally.

## Fix
Build an explicit `issue_contexts` list of dictionaries and pass it through the template context properly. Update `issue_list.html` to 
iterate over `issue_contexts` instead of `object_list`.


florentc (changes requested):
Thank you for your contribution. You are very close to the pattern we use for most of the suggestion-related templates, so as a step towards more consistency could you please go one step further and fit the pattern in question? By this I mean define an IssueContext type and have the issue component take only one parameter, `data` of type `IssueContext`, and then use the various `data.this` or `data.that` in the issue template.

src/webview/suggestions/context/types.py
```diff
@@ -270,3 +271,12 @@ def fetch_activity_log(self) -> None:
...
         )
+
+# Issues 
+
+
+@dataclass
+class IssueContext:
+    issue: NixpkgsIssue
+    suggestion_context: SuggestionContext
+    github_issue: str | None
```
florentc: This belongs alongside the issue view. Please either move the issue view from the global views.py to its specific directory like for suggestions, or just define IssueContext alongside the view. You defined the type with the suggestions here.

src/webview/templates/issue_list.html
```diff
@@ -12,8 +12,8 @@ <h1 class="page-title row gap-small centered">{% status_icon "published" %}Publi
   </div>
 
   <div class="column gap">
-    {% for object in object_list %}
-      {% issue object object.suggestion_context github_issue=object.github_issue show_permalink=True %}
+    {% for data in issue_contexts %}
+      {% issue data show_permalink=True %}
```
florentc:
```suggestion
    {% for issue_context in issue_contexts %}
      {% issue issue_context %}
```

Please move show_permalink to the issue_context too

src/webview/templatetags/viewutils.py
```diff
@@ -145,16 +146,12 @@ def iso(date: datetime.datetime) -> str:
...
     context: Context,
-    issue: NixpkgsIssue,
-    suggestion_context: SuggestionContext,
-    github_issue: str | None,
+    data: IssueContext,
     show_permalink: bool = False,
 ) -> dict:
     return {
-        "issue": issue,
+        "data": data,
```
florentc, line 152: Please move `show_permalink` to the issue_context too (this should just have `data` as a parameter in the end, and set it in the view

florentc (changes requested):

src/webview/suggestions/context/types.py
abdella-h: I moved `IssueContext` to views.py

src/webview/templatetags/viewutils.py
abdella-h: The `IssueContext` now includes `show_permalink`, and the `{% issue %}` tag only takes `data` as a parameter

fricklerhandwerk:
@florentc I think we can close this given we're not working on the old UI code paths any more?

florentc:
Indeed

PR #977, opened 2026-04-05
fix: limit retries when waiting for Git's shallow lockfile

WHOIM1205:
### Fix infinite retry loop in `update_from_ref`

This PR fixes an issue where `update_from_ref` could get stuck in an endless loop when a stale `shallow.lock` file exists or when git repeatedly returns shallow-related errors.

#### What’s changed
- Added a timeout while waiting for `shallow.lock` and clean it up if it looks stale
- Limited the number of retries for fetch attempts
- Added exponential backoff between retries
- Fixed logic so successful fetches with "shallow" warnings don’t trigger retries

#### Why this matters
Previously, this could cause evaluation workers to hang forever, blocking evaluation slots and stopping further processing. With these changes, the system will either recover or fail gracefully instead of getting stuck.

### DIFF
--- a/src/shared/git.py
+++ b/src/shared/git.py
@@ -3,6 +3,8 @@ import itertools
 import logging
 import os.path
 import pathlib
+import random
+import time
 from collections.abc import AsyncGenerator
 from contextlib import asynccontextmanager
 from dataclasses import dataclass
@@ -126,7 +128,14 @@ class GitRepo:
             )
         )
 
-    async def update_from_ref(self, object_sha1: str) -> bool:
+    async def update_from_ref(
+        self,
+        object_sha1: str,
+        # Age above which a leftover `shallow.lock` is considered stale and removed.
+        max_lock_age_seconds: int = 300,
+        # Retries between attempts happen with exponential backoff. Increase with care.
+        max_attempts: int = 7,
+    ) -> bool:
         """
         This checks if `object_sha1` is already present in the repo or not.
         If not, perform a fetch to our remote to obtain it.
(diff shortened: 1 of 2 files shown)


src/shared/git.py
```diff
@@ -146,24 +151,54 @@ async def update_from_ref(self, object_sha1: str) -> bool:
...
+                locking_problem = rc != 0 and "shallow" in stderr_text
                 if rc != 0 and not locking_problem:
-                    logger.error("git fetch stderr: %s", stderr.decode("utf8"))
+                    logger.error("git fetch stderr: %s", stderr_text)
                     raise RepositoryError(
                         f"failed to fetch {object_sha1} while running `git fetch --depth=1 {repo_clone_url} {object_sha1}`"
                     )
+                if locking_problem:
+                    retries += 1
+                    backoff = min(2**retries, 30)
```
fricklerhandwerk, line 193: How about just setting `max_retries` to 6 and adding a comment on the parameter that retries will happen with exponential backoff, so setting it it to more than a small number should be done with care?

src/shared/git.py
```diff
@@ -146,24 +151,54 @@ async def update_from_ref(self, object_sha1: str) -> bool:
...
+                    lock_waited += 1.0
                 process = await self.execute_git_command(
                     f"git fetch --porcelain --depth=1 {repo_clone_url} {object_sha1}",
                     stderr=asyncio.subprocess.PIPE,
                 )
                 _, stderr = await process.communicate()
                 rc = await process.wait()
-                locking_problem = "shallow" in stderr.decode("utf8")
+                stderr_text = stderr.decode("utf8")
+                locking_problem = rc != 0 and "shallow" in stderr_text
```
fricklerhandwerk, line 185:
```suggestion
                # TOCTOU: A different process got the lock since we found it was released.
                locking_problem = rc != 0 and "shallow" in stderr_text
```

I think that would slightly help following why we care

src/shared/git.py
```diff
@@ -126,7 +126,12 @@ async def worktrees(self) -> list[Worktree]:
             )
         )
 
-    async def update_from_ref(self, object_sha1: str) -> bool:
+    async def update_from_ref(
+        self,
+        object_sha1: str,
+        max_lock_wait: float = 300.0,
```
fricklerhandwerk, line 132:
```suggestion
        max_lock_wait_seconds: int = 10,
```

Not sure why float? But it should sure tell the unit.

Realistically, does it make sense to wait for 5 min and then back off for 1 second to wait up to 5 min again? We're just fetching a commit, can't take *that* long. And if we do fetch 10 branches for some reason, and try 6 times with exponential backoff, should still be okay to poll for not more than like 10-30 seconds each.

Just an idea. Maybe a 30s is better?

src/shared/git.py
```diff
@@ -146,24 +151,54 @@ async def update_from_ref(self, object_sha1: str) -> bool:
...
+                            lock_waited,
+                            lock_path,
+                        )
+                        try:
+                            os.remove(lock_path)
+                        except FileNotFoundError:
+                            pass  # Another process removed it first.
+                        break
                     await asyncio.sleep(1)
+                    lock_waited += 1.0
```
fricklerhandwerk, line 177: It's a bit of an opinion thing, but it does seem tasteful to count ""seconds"" here. Maybe catch timestamps and compare the difference with max_wait?

src/shared/git.py
```diff
@@ -146,24 +151,54 @@ async def update_from_ref(self, object_sha1: str) -> bool:
         if exists:
             return False
         else:
+            lock_path = os.path.join(self.repo_path, "shallow.lock")
```
fricklerhandwerk, line 154: It would help the uninitiated a bit if we explain why we do all this.

```suggestion
            # Git tries to create `shallow.lock` before attempting a shallow fetch, in order to prevent conflicts on the `shallow` metadata file:
            # https://git-scm.com/docs/shallow
            # Poll for its existence directly to avoid running Git repeatedly.
            lock_path = os.path.join(self.repo_path, "shallow.lock")
```

Although if running Git every second is unnoticeably cheap, it would save us some 20LOC. Then we may just run the loop like O(100) times before giving up and logging an error. What do you think?

src/shared/git.py
```diff
@@ -146,24 +151,54 @@ async def update_from_ref(self, object_sha1: str) -> bool:
...
                 if rc != 0 and not locking_problem:
-                    logger.error("git fetch stderr: %s", stderr.decode("utf8"))
+                    logger.error("git fetch stderr: %s", stderr_text)
                     raise RepositoryError(
                         f"failed to fetch {object_sha1} while running `git fetch --depth=1 {repo_clone_url} {object_sha1}`"
                     )
+                if locking_problem:
+                    retries += 1
+                    backoff = min(2**retries, 30)
+                    logger.warning(
```
fricklerhandwerk, line 194: Probably this should be an error. @Erethon would we get an email about it? Because if after many minutes we still have a `shallow.lock` someone needs to clean it up manually, otherwise we won't be able to fetch channels and ~a day will pass until we see that evals are not happening.

fricklerhandwerk (commented):
This is a good catch, and a nice careful approach, thank you very much. Left some comments for polishing it.

src/shared/git.py
fricklerhandwerk: If we take the "delete lockfile after a long timeout" approach from #976 we may not even need more than a warning though, as it will be self-healing. (@Erethon I'd still be interested about a code pattern for sending us error emails deliberately)

fricklerhandwerk:
Gave it another thought and simplified further. Also added tests -- it's quite the state space to check through. Thanks @WHOIM1205 for scaffolding the process mocking earlier, that helped a lot getting the tests in place.

PR #986, opened 2026-04-09
fix: unbreak API endpoint

Erethon:
I'm opening this as a reply to #985  (which merging this PR will close #985). If we don't care about DRF and think we should get rid of it for now, this PR completely removes any DRF code, the broken view it's used in and removes it from the dependencies as well.

### DIFF
--- a/src/shared/tests/test_api.py
+++ b/src/shared/tests/test_api.py
@@ -1,0 +1,57 @@
+from collections.abc import Callable
+
+from rest_framework.reverse import reverse
+from rest_framework.test import APIClient
+
+from shared.models.cve import Container
+from shared.models.issue import (
+    NixpkgsIssue,
+)
+from shared.models.linkage import (
+    CVEDerivationClusterProposal,
+)
+
+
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Fixed it instead :)

Erethon:
I think it would be best to close this then and open a new PR with the commit that fixes the API, instead of overwriting this PR that removed Django Rest Framework. Messing with the remote branch of an opened PR should be done for minor fixes so the maintainers can prevent lengthy back and forths for minor changes in my opinion. Besides the fact that your commit might be accidentally overwritten by me if I didn't notice the force push, the history of this PR (and the original message) don't make a lot of sense anymore.

Edit: I think closing this, me force pushing back the original commit so that history is retained and opening a new PR with your fix would make sense. What do you think?

fricklerhandwerk:
In this case since we know what's going on, I think the simplest is merging it as is. I don't ascribe a terrible lot of value to GitHub PRs. The source of truth is Git commits.

PR #994, opened 2026-04-14
feat: View and edit deduplicated references

florentc:
This PR addresses #876: now users view and edit references through the lens of deduplication.

On the model level, almost nothing changes in suggestions and references. A suggestion may have several references sharing the same url, like before.

On the UI side, users will only see one entry per duplicated group of reference. In other words, given the equivalence relation "having the same url", the users no longer see/edit references but the equivalence classes. When applicable, the common name is used, and the tags are merged.

To make it possible, `ReferenceOverlay` have been modified into `ReferenceUrlOverlay`. Overlays are characterized by the url in question, not by a single reference object. Computing the deduplicated references and applying the overlays is done in the cached suggestions (whose model version is bumped).

The db migration applies the deduplication as such:
- in a suggestion, if all references sharing the same url have a `ReferenceOverlay` (IGNORE), then a corresponding `ReferenceUrlOverlay` is introduced
- if some references of a given equivalence class have an overlay (aka are ignored) but not others, then we don't introduce a corresponding `ReferenceUrlOverlay` as it means the user may have just ignored a few of them to reduce visual clutter of duplication until now, with no intent to actually ignore the reference in question

This migration implies information loss so **it is irreversible**.
Migrating the events related to `ReferenceOverlay` isn't really possible (unclear semantics and hazardous implementation), therefore, reference edition events **are discarded**. This means reference ignore/restore from before this change will no longer appear in activity logs. This has been agreed as an acceptable loss with @fricklerhandwerk 
After merge, cached suggestions **will have to be regenerated**.

### DIFF
--- a/src/shared/github.py
+++ b/src/shared/github.py
@@ -116,7 +116,7 @@ def create_gh_issue(
         """
         Format active references from the cached suggestion for inclusion in GitHub issue.
         """
-        references = cached_suggestion.payload.get("categorized_references", {})
+        references = cached_suggestion.payload.get("categorized_url_references", {})
         active_refs = references.get("active", [])
 
         if not active_refs:
--- a/src/shared/listeners/cache_suggestions.py
+++ b/src/shared/listeners/cache_suggestions.py
@@ -2,6 +2,7 @@ import itertools
 import logging
 import re
 import urllib.parse
+from collections import defaultdict
 from datetime import datetime
 from itertools import chain
 from typing import Any, overload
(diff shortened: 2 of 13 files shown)


fricklerhandwerk (approved):
Very nice!

Test are still failing, but I can't tell why. Please check if migration dependencies are up to date against `main`.

In the commit message or migration code it's worth mentioning that we're deliberately dropping the old reference events for simplicity, as it would be really cumbersome to correctly reconstruct the intent, with questionable payoff.

PR #995, opened 2026-04-14
Refactor per-channel output and add tests for fetch_all_channels

DarshanCode2005:
` # FIXME(@fricklerhandwerk): Fold that into `branch_info`, so there's only one output.
        print("Parallel fetches results", results)`
**What’s changed:**
* Cleaner output: Each channel now prints one clear line with its details (release, state, commit) and result (fetched, already present, or error), after all fetches finish.
* Standard output: Replaced `print`/`pprint` with `self.stdout.write` for proper command output handling.
* Tests added:
  * 4 unit tests for reporting logic
  * 1 integration test for the full command flow with the database

### DIFF
--- a/src/shared/management/commands/fetch_all_channels.py
+++ b/src/shared/management/commands/fetch_all_channels.py
@@ -1,6 +1,6 @@
 import asyncio
 import sys
-from collections.abc import Coroutine, Sequence
+from pprint import pformat
 from typing import Any
 
 import requests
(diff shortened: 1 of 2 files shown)


src/shared/management/commands/fetch_all_channels.py
```diff
@@ -93,38 +91,65 @@ def fetch_from_monitoring() -> dict[str, MonitoredChannel]:
...
+        # the results list aligns with `registered` for the zip below.
         repo = GitRepo(
             settings.LOCAL_NIXPKGS_CHECKOUT,
             stderr=sys.stderr.fileno(),
         )
-        parallel_fetches = []
-        for channel in NixChannel.objects.iterator():
-            parallel_fetches.append(repo.update_from_ref(channel.head_sha1_commit))
+        coros = [repo.update_from_ref(ch.head_sha1_commit) for ch, _ in registered]
+        results = asyncio.run(wait_for_parallel_fetches(coros))
```
fricklerhandwerk, line 151: That probably also works

```suggestion
        results = asyncio.run(wait_for_parallel_fetches(
            repo.update_from_ref(ch.head_sha1_commit) for ch, _ in registered
        ))
```

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -93,38 +91,65 @@ def fetch_from_monitoring() -> dict[str, MonitoredChannel]:
...
+            channel, _ = NixChannel.objects.update_or_create(
+                {
+                    "staging_branch": branch_info["staging_branch"],
+                    "state": branch_info["state"],
+                    "head_sha1_commit": branch_info["head_sha1_commit"],
+                    "release_version": branch_info["release_version"],
+                },
+                channel_branch=monitored.name,
             )
+            registered.append((channel, branch_info))
```
fricklerhandwerk, line 142: Instead of repeating all the dict field names, how about instead:

```suggestion
            report_info = {"channel_branch": channel_branch, **defaults}
            registered.append((channel, report_info))
```

The change can probably be a lot smaller then. I don't think there's much point in renaming the `channel` iterator variable.

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -93,38 +91,65 @@ def fetch_from_monitoring() -> dict[str, MonitoredChannel]:
...
-        for channel in fresh_channels.values():
-            channel_branch = channel.name
-            staging_branch = staging_from_branch(channel.name)
-            branch_info = {
-                "staging_branch": staging_branch,
-                "state": state_from_status(channel.status),
-                "head_sha1_commit": channel.revision,
-                "release_version": release_from_branch(channel.name),
+
+        # Step 1: register/update all channels in the DB, collect for reporting.
```
fricklerhandwerk, line 123:
```suggestion
```

I generally discourage this sort of comment for two reasons:
- Which step in a sequence this is may change, and so the comment is likely to drift
- It doesn't add information. The code doesn't do anything fancy and looking at it for a couple of seconds will explain it fully.

Comments should almost exclusively explain *why* a non-obvious piece of code is the way it is. Very rarely, if something that should be simple looks complicated, there should be a summary of what it does. This happens sometimes in this codebase where an involved DB query for can't be broken down into procedures with appropriate names (or where such a breakdown would introduce more problems than it would solve). But that's certainly not the case here.

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -93,38 +91,65 @@ def fetch_from_monitoring() -> dict[str, MonitoredChannel]:
 
 
 async def wait_for_parallel_fetches(
-    parallel_fetches: list[Coroutine[Any, Any, bool]],
+    parallel_fetches: Sequence[Coroutine[Any, Any, bool]],
 ) -> list[Any]:
     return await asyncio.gather(*parallel_fetches, return_exceptions=True)
```
fricklerhandwerk, line 96: Arguably this procedure does nothing on its own, why not inline it while we're at it?

src/shared/tests/test_fetch_all_channels.py
```diff
@@ -0,0 +1,80 @@
...
+        "channel_branch": "nixos-unstable",
+        "release_version": None,
+    }
+    assert "unstable" in format_channel_report(unstable_info, False)
+
+
+@pytest.mark.django_db
+@patch("shared.management.commands.fetch_all_channels.fetch_from_monitoring")
+@patch("shared.management.commands.fetch_all_channels.GitRepo")
+def test_command_upserts_channels_and_reports_fetch_results(
```
fricklerhandwerk, line 50: This is a very good test to have!

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -93,38 +91,65 @@ def fetch_from_monitoring() -> dict[str, MonitoredChannel]:
...
+    release = branch_info.get("release_version") or "unstable"
+
+    if isinstance(fetch_result, BaseException):
+        status = f"ERROR ({fetch_result})"
+    elif fetch_result:
+        status = "fetched"
+    else:
+        status = "already present"
+
+    return f"[{channel}] release={release} state={state} commit={sha} -> {status}"
```
fricklerhandwerk, line 114: I imagined something much simpler: Just attach the fetch status to the already existing `branch_info` dict and `pprint` that.

Printing a line makes the log less readable, and will need work later to make it machine-readable once we start storing logs.

src/shared/management/commands/fetch_all_channels.py
fricklerhandwerk, line 142: We're not *really* using the `NixChannel` here -- that `head_sha1_commit` is also in the `branch_info` dict.

fricklerhandwerk (commented):
There's some good thinking in it (especially the new test), but I recommend keeping it simple here: Aggregate the data into the collection of dicts and `pprint` them at the end.

src/shared/management/commands/fetch_all_channels.py
DarshanCode2005: Done , passing branch_info directly to update_or_create as the updated dict.

src/shared/management/commands/fetch_all_channels.py
DarshanCode2005: Okay.

src/shared/management/commands/fetch_all_channels.py
DarshanCode2005: Okay, removed the helper function entirely; asyncio.gather is now called inline inside asyncio.run.

src/shared/tests/test_fetch_all_channels.py
DarshanCode2005: Thanks!

src/shared/management/commands/fetch_all_channels.py
DarshanCode2005: Hmm, okay format_channel_report is gone and pprints the dict directly.

src/shared/management/commands/fetch_all_channels.py
DarshanCode2005: Okay, I've simplified this.

fricklerhandwerk:
We can close this. For https://github.com/NixOS/nix-security-tracker/issues/864 I'm working on a change that will get the commits directly from Hydra, and then we don't need to fetch the channels but can take them directly from the infra repo's `channels.nix`. And we're already fetching the commit blobs right before evaluation, so that apparently was always redundant. See https://github.com/NixOS/nix-security-tracker/issues/864#issuecomment-4607690624 for work in progress.

PR #996, opened 2026-04-16
feat: Garbage collection command for stale data

adekoder:


src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+def _age_evaluation(evaluation: NixEvaluation, days: int = 400) -> None:
+    """Back-date an evaluation's updated_at to simulate ageing."""
+    NixEvaluation.objects.filter(pk=evaluation.pk).update(
+        updated_at=timezone.now() - timedelta(days=days)
+    )
+
+
+@pytest.mark.django_db
```
fricklerhandwerk, line 40:
```suggestion
```

This is generally not necessary, since the base fixtures depend on `db`.

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,282 @@
...
+
+        # The order here is intentional and required.
+        # Each step satisfies the PROTECT constraint that gates the next step.
+        # pghistory event tables use DO_NOTHING, db_constraint=False, so they
+        # are never auto-deleted — each step explicitly clears relevant events first.
+        #
+        # Step 1 — proposals:  clears events, then DerivationClusterProposalLink rows, satisfying PROTECT on NixDerivation
+        # Step 2 — derivations: clears NixDerivation rows, satisfying PROTECT on NixEvaluation
+        # Step 3 — evaluations: clears NixEvaluation rows, satisfying PROTECT on NixChannel
+        # Step 4 — channels:   clears NixChannel rows
```
fricklerhandwerk, line 58:
```suggestion
```

Don't repeat in comments what's in the code.

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    """Old PENDING proposal with a MaintainerOverlay is preserved — user engaged."""
+    suggestion = make_suggestion()
+    _age_proposal(suggestion)
+
+    maintainer = make_maintainer()
+    MaintainerOverlay.objects.create(
+        overlay_type=MaintainerOverlay.Type.ADDITIONAL,
+        maintainer=maintainer,
+        suggestion=suggestion,
+    )
```
fricklerhandwerk, line 138: This is the only variation for all the overlay types. Since we already have 3 and this is an obvious extension point, seems like it's worth deduplicating the stuff around it. I.e. I suggest to have one test that is parametrised over overlay type, and what happens for each type is handled in a conditional.

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+@pytest.mark.django_db
+def test_deletes_unmatched_derivation(
+    make_channel: Callable[..., NixChannel],
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+    make_maintainer: Callable[..., NixMaintainer],
+) -> None:
+    """Derivation with no proposal link from an old COMPLETED eval is deleted."""
```
fricklerhandwerk, line 210:
```suggestion
    """Derivation with no proposal link from an old eval is deleted."""
```

That it's completed seems insubstantial here

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+
+
+@pytest.mark.django_db
+def test_preserves_derivation_with_proposal_link(
+    make_channel: Callable[..., NixChannel],
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    """Derivation linked to an active proposal is never removed."""
```
fricklerhandwerk, line 234:
```suggestion
    """Derivation linked to a match is never removed."""
```

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+    make_maintainer: Callable[..., NixMaintainer],
+) -> None:
+    """Derivation with no proposal link from an old COMPLETED eval is deleted."""
+    channel = make_channel(branch="nixos-old-unmatched")
+    evaluation = make_evaluation(channel=channel)
+    _age_evaluation(evaluation)
+    maintainer = make_maintainer()
+    drv = make_drv(evaluation=evaluation, maintainer=maintainer)
```
fricklerhandwerk, line 215: One can as well obtain the evaluation and the maintainer from the derivation. This is a bit of a matter of taste of course, but it seems more natural to me not to introduce fixtures if you're not going to prepare anything about the object anyway.

Now that I'm thinking about it, why not set the creation date on evaluations in the fixture factory argument? Just move the logic for updating the date into there. Then it makes sense to explicitly construct e.g. the evaluation in the test case.

fricklerhandwerk (commented):
Partial review, didn't look at all the tests yet and didn't look at the implementation at all or considered the big picture yet. But the general idea of testing all conditions once is very good.

I'm slightly concerned about hard-coding the age cutoff into the fixtures. For production we'll of course only ever change the cutoff via the one place where the value is set. But in the fixtures I'd expect more like value+1 rather than a concrete number. Maybe it's worth exposing the cutoff age as a config setting?

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    assert NixDerivation.objects.filter(pk=drv.pk).exists()
+
+
+@pytest.mark.django_db
+def test_derivation_from_in_progress_eval_not_deleted(
+    make_channel: Callable[..., NixChannel],
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+) -> None:
+    """Only derivations from COMPLETED evals are eligible — IN_PROGRESS evals are skipped."""
```
fricklerhandwerk, line 252: We should also make sure that `WAITING` evals also aren't deleted.

Also don't repeat in comments what's in the code; only explain *why* the code is the way it is, otherwise there will be drift.

```suggestion
    """Only derivations from concluded evaluations are deleted"""
```

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+@pytest.mark.django_db
+def test_preserves_stable_channel(
+    channel: NixChannel,
+) -> None:
+    """STABLE channels are never touched by garbage collection."""
+    assert channel.state == NixChannel.ChannelState.STABLE
+
+    call_command("garbage_collect", stdout=StringIO())
+
+    assert NixChannel.objects.filter(pk=channel.pk).exists()
```
fricklerhandwerk, line 363: These can be folded into one parametrized test

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    make_evaluation: Callable[..., NixEvaluation],
+) -> None:
+    """COMPLETED evaluation newer than 1 year is not removed even if it has no derivations."""
+    channel = make_channel(branch="nixos-recent-empty")
+    evaluation = make_evaluation(channel=channel)
+    # Default updated_at is timezone.now(), which is recent, no explicit ageing
+
+    call_command("garbage_collect", stdout=StringIO())
+
+    assert NixEvaluation.objects.filter(pk=evaluation.pk).exists()
```
fricklerhandwerk, line 313: This can probably be folded into the other test that checks that old evals are deleted, again as a parameter

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    """REJECTED proposal older than 1 year is not deleted (user triaged it)."""
+    suggestion = make_suggestion(
+        status=CVEDerivationClusterProposal.Status.REJECTED,
+        rejection_reason=CVEDerivationClusterProposal.RejectionReason.NOT_IN_NIXPKGS,
+    )
+    _age_proposal(suggestion)
+
+    call_command("garbage_collect", stdout=StringIO())
+
+    assert CVEDerivationClusterProposal.objects.count() == 1
```
fricklerhandwerk, line 120: Checking states can be done via parameters, this could be one test together with the one about pending suggestions.

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,413 @@
...
+    make_reference: Callable[..., Reference],
+) -> None:
+    """Old PENDING proposal with a ReferenceOverlay is preserved — user engaged."""
+    suggestion = make_suggestion()
+    _age_proposal(suggestion)
+
+    reference = make_reference()
+    ReferenceOverlay.objects.create(
+        type=ReferenceOverlay.Type.IGNORED,
+        reference=reference,
```
fricklerhandwerk, line 178: This is not up to date with `main`, we're not referring to `Reference` any more but to URL. Check the fixture, you can attach references to a suggestion via `make_suggestion()`.

src/shared/migrations/0081_alter_derivationclusterproposallink_derivation_and_more.py
```diff
@@ -0,0 +1,34 @@
+# Generated by Django 5.2.9 on 2026-04-16 15:58
```
fricklerhandwerk, line 1: I think we can already merge the cascade changes separately, then there's less stuff to keep track of in this PR.

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,282 @@
...
+        Delete PENDING proposals with no reviewer engagement
+        that were created more than 1 year ago.
+
+        Before deleting proposals, explicitly removes all related pghistory
+        event records since their FKs use DO_NOTHING and won't auto-delete.
+
+        Deleting a proposal cascades to DerivationClusterProposalLink,
+        which satisfies PROTECT on NixDerivation for the next step.
+        """
+        self.stdout.write("\n[1/4] Stale proposals")
```
fricklerhandwerk, line 80: This should be printed where the steps are called. How do you know this will always be step 1/4?

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,282 @@
...
+    DerivationClusterProposalLinkEvent,
+)
+from shared.models.linkage import (
+    CVEDerivationClusterProposal,
+    DerivationClusterProposalLink,
+)
+from shared.models.nix_evaluation import NixChannel, NixDerivation, NixEvaluation
+
+
+class Command(BaseCommand):
```
fricklerhandwerk, line 21: Shouldn't we merge/dedup it with https://github.com/NixOS/nix-security-tracker/blob/main/src/shared/management/commands/garbage_collect_derivations.py somehow?

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,282 @@
...
+        #
+        # Step 1 — proposals:  clears events, then DerivationClusterProposalLink rows, satisfying PROTECT on NixDerivation
+        # Step 2 — derivations: clears NixDerivation rows, satisfying PROTECT on NixEvaluation
+        # Step 3 — evaluations: clears NixEvaluation rows, satisfying PROTECT on NixChannel
+        # Step 4 — channels:   clears NixChannel rows
+
+        self._step_stale_proposals(cutoff, batch_size, dry_run)
+        self._step_unmatched_derivations(cutoff, batch_size, dry_run)
+        self._step_empty_evaluations(cutoff, batch_size, dry_run)
+        self._step_inactive_channels(batch_size, dry_run)
```
fricklerhandwerk, line 63: Maybe clearer names?

```suggestion
        self._delete_stale_matches(cutoff, batch_size, dry_run)
        self._delete_unmatched_derivations(cutoff, batch_size, dry_run)
        self._delete_empty_evaluations(cutoff, batch_size, dry_run)
        self._delete_inactive_channels(batch_size, dry_run)
```

src/shared/management/commands/garbage_collect.py
adekoder: i wanted more context the `garbage_collect_derivation` command so as not to assume and merge them
@florentc

src/shared/management/commands/garbage_collect.py
adekoder: we are meant to call them in this order, because of the `PROTECT` Constraints, else it will crash, i implemented it this way by design, so that thing are deleted in order.

fricklerhandwerk: Yes but hard-coding the order (and the number of steps) in the inner scope seems very weird. Stuff will change as the data model evolves. I'd imagine instead something like

```python
steps = [
  self.delete_x,
  self.delete_y,
  ...
]
for i, step in enumerate(steps, start=1):
  print(f"Running step {i}/{len(steps)}") 
  step(*args)
```

That fixes the call order but doesn't pollute the procedures with unrelated information.

(Yes, that doesn't work as is because the arguments may differ, but it's just to illustrate the idea.)

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,276 @@
...
+        """
+        Delete COMPLETED evaluations older than 1 year that have no
+        derivations remaining. PROTECT on NixDerivation.parent_evaluation
+        is satisfied by step 2.
+        """
+        self.stdout.write("\n[3/4] Empty evaluations")
+
+        candidates = NixEvaluation.objects.filter(
+            updated_at__lt=cutoff,
+            state=NixEvaluation.EvaluationState.COMPLETED,
```
fricklerhandwerk: As mentioned in https://github.com/NixOS/nix-security-tracker/issues/1004#issuecomment-4280916122, we may actually not want to delete the evaluations just yet, since we could end up needing the commits associated with their approximate creation timestamp.

```suggestion
            state__in=[
                NixEvaluation.EvaluationState.FAILED,
                NixEvaluation.EvaluationState.CRASHED,
            ],
```

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,276 @@
...
+        )
+
+        self._delete_in_batches(
+            qs=candidates,
+            model=NixDerivation,
+            pk_field="id",
+            label="derivation(s)",
+            batch_size=batch_size,
+            dry_run=dry_run,
+        )
```
fricklerhandwerk, line 157: This will leave `NixDerivationMeta` dangling, IIUC. And as mentioned in the previous review round, we should probably dedup this with the existing, more superficial GC script which does clean up the metadata:

https://github.com/NixOS/nix-security-tracker/blob/4c068d09f2670c2827ccaf48ddc562a34981fe90/src/shared/management/commands/garbage_collect_derivations.py#L41-L43

src/shared/management/commands/garbage_collect.py
adekoder: i don't any issue with caling the function directly, if the `step` statement in the function is an issue i can remove them

src/shared/tests/conftest.py
```diff
@@ -214,6 +222,8 @@ def wrapped(
         if attribute is None:
             attribute = pname
 
+        print("make drv eval-pk", evaluation.pk)
+        print(evaluation.pk, evaluation.updated_at, "make_drv")
```
fricklerhandwerk:
```suggestion
```

src/shared/tests/conftest.py
```diff
@@ -130,19 +132,25 @@ def channel(
...
             channel=channel,
             commit_sha1=secrets.token_hex(16),
             state=state,
         )
 
+        if days > 0:
+            NixEvaluation.objects.filter(pk=evaluation.pk).update(
+                updated_at=timezone.now() - timedelta(days=days)
+            )
+        evaluation.refresh_from_db()
```
fricklerhandwerk:
```suggestion
            evaluation.refresh_from_db()
```

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,332 @@
...
+    )
+
+
+def test_deletes_stale_proposal_no_overlays(
+    make_drv: Callable[..., NixDerivation],
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    """PENDING proposal older than 1 year with no overlays is deleted."""
+    suggestion = make_suggestion()
+    _age_proposal(suggestion)
```
fricklerhandwerk: Why not use the same sort of fixture update on `make_suggestion` as we have it for `make_evaluation`?

```suggestion
def test_deletes_stale_proposal_no_overlays(
    make_drv: Callable[..., NixDerivation],
    make_suggestion: Callable[..., CVEDerivationClusterProposal],
) -> None:
    """PENDING proposal older than 1 year with no overlays is deleted."""
    suggestion = make_suggestion(age=timedelta(days=400))
```

src/shared/tests/conftest.py
```diff
@@ -130,19 +132,25 @@ def channel(
...
+        evaluation = NixEvaluation.objects.create(
             channel=channel,
             commit_sha1=secrets.token_hex(16),
             state=state,
         )
 
+        if days > 0:
+            NixEvaluation.objects.filter(pk=evaluation.pk).update(
+                updated_at=timezone.now() - timedelta(days=days)
+            )
```
fricklerhandwerk: Let's make meaning explicit through types.

```suggestion
        age: datetime.timedelta = timedelta(0),
) -> NixEvaluation:
        evaluation = NixEvaluation.objects.create(
channel=channel,
commit_sha1=secrets.token_hex(16),
state=state,
)

        if age > timedelta(0):
            NixEvaluation.objects.filter(pk=evaluation.pk).update(
                updated_at=timezone.now() - age
            )
```

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,276 @@
...
+        that have no proposal link.
+
+        Deleting a derivation cascades to NixDerivationMeta and all M2M
+        through-tables (maintainers, licenses, source_provenances).
+        NixMaintainer, NixLicense, NixSourceProvenance rows are never deleted.
+        """
+        self.stdout.write("\n[2/4] Unmatched derivations")
+        candidates = NixDerivation.objects.filter(
+            parent_evaluation__updated_at__lt=cutoff,
+            parent_evaluation__state=NixEvaluation.EvaluationState.COMPLETED,
```
fricklerhandwerk:
```suggestion
            parent_evaluation__state__in=[
                NixEvaluation.EvaluationState.COMPLETED,
                NixEvaluation.EvaluationState.FAILED,
                NixEvaluation.EvaluationState.CRASHED,
            ],
```

Since we need to clean those up in any case...

fricklerhandwerk (commented):
Apart from small stylistic nits, what's mainly missing is deduplicating with the existing GC script.

src/shared/management/commands/garbage_collect.py
adekoder: Not sure i understand what you mean, deleting a `NixDerivation` will already cascade to `NixDerivationMeta` 
if you check the model connection the `on_delete` is a `Cascade` and not `Protect` for this one.

src/shared/tests/conftest.py
adekoder: this means everytime one needs to pass age we need to reminder to pass it as a timedelta, why not let the fixtures handle that, we just pass the number of  days and that is all.

src/shared/management/commands/garbage_collect.py
adekoder: but the derivations would have been deleted, is the evaluation still useful after the drvs has been deleted ?

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: I wrote that one as a manual stopgap. It deletes derivations (via their metadata cascading) from failed evaluations, since we're not using them anyway. This new script does more, but we should definitely do both in one go.

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: Yes, I think it's an issue because every bit of information that requires change when unrelated things change will pile up work for the future. Separation of concerns, minimality principle, etc.

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: Yes, as discussed in that linked issue. We don't have a cheap-and-reliable way to associate commit hashes with timestamps or even order, and our time-of-evaluation is a reasonable approximation for both that we're already doing. Also the empty evaluation shells and channels barely consume any space, so it's ~free to keep them around indefinitely.

src/shared/tests/conftest.py
fricklerhandwerk: Because types embed contextual information that one cannot interpret incorrectly. Textual labels may or not be apt, and it requires interpretative effort by author and reader to get it right, so it means there's an opportunity for error. But structural properties encoded in the program -- in this case, inputs satisfying type constraints -- either hold or not. Since fixing bugs is order of magnitudes more expensive than preventing them, strong typing is a software engineering best practice in my book, and thus not optional.

src/shared/management/commands/garbage_collect.py
adekoder: noted

src/shared/management/commands/garbage_collect.py
```diff
@@ -0,0 +1,316 @@
...
+        Delete derivations from COMPLETED, FAILED and CRASHED evaluations older than 1 year
+        that have no proposal link.
+
+        M2M through-table rows (maintainers, licenses, source_provenances) are removed
+        automatically when their derivation or meta is deleted.
+        NixMaintainer, NixLicense, NixSourceProvenance lookup rows are never deleted.
+        """
+        self.stdout.write("\n[2/4] deleting unmatched derivations")
+
+        failed_crashed = NixDerivation.objects.filter(
```
adekoder, line 141: so we don't want the cut off date for this ?

src/shared/tests/test_garbage_collect.py
```diff
@@ -0,0 +1,312 @@
...
+    )
+
+    call_command("garbage_collect", stdout=StringIO())
+
+    assert CVEDerivationClusterProposal.objects.count() == remaining_suggestions
+    assert DerivationClusterProposalLink.objects.count() == remaining_links
+
+
+@pytest.fixture
+def make_overlay(
```
adekoder: why the need for making this a fixture when it only going to be used one one testcase

adekoder: i understand you want to fold the test but this seem like too much for what it is

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: Yes, we can throw those away immediately (and we did in the original script).

src/shared/tests/test_garbage_collect.py
fricklerhandwerk: Because there's no other way to run an arbitrary procedure "as a parameter". But for the purpose of the test that's exactly what we want: do one of the things we consider "user input" and check that we don't GC in that case.

But I see what you mean, we can also do that inline.

fricklerhandwerk:
I know, I made substantial changes to the tests, but I needed to convince myself very firmly that we're doing the right thing there. That was my way of making sure I understand the tests correctly, and that they have enough coverage of the interesting cases - it's still not perfect, but I reasonably trust it now, ran it locally, and would be ready to deploy this to production.

fricklerhandwerk (approved):

PR #998, opened 2026-04-17
refactor: extract oneshot for migrations

fricklerhandwerk:
Since multiple services should get going concurrently after migrations are in place, we don't want them to wait for the web service.
This makes a negligble performance difference in practice; the refactoring is primarily intended to clarify the dependency structure.


Erethon:
Since #1000 was merged this PR can be closed, right?

PR #1003, opened 2026-04-20
chore: npins update

fricklerhandwerk:


fricklerhandwerk:
@Erethon do you have any clue why -- apparently -- the Nix build for vale fails due to lack of CA certs of all things??

There's some recent update where certs were discussed: https://github.com/NixOS/nixpkgs/pull/490484

Issue #875, opened 2026-03-16
Show automatic dismissal in Activity log


fricklerhandwerk:
Follow-up on https://github.com/NixOS/nix-security-tracker/issues/804

## Acceptance criteria

Given a suggestion that is automatically dismissed,
When I view its activity log,
Then the automatic dismissal and its reason is listed as such.

Given a suggestion that is automatically dismissed,
When I view it,
Then the automatic dismissal reason is shown prominently.

florentc:
Since #888 has been merged, the dismissal reasons are shown. Here is an example from production:

<img width="2233" height="464" alt="Image" src="https://github.com/user-attachments/assets/dc2d4309-4147-4112-b2da-69da52ca5a67" />

The second acceptance criterion is already done.

As for the first one, there are two possibilities:
1. Nothing. "exclusively hosted service" is a dismissal reason that only sec tracker can assign. Seeing it on a dismissed suggestion means automatic dismissal. Showing it in the activity log would be redundant. Unless this suggestion is eventually "Accepted" anyway, in which case maybe we want to see it was originally automatically dismissed.
2. It turns out the "Created on" events in the activity logs are fake. There are no events in database for creation. Therefore implementing criterion 1 would need to implement actual events for creation. It should be doable as we already track `status` and `rejection_reason` to create the status change events.

fricklerhandwerk:
Yes, the idea behind criterion 1 was exactly that users may end up accepting something again. But we could also simply prevent that to avoid gaps in the activity log, assuming that automatic dismissal is always correct and irrevocable?

PR #1007, opened 2026-04-21
feat: Events for creation (incl. auto dismiss reason) in activity log

florentc:
This addresses #875 

In activity log, creation events used to be fake. We were hardwriting "Created automatic suggestion" and using the creation date of the suggestion.

This PR introduces real events for suggestion creation in the activity log data structure.

These events have an optional `rejection_reason` in case the events are created as already dismissed.

In that case, the reason is displayed in the activity log.

The text for creation event is shortened as it makes it more obvious when an automatic rejection happened, and it avoids display glitches when the line gets too long with the date & rejection reason.

Creation events are the first events without an attached user. It is now possible to have other events without a user. Type wise, this may have to be reworked if this becomes more than an anecdotal need.

### DIFF
--- a/src/webview/templates/components/suggestion_activity_log.html
+++ b/src/webview/templates/components/suggestion_activity_log.html
@@ -9,24 +9,30 @@
     <summary>
       <span class="details-closed">
         {% with last_entry=activity_log|last %}
-          <span class="activity-log-timestamp">updated {{ last_entry.timestamp|naturaltime }}</span>
-          by&nbsp;<strong>@{{ last_entry.username }}</strong>
+          <span class="activity-log-timestamp">
+            {% if last_entry.action == "create" %}
+              created
+            {% else %}
+              updated
+            {% endif %}
+          {{ last_entry.timestamp|naturaltime }}</span>
+          {% if last_entry.username %}by&nbsp;<strong>@{{ last_entry.username }}</strong>{% endif %}
         {% endwith %}
       </span>
       <strong class="details-open">Activity log</strong>
     </summary>
     <ul class="column align-end">
-      <li>
-        Created <span class="automatic-suggestion-text">automatic suggestion</span>
-        <span class="activity-log-entry-timestamp"
-              data-timestamp-relative="{{ suggestion.created_at|naturaltime }}"
-              data-timestamp-iso="{{ suggestion.created_at|iso }}">{{ suggestion.created_at|naturaltime }}</span>
-      </li>
       {% for log_entry in activity_log %}
         <li class="row gap-small baseline">
-          <strong>@{{ log_entry.username }}</strong>
+          {% if log_entry.username %}<strong>@{{ log_entry.username }}</strong>{% endif %}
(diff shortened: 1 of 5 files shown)


src/webview/tests/test_activity_log.py
```diff
@@ -168,7 +168,7 @@ def test_restore_maintainer_cancels_event_in_activity_log(
         expect(removed_maintainer).to_be_visible()
         expect(added_maintainer).to_be_visible()
     else:
-        expect(activity_log).to_have_count(0)
+        expect(activity_log).to_have_count(1) # 1 being the creation event
```
fricklerhandwerk: ❤️

src/webview/templates/components/suggestion_activity_log.html
```diff
@@ -9,24 +9,30 @@
...
-      </li>
       {% for log_entry in activity_log %}
         <li class="row gap-small baseline">
-          <strong>@{{ log_entry.username }}</strong>
+          {% if log_entry.username %}<strong>@{{ log_entry.username }}</strong>{% endif %}
           {# TODO(@adekoder) Refactor: the text value should come from the activity log context to aviod hardcoding #}
-          {% if "package." in log_entry.action %}
+          {% if log_entry.action == "create" %}
+            {% if log_entry.rejection_reason %}
+              Created & dismissed ({{ log_entry.rejection_reason }}) suggestion
```
fricklerhandwerk, line 31: Something about the wording rubs me the wrong way, but I don't have anything better to suggest. It's also technically correct that the suggestion is created and instantly dismissed. Probably what trips me up is the fact that we're kind of making a category error with this sort of display (which has nothing to do with this PR). We do want to record that the system looked at the CVE and performed some automatic action. But that doesn't necessarily mean it needs to show up in "dismissed", as that list is intended to be a record of human data entry. This is something to think about for later... we probably don't want to display the auto-rejected ones by default; and for non-matches we don't want to throw away the fact that we did inspect the CVE.

fricklerhandwerk (approved):

src/webview/templates/components/suggestion_activity_log.html
florentc: Haha you changed your mind on this. When you introduced automatic dismissal we had a discussion and I was opposed to having it show up in "Dismissed" while you were fine with it.

Now, I think I have also changed my mind on it, so that the invariant "we should always disagree" holds :laughing: I think it's fine that it shows up in dismissed.

Recently we have been discussing about future features such as marking automatically (not just with manual intervention) some CVE as fixed in given branches for given packages. It seems to me we are considering sec tracker to no longer follow a "the machine creates, the users mutate" model but a "both machine and users may mutate and it's all archived in activity log". As a user who browses the dismissed list, it doesn't make much difference to see the dismissal has been done by another human or by sec tracker itself, as long as both are registered in the activity log.

Besides, we have been talking about eventually removing the concept of explicit suggestion status.

PR #1010, opened 2026-04-23
perf: only evaluate one system

fricklerhandwerk:
This may lose some coverage, as it won't catch packages that only exist on the other platforms.
But it reduces evaluation time and storage consuption to 1/4.

@NixOS/security would that constitute a notable negative impact for your purposes?

### DIFF
--- a/src/shared/listeners/nix_evaluation.py
+++ b/src/shared/listeners/nix_evaluation.py
@@ -52,8 +52,17 @@ async def perform_evaluation(
     the execution runtime.
     """
     # TODO(raitobezarius): bring a Nix code generator, that'd be cuter.
-    nixpkgs_config = "{ config = { allowUnfree = true; inHydra = false; allowInsecurePredicate = (_: true); scrubJobs = false; }; };"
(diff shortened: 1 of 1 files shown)


LeSuisse:
> would that constitute a notable negative impact for your purposes?

Seems like a fair trade. We might lose a bit of coverage on specific Darwin software but I suppose that's reasonable. The only thing I have in mind is `iterm2`. Do you have the ability to check if for the published issues we have others that were only matched on Darwin aarch64?

fricklerhandwerk:
Darwin-only packages from published issues in production:
- `emacs-macport`
- `emacs30-macport`
- `ghostty-bin`
- `pam-reattach`
- `pam-watchid`
- `python312Packages.mitmproxy-macos`
- `python313Packages.mitmproxy-macos`
- `python314Packages.mitmproxy-macos`
- `swiftlint`
- `tests.fetchNextcloudApp.simple-sha512`
- `tests.fetchpatch.relative`
- `tests.fetchpatch2.relative`

So we can't meaningfully *not* evaluate for Darwin...

And looking at all packages, there's no strict superset of all the platforms. They have non-overlapping portions at least in the hundreds.

I guess the only way for now is to deduplicate the derivation metadata by a lot to save the space, and make evaluation results asynchronously available.

PR #1017, opened 2026-04-29
feat: Email notifications

florentc:
- control panel in subscriptions page to opt in and select mail address
- simplification of pgpubsub channels (now notification in sequence with caching)
- emails mention affected packages, matching packages, and useful links
- 1-1 correspondance between email and notification

In the above screenshot my user has no github email attached, when you have one, it says "Use default GitHub email" instead of "Clear email"

2 notes:
- now that notification is done after caching in sequence, we have access to cached info when notifying. It was not the case before so in `create_package_subscription_notifications` we could optimize by using cached info instead of performing redundant db queries
- email notification assumes we only have notifications about suggestions. It is true in practice for new suggestions. Text notifications are old notifications that some people still have and they won't trigger an email anyway, but still we may want to handle the various types of notifications in the future


src/shared/notify_users.py
fricklerhandwerk, line 1: Now technically this is not a listener any more. Why not keep it in one file `cache_and_notify`? Because I wouldn't know where to put this logic separately.

src/webview/models.py
```diff
@@ -92,6 +93,27 @@ class Profile(models.Model):
         default=True,
         help_text="Automatically subscribe to notifications for packages this user maintains",
     )
+    notification_email = models.EmailField(
+        blank=True,
```
fricklerhandwerk, line 97: We should probably model "not set" rigorously. `blank=True` is only about form representations of the model, and will skip the email regex validator if the input is empty.

```suggestion
        blank=True,
        null=True,
```

src/webview/models.py
```diff
@@ -101,7 +123,6 @@ def create_notification(
             user=self.user,
             suggestion=suggestion,
         )
-
```
fricklerhandwerk, line 104: Stray change?

src/webview/subscriptions/views.py
```diff
@@ -181,6 +181,89 @@ def _handle_error(self, request: HttpRequest, error_message: str) -> HttpRespons
...
+
+
+class SetNotificationEmailView(LoginRequiredMixin, TemplateView):
+    """Set email used for notifications"""
+
+    template_name = "subscriptions/components/email_setter.html"
+
+    def post(self, request: HttpRequest) -> HttpResponse:
+        """Toggle auto-subscription setting."""
+        email = request.POST.get("email", "")
```
fricklerhandwerk, line 233: With the model change to make it nullable:
```suggestion
        email = request.POST.get("email", None)
```

src/webview/subscriptions/views.py
```diff
@@ -181,6 +181,89 @@ def _handle_error(self, request: HttpRequest, error_message: str) -> HttpRespons
...
+        profile = request.user.profile
+        profile.notification_email = email
+        profile.save(update_fields=["notification_email"])
+
+        # Handle HTMX vs standard request
+        if request.headers.get("HX-Request"):
+            return self.render_to_response(
+                {
+                    "notification_email": profile.notification_email,
+                    "github_email": request.user.profile.github_email,
```
fricklerhandwerk: Oh interesting, the template tag expects this to be a string, but it could be `None` as well.

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+        user.profile.notification_email
+        if user.profile.notification_email
+        else user.profile.github_email
+    )
+
+    if not email:
+        logger.info(f"Could not send email notification to {user.username}: no address")
+        return
+
+    # Reuse existing NotificationContext logic for package matching
```
fricklerhandwerk:
```suggestion
    # Reuse existing notification rendering logic
```

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+    context = NotificationContext(notification=notification, user_profile=user.profile)
+
+    suggestion = notification.suggestion
+    subject = f"New security notification: {suggestion.cve.cve_id}"
+
+    message_parts = [
+        "Hello,",
+        "",
+        "A new security vulnerability has been identified that may affect packages you follow or maintain:",
```
fricklerhandwerk:
```suggestion
        "A vulnerability may affect packages you follow or maintain:",
```

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+        "",
+        f"CVE: {suggestion.cve.cve_id}",
+        f"Details: https://nvd.nist.gov/vuln/detail/{suggestion.cve.cve_id}",
```
fricklerhandwerk:
```suggestion
        f"Details: {urljoin(str(settings.BASE_URL), reverse('webview:suggestion:detail', kwargs={'suggestion_id': suggestion.pk}))}",
```

We're linking to upstream in the match, and it can be expected to be less useful than our presentation (primarily because we're showing the Nixpkgs match).

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
         )
+
+    message_parts.extend(
+        [
+            f"View all notifications: {urljoin(str(settings.BASE_URL), reverse('webview:notifications:center'))}",
+            f"Review suggestion details: {urljoin(str(settings.BASE_URL), reverse('webview:suggestion:detail', kwargs={'suggestion_id': suggestion.pk}))}",
+            "",
+            "Best regards,",
+            "Nix Security Tracker",
+            "",
```
fricklerhandwerk:
```suggestion
```

What we may want to have though is a link to the source or issue tracker, so people can complain to us directly if something sucks.

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+            f"Manage email notification preferences: {urljoin(str(settings.BASE_URL), reverse('webview:subscriptions:center'))}",
+        ]
+    )
+
+    message = "\n".join(message_parts)
+
+    send_mail(
+        subject=subject,
+        message=message,
```
fricklerhandwerk:
```suggestion
    send_mail(
        subject=subject,
        message="\n".join(message_parts),
```

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+        from_email=None,  # Use default
+        recipient_list=[email],
+        fail_silently=False,
+    )
+    logger.info(f"Email notification sent to {user.username} at {email}")
```
fricklerhandwerk:
```suggestion
    logger.info(f"Email notification sent to {user.username}")
```

Actually we should not log any user data. Sure it'd be good to know if a given notificaton was sent as an email, but then why not attach a flag to the notification data model, for example?

src/webview/subscriptions/views.py
```diff
@@ -181,6 +181,89 @@ def _handle_error(self, request: HttpRequest, error_message: str) -> HttpRespons
...
+            return redirect(reverse("webview:subscriptions:center"))
+
+
+class SetNotificationEmailView(LoginRequiredMixin, TemplateView):
+    """Set email used for notifications"""
+
+    template_name = "subscriptions/components/email_setter.html"
+
+    def post(self, request: HttpRequest) -> HttpResponse:
+        """Toggle auto-subscription setting."""
```
fricklerhandwerk:
```suggestion
```

src/shared/notify_users.py
```diff
@@ -19,6 +22,10 @@ def create_package_subscription_notifications(
     and for maintainers of those packages (if they have auto-subscribe enabled).
     """
 
+    # FIXME(@florentc): These queries are no longer related. We should use the
```
fricklerhandwerk, line 26: Maybe we should make that explicit in the signature then, and pass a `CachedSuggestion`?

src/shared/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+    if not (user.profile.receive_email_notifications):
+        return
+
+    email = (
+        user.profile.notification_email
+        if user.profile.notification_email
+        else user.profile.github_email
+    )
+
+    if not email:
```
fricklerhandwerk, line 113: Wouldn't it be nice if this was impossible by construction? Could be worth a...

```suggestioN
    # FIXME(@fricklerhandwerk): Make this case impossible by construction.
    # Only allow enabling email notifications if an address is set, and make removing an address disable emails atomically.
```

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+
+    if not email:
+        logger.info(f"Could not send email notification to {user.username}: no address")
+        return
+
+    # Reuse existing NotificationContext logic for package matching
+    context = NotificationContext(notification=notification, user_profile=user.profile)
+
+    suggestion = notification.suggestion
+    subject = f"New security notification: {suggestion.cve.cve_id}"
```
fricklerhandwerk:
```suggestion
    subject = f"Nixpkgs security notification: {suggestion.cve.cve_id}"
```

src/shared/listeners/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+
+    message_parts = [
+        "Hello,",
```
fricklerhandwerk:
```suggestion
        f"Hello @{user.username},",
```

src/shared/notify_users.py
```diff
@@ -77,21 +84,113 @@ def create_package_subscription_notifications(
...
+        "Hello,",
+        "",
+        "A new security vulnerability has been identified that may affect packages you follow or maintain:",
+        "",
+        f"CVE: {suggestion.cve.cve_id}",
+        f"Details: https://nvd.nist.gov/vuln/detail/{suggestion.cve.cve_id}",
+        "",
+    ]
+
+    if suggestion.cached and suggestion.cached.payload.get("affected_products"):
```
fricklerhandwerk, line 140: We could assume suggestions are cached by passing a `CachedSuggestion`:

```suggestion
    if suggestion.cached.payload.get("affected_products"):
```

fricklerhandwerk (approved):
Nice! Only some cosmetic nits.

src/webview/models.py
florentc: I remember checking about this online and reading that the idiomatic way in django was to use blank. At first I had done like you suggest.

EDIT : Found it 

https://docs.djangoproject.com/en/6.0/ref/models/fields/

> Avoid using [null](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.Field.null) on string-based fields such as [CharField](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.CharField) and [TextField](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.TextField). The Django convention is to use an empty string, not NULL, as the “no data” state for string-based fields. If a string-based field has null=False, empty strings can still be saved for “no data”. If a string-based field has null=True, that means it has two possible values for “no data”: NULL, and the empty string. In most cases, it’s redundant to have two possible values for “no data”. One exception is when a [CharField](https://docs.djangoproject.com/en/6.0/ref/models/fields/#django.db.models.CharField) has both unique=True and blank=True set. In this situation, null=True is required to avoid unique constraint violations when saving multiple objects with blank values.

src/shared/notify_users.py
florentc: It's just a remnant of when then were different listeners. I kept it separated out of readability. If I put this in `cache_and_notify.py`, I also have to put `cache_suggestions.py` there as well if the idea is "one listener - one module". Both `notify_users` and `cache_suggestions` are big already and different concerns. This would become a huge module. This is why I compromised with the current solution.

What do you think?

src/webview/models.py
florentc: Yes

src/shared/listeners/notify_users.py
florentc: The URL to the suggestion is already at the end of the mail.

If we remove the link to upstream CVE, I think we should also remove the CVE id. Maybe we should display at least the title of the CVE then?

src/shared/listeners/notify_users.py
florentc: 1. You don't want the signature of the mail anymore or it's accidental in this diff?
2. What do you mean by the source? Those mails are just supposed to be clones/sum ups of the notifications that people see on security tracker, I don't see such a thing in notifications now.

src/shared/listeners/notify_users.py
florentc: `username` is not user data? Should I use user.pk maybe instead if we really want no user data in logs.
EDIT: I used `user.pk`

As for a flag in the notification model it was not the idea behind this simple degug message. This debug message is just there to very very quickly debug the listeners at first glance in your local dev console and experiment.

A flag in the data model would be less convenient for debugging on the go, and it would not help to debug issues like mails being triggered twice for instance (arguably you could have a counter instead of flag but it's starting to be overkill IMO)

src/webview/subscriptions/views.py
florentc: Good catch

src/shared/notify_users.py
florentc: Not opposed to the idea. Note however that there are many places in our codebase where we require cached suggestions and we pass suggestions.

src/shared/notify_users.py
florentc: Yes I thought so at first but it'd not be great in fact:

Here `email` is either your explicit notification email, or your github email if you don't have one, or nothing if you don't even have a github email.

What you describe would only be for users without a GitHub email. The "clear email" button is a "use github email" button for people with a github email attached to their profile. For them, there is no concept of "removing an email address".

I think it's clearer to keep the notification toggle and the email address two separate things.

src/shared/listeners/notify_users.py
florentc: Ok if you want. A user would already know it's from sec tracker from the sender email address through.

src/shared/listeners/notify_users.py
florentc: Is `username` reliable? If they change their github handle won't it be outdated?

src/shared/listeners/notify_users.py
fricklerhandwerk: Yes, we should display the title in the message body.

src/shared/listeners/notify_users.py
fricklerhandwerk: Yes I'd remove the signature.

With email the context is less obvious though, but it's fine not to add more stuff.

src/shared/listeners/notify_users.py
fricklerhandwerk: Yeah it was more like a lateral thought, not immediately important for the user story. It's just that with such logs the best we could do is inspect single cases on request, but wouldn't realistically notice if something is off.

But in any case, we should probably not log emails.

src/shared/notify_users.py
fricklerhandwerk: We don't need to fix everything at once, just the parts we're touching now.

src/shared/notify_users.py
fricklerhandwerk: I see. Can we still somehow consolidate it with the confusion that the implementation of handlers are not listeners and thus they shouldn't live under `/listeners/`? Maybe should the business logic move one directory up, fitting the concern it's serving?

src/shared/listeners/notify_users.py
fricklerhandwerk: Ah yes, of course it should be the handle...

src/webview/templates/subscriptions/components/email_setter.html
```diff
@@ -0,0 +1,60 @@
+{% load viewutils %}
+
+<div class="email-setter column gap">
+  <div>
+    {% if notification_email %}
+      Current notification email: <span class="bold">{{ notification_email }}
+    {% elif github_email %}
+      Current notification email: default GitHub email <span class="bold">{{ github_email }}</span>
```
fricklerhandwerk:
```suggestion
      Current notification email: Nixpkgs maintainer email <span class="bold">{{ github_email }}</span>
```

It's not their GH account email,
but that thing from maintainers.nix

src/webview/templates/subscriptions/components/email_setter.html
```diff
@@ -0,0 +1,60 @@
...
+            hx-target="closest .email-setter"
+            hx-swap="outerHTML"
+            hx-indicator="find button"
+            class="row">
+        <input type="hidden" name="email" value="">
+      </input>
+      {% csrf_token %}
+      <button type="submit" class="btn btn-gray">
+        {% if github_email %}
+          Use default GitHub email
```
fricklerhandwerk:
```suggestion
          Use Nixpkgs maintainer email
```

We may want to eventually be more precise about that and maybe link to Nixpkgs docs...

src/shared/notify_users.py
fricklerhandwerk: Yeah I see, not that important in any case.

src/shared/notify_users.py
florentc: In fact we pass the notification itself in the email creation function (we need the info in the notification especially what matches the user subscriptions) and the notification object has already the suggestion. It would be redundant. I'll add a comment

src/webview/templates/subscriptions/components/email_setter.html
florentc: Oh yes right. I need to change the property name in model and across the PR too then

src/webview/models.py
florentc: Formatting refuses to let me reintroduce the blank line

src/shared/listeners/notify_users.py
florentc: Now looks like this

```
Subject: Nixpkgs security notification: CVE-2026-13272
From: 
To: test@test.com
Date: Mon, 04 May 2026 12:02:47 -0000
Message-ID: <177789616724.76516.11222844749367578166@skaro>

Hello florentc,

A vulnerability may affect packages you follow or maintain:

CVE: CVE-2026-13272
Object: Vulnerability in xob
Details: http://localhost:8000/suggestions/by-id/72259/

Affected products:
  - xob: *

Packages you maintain that may be affected:
  - xob

View all notifications: http://localhost:8000/notifications/
Manage email notification preferences: http://localhost:8000/subscriptions/
```

src/shared/notify_users.py
florentc: I moved `cache_suggestions.py` and `notify_users.py` to root of `shared`

There is a dedicated commit as it's a refactor that changes a lot of imports everywhere

fricklerhandwerk:
Feel free to rewrite history appropriately and merge when green.

PR #1018, opened 2026-04-29
feat: Add support for multiple matching algorithm versions

adekoder:
Algorithm versioning for CVE-to-derivation suggestions
Each `CVEDerivationClusterProposal` now records the algorithm version that produced it. The current version is defined as `CURRENT_ALGORITHM_VERSION` on the model, contributors bump it when matching logic changes.

This lays the groundwork for benchmarking algorithm versions before deployment: the untriaged queue filters to the active version only, while accepted/dismissed suggestions remain visible regardless of which version created them, so users never lose access to work they've already reviewed.


src/project/settings.py
```diff
@@ -169,6 +169,25 @@ class DjangoSettings(BaseModel):
             """,
             default=1_000,
         )
+        ACTIVE_MATCHING_ALGORITHM_VERSION: int = Field(
+            description="""
+            Controls which registered matching algorithm version is used when
+            linking CVEs to derivations. Must match a VERSION defined in
+            shared/listeners/algorithms/. Bump this setting to activate a new
+            algorithm version without changing code.
```
fricklerhandwerk: This doesn't work. Changing an environment variable for us is exactly equivalent to changing code, because there's no such thing as a deployment that doesn't touch the source.

That means, for our purposes at this point in development it would just as good (and a lot simpler) to hard-code it in e.g. the suggestion model like we have it for caching (and as discussed in https://github.com/NixOS/nix-security-tracker/issues/722#issuecomment-4312878481):

https://github.com/NixOS/nix-security-tracker/blob/e3ead575d9a0732c399b8166172586b765677e8d/src/shared/models/cached.py#L21-L23

src/shared/listeners/algorithms/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def build_new_links(self, container: Container) -> bool: ...
+
+
+_registry: dict[int, MatchingAlgorithm] = {}
+
+
+def register(module: MatchingAlgorithm) -> None:
+    """Register an algorithm module under its VERSION."""
+    _registry[module.VERSION] = module
```
fricklerhandwerk: Can't we just read the directory and map the `v*.py` filenames to versions automatically?

src/shared/tests/test_linkage.py
```diff
@@ -19,6 +20,7 @@
 )
 
 
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=1)
```
fricklerhandwerk: Here is where this design via settings already falls apart. Once we remove v1 this test will need to change

src/shared/listeners/algorithms/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+
+
+def _resolve(version: int) -> MatchingAlgorithm:
+    """Return a resigister alogirithm that match the version."""
+    if not _registry:
+        raise RuntimeError("No matching algorithm registered.")
+    try:
+        return _registry[version]
+    except KeyError:
+        raise KeyError(f"No matching algorithm registered for version {version}.")
```
fricklerhandwerk: It would be perfectly enough to check in a test that selecting the active and candidate version works, simply to guard against typos and omissions. Since for now we're controlling the selection in code, all we need to guarantee is basic consistency.

Then the actual selection is simply `_registry[version]` right there at the call site.

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,143 @@
...
+    status: CVEDerivationClusterProposal.Status,
+) -> None:
+    """Active-version proposal is visible; same-status candidate-version proposal is not."""
+    drv = make_drv()
+
+    active_container = make_container(cve_id="CVE-2025-1001")
+    active_proposal = make_suggestion(
+        container=active_container,
+        drvs={drv: ProvenanceFlags.PACKAGE_NAME_MATCH},
+        status=status,
```
fricklerhandwerk: This should be more explicit and independent of any particular value, something like:
```
  algorithm_version=CVEDerivationClusterProposal.AlgorithmVersion
```
and then in the counterpart
```
  algorithm_version=CVEDerivationClusterProposal.AlgorithmVersion+1
```

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,143 @@
...
+        "webview:suggestion:suggestions_by_package",
+        kwargs={"package_name": shared_drv.attribute},
+    )
+    as_staff.goto(live_server.url + url)
+
+    expect(as_staff.locator(f"#suggestion-{active_proposal.pk}")).to_be_visible()
+    expect(as_staff.locator(f"#suggestion-{candidate_proposal.pk}")).not_to_be_visible()
+
+
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=2)
```
fricklerhandwerk: I don't think we need to test this case. We already have it in the business logic test, and the UI tests above make sure the filter is used at all.

src/project/settings.py
```diff
@@ -169,6 +169,25 @@ class DjangoSettings(BaseModel):
...
+        CANDIDATE_MATCHING_ALGORITHM_VERSION: int | None = Field(
+            description="""
+            Optional. When set, identifies a new algorithm version being evaluated
+            in parallel. The candidate version does not run automatically — it is
+            only invoked by the test-run management command, which generates proposals
+            tagged with this version number for later metric comparison.
+            Set to None when no candidate is under evaluation.
+            """,
+            default=None,
+        )
```
fricklerhandwerk: I don't think we actually need that parameter (regardless of where it lives). Arguably we could just dry-run all non-active versions above the active one. We can control the number of them by putting them in that directory or not.

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,117 @@
...
+
+    assert proposal in qs
+
+
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=1)
+def test_active_excludes_proposal_with_different_version(
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    """A proposal from a candidate (non-active) algorithm is hidden from .active()."""
+    candidate_proposal = make_suggestion(algorithm_version=2)
```
fricklerhandwerk: This can be done via parameter in the above test

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,117 @@
...
+    assert qs.count() == 1
+
+
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=1)
+def test_all_returns_proposals_of_any_version(
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+    make_drv: Callable[..., NixDerivation],
+    make_container: Callable[..., Container],
+) -> None:
+    """.objects.all() is unrestricted and returns every proposal regardless of version."""
```
fricklerhandwerk: I don't think we need to test that.

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,117 @@
...
+    """A proposal from a candidate (non-active) algorithm is hidden from .active()."""
+    candidate_proposal = make_suggestion(algorithm_version=2)
+
+    qs = CVEDerivationClusterProposal.objects.active()
+
+    assert candidate_proposal not in qs
+
+
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=1)
+def test_active_returns_only_matching_version_when_both_exist(
```
fricklerhandwerk: This is actually the same as the above, no? It just creates both versions simultaneously rather than in two different tests. Maybe let's only leave this one?

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,117 @@
...
+    assert v2_proposal in all_qs
+    assert all_qs.count() == 2
+
+
+def test_active_respects_setting_change(
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+    make_drv: Callable[..., NixDerivation],
+    make_container: Callable[..., Container],
+) -> None:
+    """Changing ACTIVE_MATCHING_ALGORITHM_VERSION switches which proposals .active() returns."""
```
fricklerhandwerk: I'm inclined to suggest to also fold this into the above test.

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,117 @@
...
+        qs_v2_active = CVEDerivationClusterProposal.objects.active()
+        assert v2_proposal in qs_v2_active
+        assert v1_proposal not in qs_v2_active
+
+
+@override_settings(ACTIVE_MATCHING_ALGORITHM_VERSION=99)
+def test_active_returns_empty_when_no_version_matches(
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    """When no proposal has the active version, .active() is empty."""
```
fricklerhandwerk: This could also be a parameter

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,143 @@
...
+    as_staff: Page,
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+    make_drv: Callable[..., NixDerivation],
+    make_container: Callable[..., Container],
+) -> None:
+    """The by-package view also filters by active algorithm version."""
+    shared_drv = make_drv(pname="shared-pkg", attribute="shared-pkg")
+
+    active_container = make_container(cve_id="CVE-2025-2001")
+    active_proposal = make_suggestion(
```
fricklerhandwerk:
```suggestion
    active_proposal = make_cached_suggestion(
```

Then no need to cache afterwards

fricklerhandwerk (commented):
A simpler version config, import, and selection mechanism would substantially reduce the amount of testing required.

Then the next PRs could add the mechanism for dry-running the non-active versions and collecting metrics, respectively.

Side note: While this follows the design we discussed, I'm a bit uneasy about the algorithm versioning living in separate files. It makes diffing rather impractical. I don't have a better proposal than doing version bumps in multiple commits though: copy v1, change v1; review would look at commits separately.

src/shared/listeners/cache_and_notify.py
```diff
@@ -15,10 +15,12 @@ def cache_new_suggestions_following_new_container(
...
 ) -> None:
     logger.info(f"Cache and notify for suggestion {new.pk}")
-    cache_new_suggestions(new)
-    try:
-        create_package_subscription_notifications(new)
-    except Exception as e:
-        logger.error(
-            f"Failed to create package subscription notifications for suggestion {new.pk}: {e}"
-        )
+    # let generate cache suggestion for only current algorithm
```
fricklerhandwerk:
```suggestion
    # Generate cached value for only current algorithm
```

src/shared/models/linkage.py
```diff
@@ -75,6 +88,11 @@ class RejectionReason(models.TextChoices):
         help_text=_("Reason for rejection (automatic or manual)"),
     )
 
+    algorithm_version = models.IntegerField(
+        default=1,
```
fricklerhandwerk: Taste, but why not 0?

```suggestion
        default=0,
```

src/shared/tests/conftest.py
```diff
@@ -258,11 +258,15 @@ def wrapped(
         status: CVEDerivationClusterProposal.Status = CVEDerivationClusterProposal.Status.PENDING,
         rejection_reason: CVEDerivationClusterProposal.RejectionReason | None = None,
         age: timedelta = timedelta(0),
+        algorithm_version: int | None = None,
```
fricklerhandwerk, line 290: Why not default value?

```suggestion
        algorithm_version: int = CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION,
```

src/shared/management/commands/regenerate_cached_suggestions.py
```diff
@@ -28,18 +28,26 @@ def handle(self, *args: Any, **options: Any) -> None:
             label = "all"
             deleted_count, _ = CachedSuggestions.objects.all().delete()
             self.stdout.write(f"Purged {deleted_count} cached suggestion(s).")
-            proposals = CVEDerivationClusterProposal.objects.order_by(
-                "-updated_at"
-            ).iterator()
+            proposals = (
+                CVEDerivationClusterProposal.objects.active()
```
fricklerhandwerk: This will not regenerate the cache for e.g. accepted suggestions that were created with an older algorithm. I think that filter over active only must be applied to PENDING suggestions.

That will actually still leave a gap: pending suggestions could be user-edited already, even if it's discouraged via UI. We won't solve that problem now, but probably worth a `FIXME`.

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,40 @@
...
+) -> None:
+    """With proposals from two versions in the DB, .active() returns only the current one."""
+    current_proposal = make_suggestion()
+
+    other_container = make_container(cve_id="CVE-2025-9999")
+    other_drv = make_drv(pname="other-pkg", attribute="other-pkg")
+    outdated_proposal = make_suggestion(
+        container=other_container,
+        drvs={other_drv: ProvenanceFlags.PACKAGE_NAME_MATCH},
+        algorithm_version=0,  # any version lower than CURRENT_ALGORITHM_VERSION
```
fricklerhandwerk: Let's be explicit then

```suggestion
        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION-1,
```

src/webview/suggestions/views/lists.py
```diff
@@ -55,7 +55,8 @@ def get_queryset(self) -> QuerySet[CVEDerivationClusterProposal]:
         if self.package_filter is not None:
             query_filters &= Q(cached__payload__packages__has_key=self.package_filter)
         return (
-            CVEDerivationClusterProposal.objects.select_related(
+            CVEDerivationClusterProposal.objects.active()
```
fricklerhandwerk: This can't work for anything but pending suggestions that have already been regenerated with a new version.

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,94 @@
...
+from playwright.sync_api import Page, expect
+from pytest_django.live_server_helper import LiveServer
+
+from shared.models.cve import Container
+from shared.models.linkage import CVEDerivationClusterProposal, ProvenanceFlags
+from shared.models.nix_evaluation import NixDerivation
+
+
+@pytest.mark.parametrize(
+    "url_path, status",
```
fricklerhandwerk: This test is wrong. Anything but pending suggestions with older algo versions need to be shown, so there needs to be a `count_visible` parameter.

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,94 @@
...
+        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION - 1,  # type: ignore
+    )
+
+    as_staff.goto(live_server.url + url_path)
+
+    expect(as_staff.locator(f"#suggestion-{current_proposal.pk}")).to_be_visible()
+    expect(as_staff.locator(f"#suggestion-{outdated_proposal.pk}")).not_to_be_visible()
+
+
+def test_by_package_view_shows_active_version_only(
```
fricklerhandwerk: This would be a bit expensive, but arguably this should play through the same parameters as above. Or maybe there's a cleverer strategy to cover that view (implicitly)?

fricklerhandwerk (dismissed):
Somehow this went astray semantically, that needs fixing. But most of the comments are cosmetic.

src/webview/suggestions/views/lists.py
adekoder: That means wre are to show all suggestions  irrespective if the version.

fricklerhandwerk: No, but the filter needs to be aware of the status to be displayed.

src/shared/tests/conftest.py
adekoder: type error, and is way resolves it without having to type ignore

src/shared/models/linkage.py
adekoder: i'm following software versioning style here which typically start for 1 and not 0, but if you really prefer this i can change it, i don't have a preference.

src/shared/tests/conftest.py
fricklerhandwerk: But that's just because the static analysis doesn't get through the `@classproperty` decorator. Nothing's wrong with being smarter than the tools, except the choice of tools which is now fixed here.

src/webview/suggestions/views/lists.py
```diff
@@ -54,10 +54,17 @@ def get_queryset(self) -> QuerySet[CVEDerivationClusterProposal]:
             query_filters &= Q(status=self.status_filter)
         if self.package_filter is not None:
             query_filters &= Q(cached__payload__packages__has_key=self.package_filter)
+
+        qs = CVEDerivationClusterProposal.objects
+        # Only pending suggestions are filtered to the active algorithm version.
+        # Accepted and dismissed suggestions remain visible regardless of which
+        # algorithm generated them, so triagers don't lose access to work they
+        # have already reviewed.
+        if self.status_filter == CVEDerivationClusterProposal.Status.PENDING:
```
fricklerhandwerk: I expect this to break for the default `by-package` view: There's no status filter, and then it won't display any existing suggestions with outdated version. Needs a test.

src/shared/tests/test_regenerate_cached_suggestions.py
```diff
@@ -0,0 +1,182 @@
...
+    current_pending = make_suggestion(
+        status=CVEDerivationClusterProposal.Status.PENDING,
+    )
+    outdated_pending = make_suggestion(
+        container=make_container(cve_id="CVE-2025-1001"),
+        drvs={
+            make_drv(
+                pname="pkg-a", attribute="pkg-a"
+            ): ProvenanceFlags.PACKAGE_NAME_MATCH
+        },
```
fricklerhandwerk: Are these strictly required? (I don't think so)

```suggestion
```

Same in the last test

src/shared/tests/test_regenerate_cached_suggestions.py
```diff
@@ -0,0 +1,182 @@
...
+        status=CVEDerivationClusterProposal.Status.PENDING,
+        algorithm_version=_OUTDATED_VERSION,
+    )
+    assert CachedSuggestions.objects.count() == 3
+
+    call_command("regenerate_cached_suggestions", "--all", stdout=StringIO())
+
+    # we should have 2 cache suggestion, outdated pending would have been removed
+    assert CachedSuggestions.objects.count() == 2
```
fricklerhandwerk: Maybe it's easier to read if done as in the first test in the file, where you give names and say which of them are in the qs.

src/shared/models/linkage.py
fricklerhandwerk: For the cache schema the intuition was: we set 0 because there was no versioning beforehand (changes were ad hoc). No strong opinion, but I lean towards 0 because of that and for consistency.

src/shared/tests/test_proposal_queryset.py
```diff
@@ -0,0 +1,42 @@
...
+    """With proposals from two versions in the DB, .active() returns only the current one."""
+    current_proposal = make_suggestion()
+
+    other_container = make_container(cve_id="CVE-2025-9999")
+    other_drv = make_drv(pname="other-pkg", attribute="other-pkg")
+    outdated_proposal = make_suggestion(
+        container=other_container,
+        drvs={other_drv: ProvenanceFlags.PACKAGE_NAME_MATCH},
+        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION - 1,  # type: ignore
+    )
```
fricklerhandwerk: This works too

```suggestion
    outdated_proposal = make_suggestion(
        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION - 1,  # type: ignore
    )
```

The point of the fixtures is to avoid specifying information irrelevant to the test. Because that always takes space/effort to construct, as you see, and that distracts from trying to understand what's actually going on.

src/shared/management/commands/regenerate_cached_suggestions.py
```diff
@@ -23,23 +24,47 @@ def add_arguments(self, parser: ArgumentParser) -> None:
...
+
+        Pending proposals are only regenerated for the current algorithm version —
+        outdated-version pending proposals will be superseded by fresh ones once the
+        algorithm runs again, so regenerating their caches would be wasted effort.
+
+        Accepted and dismissed proposals are regenerated regardless of algorithm
+        version, since triagers have already reviewed them and they must remain
+        accessible in the UI.
+        """
+        return CVEDerivationClusterProposal.objects.filter(
```
fricklerhandwerk: Given we'll need essentially the same thing for the package view to work correctly (see other comment), maybe that should be a manager on the model?

fricklerhandwerk (commented):
Very good, excellent tests (again, I have only some tiny cosmetic comments).

I'm a bit surprised you removed the package view tests entirely, although maybe the previous version wasn't testing the exact condition that would currently still break. That needs coverage and the appropriate fix, and then this is good to merge.

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,121 @@
...
+    )
+
+    outdated_container = make_container(cve_id="CVE-2025-1002")
+
+    outdated_proposal = make_cached_suggestion(
+        container=outdated_container,
+        drvs={drv: ProvenanceFlags.PACKAGE_NAME_MATCH},
+        status=CVEDerivationClusterProposal.Status.PENDING,
+        algorithm_version=_OUTDATED_VERSION,
+    )
```
fricklerhandwerk, line 55:
```suggestion
    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
    drv: NixDerivation,
) -> None:
    """Only current-version proposals appear on the untriaged (pending) list."""
    current_proposal = make_cached_suggestion(
        status=CVEDerivationClusterProposal.Status.PENDING,
    )

    accepted_outdated = make_cached_suggestion(
        status=CVEDerivationClusterProposal.Status.ACCEPTED,
        algorithm_version=_OUTDATED_VERSION,
    )

    outdated_proposal = make_cached_suggestion(
        status=CVEDerivationClusterProposal.Status.PENDING,
        algorithm_version=_OUTDATED_VERSION,
    )
```

This should not change what the test is supposed to do.

I don't like using implicit knowledge that `make_cached_suggestion` will use the same `drv` underneath, but with appropriate care while reading it's okay to trade for a compact test case.

src/webview/tests/test_algorithm_version_filter.py
```diff
@@ -0,0 +1,121 @@
...
+
+    dismissed_container = make_container(cve_id="CVE-2025-2002")
+    dismissed_drv = make_drv(pname="dismissed-pkg", attribute="dismissed-pkg")
+    dismissed_outdated = make_cached_suggestion(
+        container=dismissed_container,
+        drvs={dismissed_drv: ProvenanceFlags.PACKAGE_NAME_MATCH},
+        status=CVEDerivationClusterProposal.Status.REJECTED,
+        algorithm_version=_OUTDATED_VERSION,
+    )
```
fricklerhandwerk, line 81:
```suggestion
    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
) -> None:
    """Accepted and dismissed lists show proposals from any algorithm version."""
    accepted_outdated = make_cached_suggestion(
        status=CVEDerivationClusterProposal.Status.ACCEPTED,
        algorithm_version=_OUTDATED_VERSION,
    )

    dismissed_outdated = make_cached_suggestion(
        status=CVEDerivationClusterProposal.Status.REJECTED,
        algorithm_version=_OUTDATED_VERSION,
    )
```

Same as above

fricklerhandwerk (approved):
Nice!

PR #1020, opened 2026-04-30
feat: API for changing suggestion state

fricklerhandwerk:


fricklerhandwerk:
Thanks @florentc for the review discussion yesterday! We should probably write down some of it publicly somewhere. No clue when there will be time for it, so just a brain dump in this random place:
- The general idea is to indeed get rid of the whole notion of "triage status", and infer it from whether user data was attached
- As discussed elsewhere, resolve the notion of suggestion into:
  - Match (mapping from cve:1 - n:derivation)
  - Issue (that has n matches, and a bunch of extra data, such as GitHub events)
    - An issue is considered published when there's an initial GitHub issue opened
- CVEs should become more first-class, with their own overlaid user-data attached and viewable separately
- We'll gradually expose all application logic interactions via API, and also gradually move the UI to use [TanStack](https://tanstack.com/) for more transparent state management in the front-end on top of that UI

florentc (approved):
Sorry for this review in message. In-code reviews seem down currently.

----

`v1_router = routers.DefaultRouter(trailing_slash=False)`

Does it matter?

---

`src/api/views.py`: Maybe we should already start to organize in submodules before it gets too crowded like our old global web view `views.py`


---

Custom exception handler: shouldn't it reside in `api` rather than shared? Alternatively call it `api_custom_exception_handler`.

fricklerhandwerk:
> execption handler

Oh yes, good catch!

> submodules

I thought so too, but arguably we should then also subdivide it into versions, etc. We can do this at ~0 cost any time in the future.

> `trailing_slash`

No, it's really just a matter of taste and saves 1 byte on every request. :exploding_head:

florentc:
> I thought so too, but arguably we should then also subdivide it into versions, etc. We can do this at ~0 cost any time in the future.

Since we know we will and we have a general plan in mind, IMO we should start now and ensure nice diffs for future PRs.

Issue #179, opened 2024-09-23
offers an RSS/Atom feed for status updates


fricklerhandwerk:

fricklerhandwerk:
Closed by #1023

PR #1023, opened 2026-05-06
feat: atom feeds for each package

florentc:
Introduces a feeds package (routes under `/feeds/`) and an atom feed per package at `/feeds/{package_name}` listing creation and status change events for suggestions which have the package as a non-ignored package.

Addresses #179 

----------


src/shared/logs/fetchers.py
```diff
@@ -157,3 +158,72 @@ def fetch_suggestion_events(
...
+        return []
+
+    suggestion_ids = list(
+        CVEDerivationClusterProposal.objects.filter(
+            derivations__attribute=package_name,
+        )
+        .exclude(
+            package_overlays__package_attribute=package_name,
+            package_overlays__overlay_type=PackageOverlay.Type.IGNORED,
+        )
```
fricklerhandwerk, line 186: This will violate the append-only nature of the feed, i.e. what people observed at one point in time won't be the same when queried later. (This is of course my assumtion that this is desirable.) To be precise we would need to scan *all* events and filter out those periods where the given package was ignored.

It does solve the problem elegantly though that if users query the feed after the package got ignored, they won't get spammed with irrelevant information. If we provide a `Cache-Control:` header, it would reduce the likelihood users ever get notifications that get discarded anyway.

fricklerhandwerk: Another case that seems not covered yet is if triagers click "dismissed (not in Nixpkgs)" we currently don't auto-ignore the related packages. This means users will get a notification on dismissal although their package is not even affected. The dismissal mechanism is in a sense unrelated, but here it interferes.

fricklerhandwerk: What we could also do is only update the result at fixed intervals to enforce this sparseness. Not sure if it's worth the hassle though.

fricklerhandwerk:
```suggestion
        # This will not display any events to users who query the feed after the package got ignored.
        # That is by design, to avoid noise.
        # This trades against the case where users who *did* query the feed *before* the package got ignored will never get notified that the event was irrelevant.
        # We deem it to be more beneficial like this than the other way round.
        .exclude(
            package_overlays__package_attribute=package_name,
            package_overlays__overlay_type=PackageOverlay.Type.IGNORED,
        )
```

src/shared/logs/fetchers.py
```diff
@@ -157,3 +158,72 @@ def fetch_suggestion_events(
...
+            "pgh_context"
+        ).filter(pgh_label="insert", pgh_obj_id__in=suggestion_ids)
+    )
+    for creation_event in creation_qs.iterator():
+        events.append(
+            RawCreationEvent(
+                suggestion_id=creation_event.pgh_obj_id,
+                timestamp=creation_event.pgh_created_at,
+                rejection_reason=CVEDerivationClusterProposal.RejectionReason(
+                    creation_event.rejection_reason
```
fricklerhandwerk, line 215: This will show automatic rejection events. I don't think we want this particular thing. We only want status change events that produce a rejection. So I guess those which have a rejection reason on insertion should be filtered out?

src/shared/logs/fetchers.py
```diff
@@ -157,3 +158,72 @@ def fetch_suggestion_events(
...
+                status_value=status_event.status,
+                rejection_reason=CVEDerivationClusterProposal.RejectionReason(
+                    status_event.rejection_reason
+                ).label.__str__()
+                if status_event.rejection_reason is not None
+                else None,
+            )
+        )
+
+    return sorted(events, key=lambda e: e.timestamp, reverse=True)
```
fricklerhandwerk:
```suggestion
    return events
```

@florentc said we don't need the sort since the feed entries will allow the client to sort by timestamp.

fricklerhandwerk: We probably want to limit the number of results though, so clients won't fetch *very many* events every time. We should pre-filter the timestamp in the queries, since we're querying multiple tables and can't meaningfully say anything about the actual result.

If Django supports [RFC 5005](https://www.rfc-editor.org/rfc/rfc5005), we could paginate the feed.

(This is definitely not a problem traffic-wise now, but putting a reasonable bound will make the probability of anyone eventually having to take care of unbounded growth go below 1.)

src/feeds/feeds.py
```diff
@@ -0,0 +1,118 @@
...
+        return f"Activity involving package '{package_name}'"
+
+    def link(self, package_name: str) -> str:
+        return reverse(
+            "webview:suggestion:suggestions_by_package",
+            kwargs={"package_name": package_name},
+        )
+
+    def description(self, package_name: str) -> str:
+        return f"Status changes for suggestions involving package '{package_name}' on Nixpkgs security tracker. Including creation of new suggestions, dismissal, acceptance, and publication. Events on suggestions after they ignore this package are no longer tracked."
```
fricklerhandwerk:
```suggestion
        return f"Status changes for suggestions involving package '{package_name}' on Nixpkgs security tracker. Including creation of new suggestions, dismissal, acceptance, and publication. After a package is ignored on a suggestion, related events are no longer tracked."
```

src/feeds/feeds.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def items(self, package_name: str) -> list[FeedItem]:
+        events = fetch_status_events_for_package(package_name)
+
+        suggestion_ids = list({e.suggestion_id for e in events})
+        suggestions = {
+            s.pk: s
+            for s in CVEDerivationClusterProposal.objects.filter(
+                pk__in=suggestion_ids,
+                cached__isnull=False,  # NOTE(@florentc): Should we filter them out or generate the caches with ensure_fresh_cache()?
```
fricklerhandwerk: This is fine. If we don't have the cached value for something in a given moment, the next pull will fit it in, and the client should deduplicate.

src/feeds/feeds.py
```diff
@@ -0,0 +1,118 @@
...
+class FeedItem:
+    suggestion: CVEDerivationClusterProposal
+    event: RawEventType
+
+
+def _describe_event(event: RawEventType) -> str:
+    """Return a human-readable label for a creation or status-change event."""
+    if isinstance(event, RawCreationEvent):
+        if event.rejection_reason:
+            return f"created (auto-dismissed: {event.rejection_reason})"
```
fricklerhandwerk, line 32: Since we shouldn't have those, let's not print it either


```suggestion
```

src/feeds/feeds.py
```diff
@@ -0,0 +1,118 @@
...
+    def item_title(self, item: Any) -> SafeText:
+        fi = cast(FeedItem, item)
+        cve_id = fi.suggestion.cve.cve_id
+        return mark_safe(f"{cve_id}: {_describe_event(fi.event)}")
+
+    def item_link(self, item: Any) -> str:
+        fi = cast(FeedItem, item)
+        return reverse(
+            "webview:suggestion:detail",
+            kwargs={"suggestion_id": fi.suggestion.pk},
```
fricklerhandwerk, line 114: You need to identify the event, since the client deduplicates by item ID ([and falls back to the link](https://docs.djangoproject.com/en/6.0/ref/contrib/syndication/#:~:text=item_guid%28self%2C%20obj%29%3A))

src/feeds/tests/test_package_feed.py
```diff
@@ -0,0 +1,173 @@
...
+    )
+    assert response.status_code == 404
+
+
+def test_feed_known_package_succeeds(
+    client: Client,
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    drv: NixDerivation,
+) -> None:
+    make_cached_suggestion()
```
fricklerhandwerk:
```suggestion
    cached_suggestion: CVEDerivationClusterProposal,
    drv: NixDerivation,
) -> None:
```

The other instances too

src/feeds/tests/test_package_feed.py
```diff
@@ -0,0 +1,173 @@
...
+    response = client.get(
+        reverse("feeds:package_feed", kwargs={"package_name": drv.attribute})
+    )
+    assert response.status_code == 200
+    content = response.content.decode()
+    assert "created" in content
+    assert "CVE-2025-0001" in content
```
fricklerhandwerk:
```suggestion
    assert cached_suggestion.cve.cve_id in content
```
not sure I remember the exact names correctly

src/feeds/tests/test_package_feed.py
```diff
@@ -0,0 +1,173 @@
...
+
+def test_feed_contains_status_change_event(
+    client: Client,
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    drv: NixDerivation,
+) -> None:
+    suggestion = make_cached_suggestion()
+    # Change status to accepted to produce a status-change event
+    suggestion.status = CVEDerivationClusterProposal.Status.ACCEPTED
+    suggestion.save()
```
fricklerhandwerk:
```suggestion
    cached_suggestion: CVEDerivationClusterProposal,
    drv: NixDerivation,
) -> None:
    cached_suggestion.change_status(CVEDerivationClusterProposal.Status.ACCEPTED)
```

yay!

src/feeds/tests/test_package_feed.py
```diff
@@ -0,0 +1,173 @@
...
+    client: Client,
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    drv: NixDerivation,
+) -> None:
+    make_cached_suggestion()
+    response = client.get(
+        reverse("feeds:package_feed", kwargs={"package_name": drv.attribute})
+    )
+    content = response.content.decode()
+    assert "nvd.nist.gov/vuln/detail/CVE-2025-0001" in content
```
fricklerhandwerk: Use the cve id from the cached_suggestion as well here

src/feeds/tests/test_package_feed.py
```diff
@@ -0,0 +1,173 @@
...
+    content = response.content.decode()
+
+    active_url = reverse(
+        "webview:suggestion:detail", kwargs={"suggestion_id": suggestion_active.pk}
+    )
+    ignored_url = reverse(
+        "webview:suggestion:detail", kwargs={"suggestion_id": suggestion_ignored.pk}
+    )
+    assert active_url in content
+    assert ignored_url not in content
```
fricklerhandwerk:
```suggestion
```

Seems redundant with the above, we're already testing both cases.

src/feeds/tests/test_package_feed.py
fricklerhandwerk:
```suggestion
    assert "application/atom+xml" in response["Content-Type"]
    content = response.content.decode()
    assert "created" in content
    assert cached_suggestion.cve.cve_id in content
```

fricklerhandwerk (commented):
It would be great to have [feed autodiscovery](https://www.rssboard.org/rss-autodiscovery) on the per-package view template. But since feeds are old-people things, it would also need a visual cue somewhere in the header:

Image

Most important issue is item IDs.

florentc:
Post review changes:

* Feed link & autodiscovery in package search pages 
* Clarify exclusion of ignored packages from feed events
* Filter out events older than 1 month in feeds (hardcoded for now)
* Fix weirdly phrased feed description
* Comment: elaborate on events of not-yet-cached suggestions in feeds
* Exclude auto-dismissals from package feeds
* Add feed item guid to distinguish between suggestion-based feed links
* Improve package feeds tests

src/shared/logs/fetchers.py
florentc: Added a hardcoded limit of 30days

src/feeds/feeds.py
florentc: Explained in comment

src/feeds/feeds.py
florentc: I removed auto dismiss events from the feeds but left the pretty printing as this is a standalone function expected to handle all the cases.

src/feeds/feeds.py
florentc: Done

src/feeds/tests/test_package_feed.py
florentc: Done

src/shared/logs/fetchers.py
florentc: Auto dismiss events have been removed from feeds (and we test that they are)

PR #1028, opened 2026-05-08
fix: Prevent a KeyError on certain cases

Erethon:
Reported on Matrix via @LeSuisse and I spotted it on the logs.


fricklerhandwerk (approved):
Good catch, thank you.

PR #1029, opened 2026-05-08
fix: disable setting custom email address for now

fricklerhandwerk:
At the moment the worst that could happen is that NixOS GitHub org members would be able to send notifications on their behalf and with their name to unsuspecting third parties.
That's still undesirable to allow in the first place.
Once any GitHub user can log in with the system, there won't be quick and easy way to stop abuse.
And in any case that would be easiest to avoid by verifying address ownership.


Erethon (approved):
Thank you for quickly handling this!

Issue #1024, opened 2026-05-06
REST API


florentc:
This issue is about ongoing work to implement a rest API for users directly (to use in their automated workflows) and also ultimately for the security tracker itself (to be called by the frontend: this would de-correlate business logic from presentation and improve maintainability and speed up implementing new features).

The following is an informal draft reflecting which endpoints we currently need for 1-1 feature parity between web views and API.

As endpoints are implemented, please ~~cross them out~~ from the following list.

Use this issue to link any relevant message or PR.

----------------------

- `api/v1/`

  - `suggestions/`

    - **GET:** list(Suggestion)
      - status filter (opt)
      - package filter (opt)
      - per page (opt)
      - page (opt)
       
    - `activity-log/`
      - **GET:** ActivityLog  

    - `by-id/{id}/` *or* `by-cve/{cve_id}`

      - **GET:** Suggestion

      - ~`status/`~

        - **GET:** SuggestionStatus

        - **POST AUTH (role: committer)**
          - new status
          - dismissal reason (opt)
          - comment (opt)

      - `packages/` (packages alongside their status in suggestion aka
        active/ignored)

        - **GET:** list(package with metadata including branches and
          versions)
          - per page (opt)
          - page (opt)
          - status filter (opt): active vs ignored

        - `{package_name}/`

          - **GET:** package with metadata including branches and
            versions

          - `ignore/`

            - **POST AUTH (role: committer)**

          - `restore/`

            - **POST AUTH (role: committer)**

      - `maintainers/` (maintainer detail (name email etc) alongside their
        status in suggestion aka active/ignored/additional)

        - **GET:** list(NixMaintainer)
          - per page (opt)
          - page (opt)
          - status filter (opt): active/ignored/additional

        - `ignore/{maintainer_id}/`

          - **POST AUTH (role: committer)**

        - `restore/{maintainer_id}/`

          - **POST AUTH (role: committer)**

        - `delete/{maintainer_id}/`

          - **POST AUTH (role: committer)**

        - `add`

          - **POST AUTH (role: committer)**
            - new maintainer\'s github handle

      - `references/` (deduplicated references with their deduplicated
        name and tag list, alongside their status aka active/ignored)

        - **GET:** list(DeduplicatedReference)
          - per page (opt)
          - page (opt)
          - status filter (opt): active/ignored

        - `ignore/`

          - **POST AUTH (role: committer)**
            - reference url

        - `restore/`

          - **POST AUTH (role: committer)**
[...]

PR #1031, opened 2026-05-09
feat: GET /api/v1/suggestions/{id}/status

DarshanCode2005:
Adds a public read-only endpoint that returns the same triage fields as change_status (status, optional rejection_reason, optional comment). POST /change_status is unchanged (still authenticated committers/staff). Includes API tests for success, 404, rejected payload, and comment.

Related to #1024


fricklerhandwerk (approved):
Nice, thanks!

fricklerhandwerk:
Rebased and removed the unnecessary `pytestmark =` line.

PR #1032, opened 2026-05-10
feat: get issue by code in API

DarshanCode2005:
Issue detail is now GET `/api/v1/issues/<code>/` where `<code>` matches the web (NIXPKGS-YYYY-NNNN, same regex as the UI). List behaviour and filters are unchanged.

**Breaking Change** : detail no longer accepts the numeric primary key in the URL. Update callers to use `code`.

Related to #1024


fricklerhandwerk:
We shouldn't break any API just like that, even if there's ~0 users at the moment. I put another endpoint next to it. For v2 we can factor it more nicely.

Thanks a lot though, good observation!

PR #1035, opened 2026-05-15
feat: cluster derivations by package

adekoder:
This PR is the Phase 1 of Nix Derivation data De-duplication.

The `homepage` and `description` are duplicate across evaluation of the same packages, we are extracting them into 
`NixDerivationHomepage` and `NixDerivationDescription` tables as a first step toward package-level deduplication.

Notes:
 ```After Deployment we need to run the following commands
 1. Run manage.py migrate_homepage_description --dry-run — verify counts
 2. Run manage.py migrate_homepage_description
 3. Run manage.py migrate_homepage_description --null-after to clear the old columns
 ```


src/shared/tests/test_migrate_homepage_description.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_migrates_homepage_and_description_to_new_table(
+    legacy_meta: NixDerivationMeta,
+) -> None:
+
+    call_command("migrate_homepage_description", stdout=StringIO())
+
+    assert NixDerivationHomepage.objects.filter(
+        metadata=legacy_meta, url="https://example.com"
```
fricklerhandwerk: I think we should not assume intimate knowledge of fixture internals. In the other PR recently we made a small exception because it would have been really cumbersome to avoid, but here that's not the case.

```suggestion
        metadata=legacy_meta, url=legacy_meta.homepage
```

...or whatever would be correct here.

src/shared/tests/test_migrate_homepage_description.py
```diff
@@ -0,0 +1,59 @@
...
+    ).exists()
+    assert NixDerivationDescription.objects.filter(
+        metadata=legacy_meta, text="My package"
+    ).exists()
+
+    legacy_meta.refresh_from_db()
+    assert legacy_meta.homepage == "https://example.com"
+    assert legacy_meta.description == "My package"
```
fricklerhandwerk: And then I guess just make sure it stays what it was.

```suggestion
    assert legacy_meta.homepage 
    assert legacy_meta.description
```

Or maybe better, remember the values before running the command

src/shared/tests/test_migrate_homepage_description.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def test_null_after_clears_homepage_and_description_column(
```
fricklerhandwerk: It seems like this is essentially the same test as above, except the flag would make the final assertion compare to None instead of the original value. Not insisting it should be done via parameter, but the test should at least have the exact same shape.

src/shared/tests/test_migrate_homepage_description.py
```diff
@@ -0,0 +1,59 @@
...
+
+    legacy_meta.refresh_from_db()
+    assert legacy_meta.homepage is None
+    assert legacy_meta.description is None
+
+    assert NixDerivationHomepage.objects.filter(metadata=legacy_meta).exists()
+    assert NixDerivationDescription.objects.filter(metadata=legacy_meta).exists()
+
+
+def test_command_is_idempotent_homepage(legacy_meta: NixDerivationMeta) -> None:
```
fricklerhandwerk:
```suggestion
def test_command_is_idempotent(legacy_meta: NixDerivationMeta) -> None:
```

It seems that generally the naming is lagging a bit behind the fact that we're also migrating the description. Could you please check and clean that up globally? (I don't think it should be `migrate_homepage_and_description` everywhere but something brief and locally meaningful such as `split_meta`)

src/shared/tests/test_migrate_homepage_description.py
```diff
@@ -0,0 +1,59 @@
...
+
+    call_command("migrate_homepage_description", stdout=StringIO())
+    call_command("migrate_homepage_description", stdout=StringIO())
+
+    assert NixDerivationHomepage.objects.count() == 1
+    assert NixDerivationDescription.objects.count() == 1
+
+
+def test_already_migrated_rows_are_skipped(drv: NixDerivation) -> None:
```
fricklerhandwerk: Why are we testing this? The test does not seem to match the description

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+    DerivationClusterProposalLink.objects.create(
+        proposal=suggestion,
+        derivation=drv,
+        provenance_flags=ProvenanceFlags.PACKAGE_NAME_MATCH,
+    )
+    cache_new_suggestions(suggestion)
+
+    cached = CachedSuggestions.objects.get(proposal=suggestion)
+    pkg = cached.payload["packages"][drv.attribute]
+    assert pkg["description"] == "Dummy derivation"
```
fricklerhandwerk: Wait, this is not actually testing what is advertised. Don't you want more like taking an existing regular cached suggestion, run the migration, update the homepage or description in the split table, recache, and check that the cached value was updated correctly?

src/shared/tests/conftest.py
```diff
@@ -241,14 +243,19 @@ def wrapped(
...
+            description=description,
+            homepage=homepage,
             insecure=False,
             available=True,
             broken=False,
             unfree=False,
             unsupported=False,
         )
+        NixDerivationDescription.objects.create(metadata=meta, text=description)
+        NixDerivationHomepage.objects.create(metadata=meta, url=homepage)
```
fricklerhandwerk: Should we do this though? This way the fixtures present
1. a post-migration state, which is not a precondition for any test
2. not something ingestion is intended to produce; we're currently planning to populate these rows after the fact

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+
+def test_cache_description_falls_back_when_no_new_table_entry(
+    cve: Container,
+    drv: NixDerivation,
+) -> None:
+    """
+    When no NixDerivationDescription row exists (unmigrated derivation),
+    the cached payload must fall back to NixDerivationMeta.description.
+    """
+    NixDerivationDescription.objects.filter(metadata=drv.metadata).delete()
```
fricklerhandwerk: Not needed if not created automatically (which I strongly suggest!)

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+    # Confirm the legacy column still holds the value
+    assert drv.metadata is not None
+    assert drv.metadata.description == "Dummy derivation"
```
fricklerhandwerk: Hm this just tests the fixture

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+    DerivationClusterProposalLink.objects.create(
+        proposal=suggestion,
+        derivation=drv,
+        provenance_flags=ProvenanceFlags.PACKAGE_NAME_MATCH,
+    )
+    cache_new_suggestions(suggestion)
+
+    cached = CachedSuggestions.objects.get(proposal=suggestion)
+    pkg = cached.payload["packages"][drv.attribute]
+    assert pkg["description"] == "Dummy derivation"
```
fricklerhandwerk: I don't understand, this just tests what already happens

fricklerhandwerk (dismissed):
Only reviewed the tests, let's fix those first

src/shared/cache_suggestions.py
```diff
@@ -296,6 +304,23 @@ def get_src_position(derivation: NixDerivation) -> str | None:
     return None
 
 
+def _get_description(metadata: NixDerivationMeta) -> str | None:
```
fricklerhandwerk: This should be a method (or maybe better `@property`?) on `NixDerivationMeta`.

src/shared/cache_suggestions.py
```diff
@@ -296,6 +304,23 @@ def get_src_position(derivation: NixDerivation) -> str | None:
...
+    has been run.
+    """
+
+    description: str | None = None
+    try:
+        description = metadata.description_obj.text
+    except NixDerivationDescription.DoesNotExist:
+        description = metadata.description
+
+    return description
```
fricklerhandwerk:
```suggestion
    try:
        return metadata.description_obj.text
    except NixDerivationDescription.DoesNotExist:
        return metadata.description
```

Is this variant what CodeQL originally complained about?

src/shared/cache_suggestions.py
```diff
@@ -309,10 +334,9 @@ def channel_structure(
         if attribute_path not in packages:
             packages[attribute_path] = CachedSuggestion.Package()
             if derivation.metadata:
-                if derivation.metadata.description:
-                    packages[
-                        attribute_path
-                    ].description = derivation.metadata.description
+                description = _get_description(derivation.metadata)
+                if description:
+                    packages[attribute_path].description = description
```
fricklerhandwerk:
```suggestion
                packages[attribute_path].description = _get_description(derivation.metadata)
```

The types match anyway

src/shared/cache_suggestions.py
adekoder: yes

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+    drv: NixDerivation,
+) -> None:
+    """
+    When no NixDerivationDescription row exists (unmigrated derivation),
+    the cached payload must fall back to NixDerivationMeta.description.
+    """
+    NixDerivationDescription.objects.filter(metadata=drv.metadata).delete()
+    # Confirm the legacy column still holds the value
+    assert drv.metadata is not None
+    assert drv.metadata.description == "My package"
```
fricklerhandwerk:
```suggestion
    description = drv.metadata.description
```
And then compare at the end with that string

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+
+    suggestion = CVEDerivationClusterProposal.objects.create(
+        status="pending",
+        cve=cve.cve,
+    )
+    DerivationClusterProposalLink.objects.create(
+        proposal=suggestion,
+        derivation=drv,
+        provenance_flags=ProvenanceFlags.PACKAGE_NAME_MATCH,
+    )
```
fricklerhandwerk: I don't think we should reconstruct what the fixture is doing. Either obtain the drv for manipulating the NixDerivationDescription from a suggestion fixure, or (less optimal, but we do it in multiple places and it's okay-ish) use `drv` and `suggestion` fixtures with implicit knowledge that they belong together.

src/shared/tests/test_suggestion_caching.py
```diff
@@ -83,3 +84,54 @@ def test_cache_stale_check(
...
+    cache_new_suggestions(suggestion)
```
fricklerhandwerk: Take `suggestion` fixture instead, and manipulate the NixDerivationDescription in preparation, so we know that it's not the original description in the cached value.

fricklerhandwerk:
Rebased and reordered the commits to have the test refactoring be a preparation rather than an afterthought.

fricklerhandwerk:
Rebased on top of https://github.com/NixOS/nix-security-tracker/pull/537, since we decided to just go ahead and do the package clustering in this PR already. Reason: We'd be moving >100M rows, which we'll have to go through *again* if we clustered in a follow-up. We're already wasting disk space, and may as well deduplicate immediately.

fricklerhandwerk:
I've implemented package clustering as discussed with @adekoder:
- on trigger after complete eval
- as a command
- tests
- removed the previous merely-moving of data

There were few interesting edge cases to consider that weren't obvious so far, such as how to organize the catch-up so it doesn't clobber a fresh eval if it's still running.

Closes https://github.com/NixOS/nix-security-tracker/issues/790 now.

Marked as draft, since there are still somewhat unrelated preliminary changes baked in. I split them out into separate PRs, after which I'll rebase this one:
- https://github.com/NixOS/nix-security-tracker/pull/1049
- https://github.com/NixOS/nix-security-tracker/pull/1050
- https://github.com/NixOS/nix-security-tracker/pull/1051
- https://github.com/NixOS/nix-security-tracker/pull/1052

src/shared/package_clustering.py
```diff
@@ -0,0 +1,89 @@
...
+            new_links.append(PackageDerivation(derivation=drv, package=pkg))
+
+    # `seen` deduplicates across groups that share the same `(pname, homepage)` but have different derivation paths.
+    seen: dict[tuple[str, str | None], Package] = {}
+    for group in remainder:
+        # All derivations in a group share the same `derivation_path` and therefore the same name and homepage.
+        # Any element is representative.
+        first = group[0]
+        pname, _ = parse_drv_name(first.name)
+        homepage = first.metadata.homepage if first.metadata else None
```
adekoder: if we are getting the  `homepage` and `description` from metadata directly, then there is no point of the initial work i did where we have a new table for this data and storing data to the new table during ingestion

src/shared/package_clustering.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    Assign derivations to packages.
+
+    For each derivation-path group:
+    1. If any attrpath in the group is already registered, use the associated package.
+    2. Otherwise find or create a package by `(pname, homepage)`, with intra-batch deduplication so two derivations with the same key share one package.
+    3. Register any new attrpaths for the resolved package.
+    """
+    unmatched = (
+        derivations.filter(package_link__isnull=True)
```
adekoder: since this is going to be fetching a lot of data, would suggest we use `.values()` to specify set of fields we need and load just those in memory

src/shared/package_clustering.py
```diff
@@ -0,0 +1,89 @@
...
+        for drv in group:
+            if drv.attribute not in attrpath_to_pkg:
+                new_attrpaths.append(
+                    PackageAttrpath(attrpath=drv.attribute, package=pkg)
+                )
+            new_links.append(PackageDerivation(derivation=drv, package=pkg))
+
+    # `seen` deduplicates across groups that share the same `(pname, homepage)` but have different derivation paths.
+    seen: dict[tuple[str, str | None], Package] = {}
+    for group in remainder:
```
adekoder: i will suggest we wrap this part within transaction block, so  that we can rollback the Package creation if anything crashes 

`with transaction.atomic():`

src/shared/listeners/package_clustering.py
```diff
@@ -0,0 +1,28 @@
...
+def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
+    if old.state == new.state:
+        return
+    if new.state != NixEvaluation.EvaluationState.COMPLETED:
+        return
+    evaluation = NixEvaluation.objects.select_related("channel").get(pk=new.pk)
+    logger.info("Clustering derivations from evaluation %s", evaluation)
+    result = cluster_packages(
+        NixDerivation.objects.filter(parent_evaluation_id=new.pk),
+        update_packages=evaluation.channel.is_rolling,
```
adekoder: i want to understand the context why we have the `update_packages=evaluation.channel.is_rolling`

adekoder: never mind i found the in the usage, would be nice if there is comment here too...  also why `is_rolling` ?

fricklerhandwerk: Could have been a richer name, e.g. `is_rolling_release`. See the other PR; it says whether an evaluation's source channel represents a rolling release. But yes, here we also need to say why we care: we only want to use the latest information about the package (as a higher level object than the point-in-time derivation).

src/shared/management/commands/backfill_package_clustering.py
```diff
@@ -0,0 +1,46 @@
...
+        rolling = unmatched.filter(parent_evaluation__channel__in=rolling_channels)
+        non_rolling = unmatched.exclude(parent_evaluation__channel__in=rolling_channels)
+        updated: set[int] = set()
+        created: set[int] = set()
+
+        # Rolling derivations first so they seed `homepage`/`description` on new packages.
+        for label, qs in [("rolling", rolling), ("non-rolling", non_rolling)]:
+            self.stdout.write(f"Pass: {label}...")
+            # Never overwrite package fields here to avoid races with the listener.
+            result = cluster_packages(qs, update_packages=False, batch_size=batch_size)
```
fricklerhandwerk: Name seems not to be obvious?
```suggestion
            result = cluster_packages(qs, update_package_metadata=False, batch_size=batch_size)
```

fricklerhandwerk:
Reviewed with @adekoder, who correctly pointed out the remaining race condition between the post-eval listener and the backfill.

- Renamed the command to highlight that it's for backfilling

  It now starts with the services to recover what crashed listeners may have left over
- Cut off the backfill query up to the latest complete evaluation.

  This ensures that listener and backfill command never touch the same rows
  
- Wrapped each batch into a transaction to avoid partial updates (especially deleting the metadata from the source derivation!)

  @adekoder Probably we don't need row locking after all (it's still there after the messy fix), since the concurrent queries are now non-overlapping

Left to do:
- Review tests
- Test the concurrency scenario
- Resolve merge conflicts
- Massage history

src/shared/management/commands/backfill_package_clustering.py
```diff
@@ -0,0 +1,68 @@
...
+        # is safe here: the cutoff ensures we only touch evaluations the listener has
+        # already finished with (or crashed on), so there is no concurrent listener
+        # that could race on the same packages.
+        # Non-rolling channels must never overwrite metadata set by rolling.
+        for label, qs, update in [
+            ("rolling", rolling, True),
+            ("non-rolling", non_rolling, False),
+        ]:
+            self.stdout.write(f"Pass: {label}...")
+            result = cluster_packages(qs, update_packages=update, batch_size=batch_size)
```
fricklerhandwerk: I think this is actually wrong. What guarantees that the backfill has already passed a given package when the post-eval listener hits it?  

```suggestion
        # Derivations from rolling channels first so they seed `homepage`/`description` on new packages.
        for label, qs in [
            ("rolling", rolling),
            ("non-rolling", non_rolling),
        ]:
            self.stdout.write(f"Pass: {label}...")
            # Never overwrite package fields here to avoid races with the listener.
            result = cluster_packages(qs, update_packages=False, batch_size=batch_size)
```

src/shared/package_clustering.py
```diff
@@ -0,0 +1,163 @@
...
+                "pk",
+                "attribute",
+                "derivation_path",
+                "name",
+                "metadata_id",
+                "metadata__homepage",
+                "metadata__description",
+            )
+            .order_by("derivation_path")
+        )
```
fricklerhandwerk: Setting [`named=True`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#django.db.models.query.QuerySet.values_list) would give us attribute access via namedtuples, and reduce visual noise quite a bit. The code is dense enough as it is.

```suggestion
        rows = list(
            NixDerivation.objects.filter(pk__in=batch_pks, package_link__isnull=True)
            .select_for_update(skip_locked=True)
            .exclude(metadata=None)
            .values_list(
                "pk",
                "attribute",
                "derivation_path",
                "name",
                "metadata_id",
                "metadata__homepage",
                "metadata__description",
                named=True,
            )
            .order_by("derivation_path")
        )
```

Alternatively set up a `@dataclass` that would declare what we're dealing with. Not sure the second option is worth the extra LOC though.

adekoder: yes, i agree this make it more readable, but there is a performance issue converting to nametuple

src/shared/management/commands/backfill_package_clustering.py
fricklerhandwerk: The flag only prevents updating metadata (homepage, description) on existing packages. This will always be caught up by the next successful post-eval trigger. In any case, even if the post-eval trigger crashes, the backfill command will always link newly ingested derivations to existing packages.

src/shared/package_clustering.py
fricklerhandwerk: I think the performance is negligible; by far most of the work is probably in the query at this point.

src/shared/cache_suggestions.py
```diff
@@ -208,7 +208,10 @@ def cache_new_suggestions(suggestion: CVEDerivationClusterProposal) -> None:
 
     derivations = list(
         suggestion.derivations.select_related(
-            "metadata", "parent_evaluation", "parent_evaluation__channel"
+            "metadata",
+            "package_link_package",
```
fricklerhandwerk:
```suggestion
            "package_link__package",
```
I think that's why the test fails

src/shared/tests/conftest.py
```diff
@@ -231,10 +232,36 @@ def maintainer(make_maintainer: Callable[..., NixMaintainer]) -> NixMaintainer:
     return make_maintainer()
 
 
+@pytest.fixture
+def make_metadata(maintainer: NixMaintainer) -> Callable[..., NixDerivationMeta]:
```
fricklerhandwerk: Seems like an artefact from history rewriting

src/shared/package_clustering.py
```diff
@@ -0,0 +1,170 @@
...
+                    new_attrpaths.append(
+                        PackageAttrpath(attrpath=row["attribute"], package=pkg)
+                    )
+                new_links.append(
+                    PackageDerivation(derivation_id=row["pk"], package=pkg)
+                )
+                done.append(row["metadata_id"])
+
+        PackageAttrpath.objects.bulk_create(new_attrpaths, ignore_conflicts=True)
+        PackageDerivation.objects.bulk_create(new_links, ignore_conflicts=True)
```
fricklerhandwerk, line 179: `skip_locked=True` only prevents concurrent access to the same derivation, but not `Package`, `PackageAttrpath`, or the overall clustering decision.

If backfill and post-eval trigger race at the batch boundary, we can end up with
- attrpath `foo.bar` registered to package P1
- some derivations with `foo.bar` linked to P1
- other derivations with `foo.bar` linked to P2

This will produce cleanup work and, likely, extreme confusion. I prefer fixing now this since we don't have UI for cleanups, and it will take a while until we get to that UI. Working on a draft.

fricklerhandwerk:
What an exciting set of new problems we found there again... okay, I'm now testing this locally to see the effect, otherwise looks ready to merge once you look at the fixups @adekoder and we squash everything again.

fricklerhandwerk:
Tested it and found even more fun little things. For instance, with a too high batch size we'd be running in to limits on how many locks we can hold. This can be tuned in Postgres settings, but for simplicity I just reduced the batch size a bit. What seems to work IRL is crashing and picking it up again. This is important, because the migration will likely run for many many hours, and we may get deployments in between. I didn't do the full thing, but it quickly approximates 85k packages with 100k attribute paths.

fricklerhandwerk:
Thank you for the endurance @adekoder. Although much churn came from me, this is largely your work;  design, investigation, bugfixing. I just did the note-taking and polishing.

PR #1043, opened 2026-05-20
fix: automatically create nixpkgs directory 

### LINKED ISSUE #1026: manage help command not working
hello , newbie here. I was reading [contribution guideline](https://github.com/NixOS/nix-security-tracker/blob/main/CONTRIBUTING.md) to start working on an issue that I recently was assigned #627 . the guideline mentions to use `manage help` for available commands however i get below issue 

Image

devnchill:
resolves #1026 

- renames `dummy-credentials` to `shell-config-placeholder` (both nix shell and docs) 
- this command would now also create `nixpkgs` dir.
- move `USER_SETTINGS_FILE` to `env` from `shellhook export` 

Some thoughts i'd like to share ... 
- Atm we create `settings.py` file inside shellhook, so I was thinking it would be better to move that too inside the `shell-config-placeholder` 
- maybe rename `shell-config-placeholder` to something like `quickstart` or `bootstrap` or `setup-local-env` ... something like it ?

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -62,13 +62,15 @@ rec {
       '';
       # Run this for a quick start.
       # Login and publishing issues requires setting up credentials properly.
-      dummy-credentials = pkgs.writeShellApplication {
-        name = "dummy-credentials";
+      shell-config-placeholder = pkgs.writeShellApplication {
+        name = "shell-config-placeholder";
         runtimeInputs = [ pkgs.python3 ];
         text = ''
-          dir="${toString ./.credentials}"
-          mkdir -p "$dir"
-          cd "$dir"
+          credentialsDir="${toString ./.credentials}"
+          nixpkgsDir="${toString ./nixpkgs}"
+          mkdir -p "$credentialsDir"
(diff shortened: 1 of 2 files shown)


default.nix
```diff
@@ -86,6 +88,7 @@ rec {
         PGDATABASE = "nix-security-tracker";
         PGUSER = "nix-security-tracker";
         CREDENTIALS_DIRECTORY = toString ./.credentials;
+        USER_SETTINGS_FILE = toString ./.settings.py;
```
fricklerhandwerk: This is quite unrelated. Arugably we should just remove this altogether? But I haven't thought this through, I don't remember when this file even played a role last time. Please let's figure that out in a different PR if you want that cleaned up.

devnchill: alright

fricklerhandwerk:
Please rebase and squash to one commit. I can do it too when back at the keyboard, but if you do it I can click it in any time.

devnchill:
I'm outside now, will do it once I reach home 😅

PR #1049, opened 2026-05-22
test: send notification explicitly

fricklerhandwerk:
This allows more clearly expressing that we're modeling the fixtures after how ingestion works.

Note how `make_drv` now gets "parsed data" including the metadata.
It adds the separate homepage and description rows, and then completes the derivation.

### DIFF
--- a/src/shared/tests/test_notifications.py
+++ b/src/shared/tests/test_notifications.py
@@ -2,18 +2,19 @@ from collections.abc import Callable
 
 from django.contrib.auth.models import User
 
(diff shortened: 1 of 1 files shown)


adekoder (commented):
i thought this changes already part of my PR, why are we creating this one ?

fricklerhandwerk:
Because this is not strictly related to the PR and we already agreed this is not wrong. Having this merged will make the other diff smaller.

fricklerhandwerk:
Turns out we don't need the refactoring for #1035 after all. Reduced it to the clarification on the notification test that the refactoring revealed.

PR #1051, opened 2026-05-22
refactor: move rolling-release predicate to NixChannel

fricklerhandwerk:


src/shared/models/nix_evaluation.py
```diff
@@ -133,6 +133,12 @@ class Meta:  # type: ignore[override]
         ]
 
 
+class NixChannelQuerySet(models.QuerySet):
+    def rolling(self) -> "NixChannelQuerySet":
+        # FIXME(@fricklerhandwerk): This will fall apart when we obtain the channel structure dynamically [ref:channel-structure]
```
adekoder, line 138: if this will cause some issues why are we doing it ?

src/shared/models/nix_evaluation.py
```diff
@@ -174,6 +182,11 @@ class ChannelState(models.TextChoices):
     def __str__(self) -> str:
         return f"{self.staging_branch} -> {self.channel_branch} (Release: {self.release_version})"
 
+    @property
+    def is_rolling(self) -> bool:
+        # FIXME(@fricklerhandwerk): This will fall apart when we obtain the channel structure dynamically [ref:channel-structure]
```
adekoder, line 193: same as above

src/shared/models/nix_evaluation.py
fricklerhandwerk: The comment is more of a heads-up for when a change happens, and then this is expected to break. But the actual change will likely be involved, and what exactly it's supposed to be like is also not fully elaborated yet. So it doesn't make sense to "fix" it or do it "more right", since it's not defined what that means, and figuring out is out of scope for the moment. A textbook FIXME :)

PR #1053, opened 2026-05-28
docs: simplify and update architecture diagram to match current deployment

### LINKED ISSUE #931: Update architecture diagram
We changed a few things since it was merged, and it's also syntactically broken (bold and italics need to be syntactically correct).

- Evaluation doesn't currently touch the Nix store
- Evaluation ingests into PG
- CVE ingestion triggers matching, not some generic worker
- Would be super cool to have links to the code that does the work named in a given box (make sure lychee checks those too!)

https://github.com/NixOS/nix-security-tracker/blob/main/docs/architecture.mermaid

DarshanCode2005:
- Reworked `docs/architecture.mermaid` to reflect the current production architecture (Daphne/ASGI, scheduled jobs, pgpubsub workers, storage, and external dependencies).
- Simplified the diagram layout to reduce arrow clutter and make the core data flows easier to follow.
- Updated `docs/README.md` to reference the current channel monitoring endpoint (`monitoring.nixos.org`) so it matches runtime configuration.

## Why
The previous diagram had outdated labels and too much visual noise, which made onboarding and operational understanding harder. This update keeps the architecture documentation accurate and easier to read.

Closes #931

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -15,7 +15,7 @@ The tracker needs to communicate with third party services, namely:
 - GitHub repositories:
   - https://github.com/nixos/nixpkgs to pull the latest changes from Nixpkgs
   - https://github.com/CVEProject/cvelistV5 to pull CVE data
-- https://prometheus.nixos.org/ to get information about the latest channels
+- https://monitoring.nixos.org/prometheus/api/v1/query?query=channel_revision to get information about the latest channels
 
 ## Storage space considerations
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -47,6 +47,25 @@ Ideally we will eventually get rid of the cache, but it requires incremental rew
 
 **Activity log**: Issue status changes and metadata edits are tracked automatically via [`django-pghistory`](https://github.com/AmbitionEng/django-pghistory).
 
+## Diagram to implementation map
+
+Maps each diagram box to its source file.
+
+| Diagram box | Source |
+|-------------|--------|
+| Daphne ASGI Django | [src/project/asgi.py](../src/project/asgi.py) |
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
The visual noise was due to malformed syntax though. In fact it was more readable in the graphical rendering due to the box-grouping and colors. I recommend taking a closer look at the issue description and making more granular changes. Wholesale rewrites generally don't fly well when reviewing, regardless of what it is.

DarshanCode2005:
architecture

docs/README.md
```diff
@@ -47,6 +47,25 @@ Ideally we will eventually get rid of the cache, but it requires incremental rew
 
 **Activity log**: Issue status changes and metadata edits are tracked automatically via [`django-pghistory`](https://github.com/AmbitionEng/django-pghistory).
 
+## Diagram to implementation map
+
+Maps each diagram box to its source file.
```
fricklerhandwerk, line 52: Mermaid syntax supports links. Not sure it will work from GH rendering, but maybe it does?

DarshanCode2005: Although in the IDE's preview it was not opening. That's why I have added it.

fricklerhandwerk: Then we'd need to discuss (probably out of band) whether viewing the diagram from IDE is a use case worth supporting. Because I don't think there's a straightforward *and* reliable way to do it that doesn't involve shipping our own tools for viewing

fricklerhandwerk:
Making it more orderly is generally a very good idea! The trick is now to figure out step by step which information is important to show and whether it will produce the right mental model.

PR #1054, opened 2026-05-28
feat: toggle subscribing to maintained packages via API

fricklerhandwerk:
Another step towards #1024

### DIFF
--- a/src/api/subscriptions/views.py
+++ b/src/api/subscriptions/views.py
@@ -1,0 +1,41 @@
+from rest_framework import serializers, status, viewsets
+from rest_framework.decorators import action
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from webview.models import Profile
+
+
+class SubscriptionsViewSet(viewsets.GenericViewSet):
+    queryset = Profile.objects.none()  # Add collection resource here later
+    permission_classes = [IsAuthenticated]
+
+    class AutoSubscribeSerializer(serializers.Serializer):
+        enabled = serializers.BooleanField()
+
+    @action(
+        detail=False,
+        methods=["get", "put"],
+        url_path="auto-subscribe-to-maintained-packages",
+        serializer_class=AutoSubscribeSerializer,
+    )
+    def auto_subscribe_to_maintained_packages(self, request: Request) -> Response:
+        profile = request.user.profile
+        if request.method == "GET":
+            serializer = self.get_serializer(
+                {"enabled": profile.auto_subscribe_to_maintained_packages}
+            )
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        elif request.method == "PUT":
+            serializer = self.get_serializer(data=request.data)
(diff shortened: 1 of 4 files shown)


src/api/subscriptions/views.py
```diff
@@ -0,0 +1,37 @@
...
+        detail=False,
+        methods=["get", "put", "delete"],
+        url_path="watch-maintained-packages",
+    )
+    def watch_maintained_packages(self, request: Request) -> Response:
+        profile = request.user.profile
+        if request.method == "GET":
+            enabled = profile.auto_subscribe_to_maintained_packages
+            return Response(
+                status=status.HTTP_200_OK if enabled else status.HTTP_404_NOT_FOUND
```
florentc: I find this very weird. We are querying a boolean option. IMO it should always return 200 and inside the response body either true or false.

404 would be for non existing resource. For instance if we have an endpoint to query whether a package is subscribed to or not, I'd expect 200 + true, or 200 + false, or 404 if the package doesn't exist. But I'd not expect 404 to encode "false".

40x errors are for bad requests and there is nothing wrong in the request if the auto_subscribe_to_maintained_package is set to false

src/api/subscriptions/views.py
```diff
@@ -0,0 +1,37 @@
...
+            )
+        elif request.method == "PUT":
+            profile.auto_subscribe_to_maintained_packages = True
+            profile.save()
+            return Response(status=status.HTTP_200_OK)
+        elif request.method == "DELETE":
+            profile.auto_subscribe_to_maintained_packages = False
+            profile.save()
+            return Response(status=status.HTTP_204_NO_CONTENT)
```
florentc, line 37: Why PUT and DELETE?

I'd expect PUT and DELETE for an API route about a collection of stuff (e.g. subscribed packages) in which you add or delete something. But not for setting a fixed parameter (that I'd see as a resource that's always present but can have 2 values) to true or false. Why not PATCH or POST (or even PUT for both true and false) with a boolean parameter in the request body?

------------------

Why not 204 for PUT as well?

florentc (changes requested):
I disagree about the choice of http codes and verbs.

It seems to me your representation is the following: a flag (here the auto-subscribe parameter) is a resource that can be here or not, hence the PUT, DELETE, and 404

I find this quite counterintuitive.

I don't have much against PUT and DELETE but I'm strongly opposed to encoding "false" with a 404 (there is nothing wrong with the user request)

EDIT:

I forgot to add a nitpick about the name `watch_maintained_packages`

This sounds like it's about the collection of subscribed packages. In fact it's about the auto subscribe flag. I think we should match the names to avoid confusion (`watch_auto_subscribe_to_maintained_packages`)

fricklerhandwerk:
Agreed on the design, I was trying to be too clever there.

I'd prefer a shorter name for the method/path, and I do think "watch packages you maintain" says the right thing. Not convinced we need to emphasize that it's automatic. Won't die on that hill though. Changed everything as requested, and while at it also automated the input validation.

florentc (approved):

PR #1057, opened 2026-05-29
docs: correct evaluation data flow in architecture diagram

DarshanCode2005:
Topologically sorted 2nd change:

**Removed**: 
FetchAllChannels -->|4 Evaluate Nix| NixEval --> NixStore

**Added**:

FetchAllChannels -->|4 update channels| PostgreSQL # **channel updates from the command**
PostgreSQL -.->|NixEvaluationChannel NOTIFY| NixEval # **async trigger via pgpubsub**
NixEval -->|ingest derivations| PostgreSQL # **evaluation writes derivations to the DB**

Related to #931

### DIFF
--- a/docs/architecture.mermaid
+++ b/docs/architecture.mermaid
@@ -44,7 +44,7 @@ graph TB
     FetchAllChannels -->|2 Git pull| GitHubNixos
     FetchAllChannels -->|3 Update Repo| LocalGitCheckout
     FetchAllChannels -->|4 update channels| PostgreSQL
-    PostgreSQL -.->|NixEvaluationChannel NOTIFY| NixEval
+    FetchAllChannels -.->|triggers evaluation| NixEval
     NixEval -->|ingest derivations| PostgreSQL
 
     SystemdTimerCVEs -.->|Triggers Daily| IngestCVEs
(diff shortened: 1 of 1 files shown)


docs/architecture.mermaid
fricklerhandwerk:
```suggestion
          NixEval["Evaluate Nixpkgs"]
```

docs/architecture.mermaid
```diff
@@ -43,7 +43,9 @@ graph TB
     FetchAllChannels -->|1 Fetch Channels| NixMonitoring
     FetchAllChannels -->|2 Git pull| GitHubNixos
     FetchAllChannels -->|3 Update Repo| LocalGitCheckout
-    FetchAllChannels -->|4 Evaluate Nix| NixEval --> NixStore
+    FetchAllChannels -->|4 update channels| PostgreSQL
+    PostgreSQL -.->|NixEvaluationChannel NOTIFY| NixEval
```
fricklerhandwerk: Hm, not sure whether it's helpful to show that the DB is using NOTIFY to start eval. I'd say it's an implementation detail one layer below what we're showing here. What's actually important is that channel bumps trigger evaluation.

fricklerhandwerk (changes requested):

docs/architecture.mermaid
DarshanCode2005: Agreed, NOTIFY is an implementation detail. Updated the diagram so fetch_all_channels still shows step 4 updating channels in PostgreSQL, and a dotted edge from fetch_all_channels to the evaluator labeled “triggers evaluation”. Kept “ingest derivations” into PostgreSQL on the evaluator side.

fricklerhandwerk (approved):

PR #1062, opened 2026-05-30
fix: restore publication date on published issues

### LINKED ISSUE #993: Published issues no longer show the publication date
Image

Publication still appears in the activity log though

TheCodedKid:
fix: restore publication date on published issues

  Published issues stopped showing their publication date in the issue
  view, while it still appeared in the activity log.

  The issue component template rendered the date from `issue.created`,
  but that field was renamed to `created_at`.

  Add a regression test that publishes an issue and asserts the date is
  rendered, so a future field rename can't reintroduce the silent failure.

  Closes #993

### DIFF
--- a/src/webview/templates/components/issue.html
+++ b/src/webview/templates/components/issue.html
@@ -1,12 +1,14 @@
+{% load humanize %}
 {% load viewutils %}
 
 <article class="rounded-box column gap" id="issue-{{ issue.code }}">
 
   <div class="row gap spread">
     <div class="uppercase bold">{{ issue.code }}</div>
-    <div>
+    <div id="issue-{{ issue.code }}-github">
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Thank you @TheCodedKid, great job! Love those little quality of life improvements.

I took the liberty to make some small fixups that would have taken longer to discuss:
- render as relative time to make it consistent with the rest

  Good follow-up would be to allow toggling to absolute time, as with the activity log, but definitely out of scope for this change. It would either require a bit of refactoring to avoid repeating the hard-coded script we're using in the activity log, or waiting for @florentc to replace how we're rendering the frontend in the first place. 
- remove comments that repeat what the code already says
- since the wording change "GitHub issue published \<relative time\>" doesn't allow matching on "published" alone (there's another "published" somewhere), added an `id` to grab the element directly and adapted the regex accordingly.

fricklerhandwerk (approved):

PR #1067, opened 2026-06-01
docs: add architecture diagram implementation map

DarshanCode2005:
- Added a **Diagram to implementation map** section in `docs/README.md` linking labeled diagram components to their source files.
- Added Mermaid `click` handlers on key diagram nodes in `docs/architecture.mermaid` pointing at GitHub blob URLs on `main`.

### Why
The architecture issue asked for links from diagram boxes to the code that implements them, in a form that works for readers and for link checking. The README list uses normal Markdown links (lychee checks `*.md`). Diagram `click` URLs are an optional convenience when viewing on GitHub.

Related to #931

### DIFF
--- a/docs/architecture.mermaid
+++ b/docs/architecture.mermaid
@@ -67,3 +67,9 @@ graph TB
     class SystemdTimerChannels,SystemdTimerCVEs,NixEval,MatchingListener backgroundClass
     class PostgreSQL,LocalGitCheckout,NixStore storageClass
     class Storage,Background,ManageCommands,Web subgraphClass
+
+    click FetchAllChannels "https://github.com/NixOS/nix-security-tracker/blob/main/src/shared/management/commands/fetch_all_channels.py" "fetch_all_channels"
+    click IngestCVEs "https://github.com/NixOS/nix-security-tracker/blob/main/src/shared/management/commands/ingest_delta_cve.py" "ingest_delta_cve"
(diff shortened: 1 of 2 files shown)


docs/architecture.mermaid
```diff
@@ -67,3 +67,9 @@ graph TB
     class SystemdTimerChannels,SystemdTimerCVEs,NixEval,MatchingListener backgroundClass
     class PostgreSQL,LocalGitCheckout,NixStore storageClass
     class Storage,Background,ManageCommands,Web subgraphClass
+
+    click FetchAllChannels "https://github.com/NixOS/nix-security-tracker/blob/main/src/shared/management/commands/fetch_all_channels.py" "fetch_all_channels"
```
fricklerhandwerk, line 71: Will that work for relative paths as well? I'm a bit wary of having this information twice. It's not really discoverable from the diagram, and it's weird/non-obvious that it maps back to the diagram when looking at the README in isolation (which one does on first encounter).

Ideally we'd link to the files directly from the diagram. Unless that's for some reason impossible, then we need to reconsider.

DarshanCode2005: Relative click targets don’t work on GitHub (verified). Absolute blob URLs work when viewing architecture.mermaid directly.

fricklerhandwerk (approved):

docs/architecture.mermaid
fricklerhandwerk: Okay let's do it that way then.

PR #1072, opened 2026-06-02
docs: add contributor quickstart and onboarding hub

DarshanCode2005:
## Description

Adds a short path for new contributors to run the tracker locally without reading the full contributing guide.

- New [docs/QUICKSTART.md](docs/QUICKSTART.md): `nix-shell`, PostgreSQL (link to CONTRIBUTING for now), `manage migrate`, `manage runserver`, optional GitHub setup, and three quick verification checks
- [docs/README.md](docs/README.md): onboarding section at the top linking Quickstart and CONTRIBUTING

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,6 +6,7 @@ This document is for anyone wanting to contribute to the implementation of the N
 
 Resources to help you get started:
 
+- [**Quickstart guide**](./docs/quickstart.md): Set up a database, run the service locally.
 - [**Architecture Overview**](docs/README.md): High-level system design and component interaction.
 - [**Architecture Diagram**](docs/architecture.mermaid): Visual representation of the system (Mermaid source).
 - [**Design Documents**](docs/design/): Detailed design specifications for individual features (E.g., linkage).
@@ -25,8 +26,6 @@ From here, it follows standard Django patterns:
 
 The service is implemented in Python using [Django](https://www.djangoproject.com/).
 It is built and deployed with [Nix](https://nix.dev).
-
-To get going, all you need is to [install Nix](https://nix.dev/install-nix).
 
 ## Running the service in a development environment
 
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
Generally, how about moving what we have about what would be strictly quickstart over into the separate document? As we discussed, rewriting existing text tends to be problematic, because each item is already rather condensed and unlikely to get better without substantial effort. But once one workflow is in place, one can work on the points of friction systematically. Alternatively, one can do the same extraction with another workflow (such as setting up the GitHub org/app/repo...).

DarshanCode2005:
> Generally, how about moving what we have about what would be strictly quickstart over into the separate document? As we discussed, rewriting existing text tends to be problematic, because each item is already rather condensed and unlikely to get better without substantial effort. But once one workflow is in place, one can work on the points of friction systematically. Alternatively, one can do the same extraction with another workflow (such as setting up the GitHub org/app/repo...).

I agree with moving the strictly quickstart-related content out of CONTRIBUTING rather than paraphrasing it in place. I'll push an update that extracts those sections and leaves pointers from CONTRIBUTING.

While working on that, I wanted to share one observation about the current quickstart draft: as a first-run document, it's easier to follow. The numbered steps, clear "do X -> now you have Y" progression, and the fact that optional pieces (direnv, GitHub setup, full ingest, etc.) are deferred make it easier for a newcomer to build confidence and reach a working setup. The extracted version will be accurate and avoid duplication, but it may read more like a subset of the contributing guide than a dedicated onboarding flow.

I don't think that implies rewriting the underlying instructions. My suggestion would be to add a thin layer on top of the extracted content: a short happy path up front (example: nix-shell -> Postgres -> migrate -> runserver -> localhost), along with a few outcome checkpoints. The moved CONTRIBUTING sections could then remain verbatim underneath, or be placed in collapsible details blocks.
Happy to do that as a follow-up to this PR if you think it's worthwhile. I mainly wanted to flag the potential UX gap while the structure is still evolving.

fricklerhandwerk:
Yes, we already agreed that we should pick the relevant bits from the current CONTRIBUTING file and put them into workflow-oriented sequences. Possibly a brief overview would be good to, but let's take it step by step. If the respective parts are small and cluster well into sections, there's no point making summaries, readers can just use the generated TOC.

PR #1077, opened 2026-06-03
docs: document setting OAuth callback URL

fricklerhandwerk:


DarshanCode2005 (approved):

DarshanCode2005:
I am not sure, what's the reason behind the underline in the preview diff.

fricklerhandwerk:
Seems to be a GitHub rendering bug. The final rendering is authoritative. That the note admonition doesn't render is arguably also a rendering bug, it shouldn't matter at which level of indentation it lives. I expect it to go away when we split it into a separate document.

Erethon (approved):

Issue #1039, opened 2026-05-18
Generate API token


fricklerhandwerk:
Since we're moving towards accessing the business logic via API, we need authentication to work there. Make it work.

## Acceptance criteria

Assume the  user is logged in via GitHub OAuth
                                                                                                                                                                                                                                        
**Scenario: Generate a token**
Given the user has no active API token
When the user clicks "Generate API token"
Then a new token is created with a default lifetime of `n` days (fixed at single source of truth)
And the token value is displayed in full exactly once
And a notice informs the user that the value will not be shown again                                                                                                                                                                  

**Scenario: Token already exists**                                                                                                                                                                                                          
Given the user has an active API token
When the user visits the token management page                                                                                                                                                                                        
Then they see the token's creation date and expiry date
But they do not see the token value                                                                                                                                                                                                   

**Scenario: Revoke a token**                                                                                                                                                                                                                
Given the user has an active API token
When the user clicks "Revoke"
Then the token is deleted                                                                                                                                                                                                             
And any API request using that token returns 401
                                                                                                                                                                                                                                        
**Scenario: Extend a token**                                                                                                                                                                                                                
Given the user has an active API token
When the user clicks "Extend by `n` days"                                                                                                                                                                                                         
Then the token's expiry is pushed forward by `n` days from today
                                                                                                                                                                                                                                        
**Scenario: Expired token**
Given the user has a token whose expiry date is in the past                                                                                                                                                                           
When an API request is made with that token                                                                                                                                                                                           
Then the response is 401
                                                                                                                                                                                                                                        
**Scenario: Authenticate an API request**
Given the user has a valid unexpired API token
When a request is made with "Authorization: Bearer <token>"                                                                                                                                                                           
Then the request is authenticated as that user

nic11:
Hey @fricklerhandwerk, I just stumbled upon this project and tried logging in to try subscribing to the updates on my packages. However from the [login page](https://tracker.security.nixos.org/accounts/github/login/?next=%2F), I got redirected to a 404 page, specifically the full URL (redacted) here, as well as the screenshot prior to the redirect:

<details>

<img width="780" height="170" alt="Image" src="https://github.com/user-attachments/assets/061a4d79-d0e3-4f1c-aa65-1c5a07e6a74f" />

```
[...]

PR #1078, opened 2026-06-03
feat: API tokens control panel and API support

florentc:
This implements #1039

### DIFF
--- a/src/api/tests/test_token_auth.py
+++ b/src/api/tests/test_token_auth.py
@@ -1,0 +1,107 @@
+from collections.abc import Callable
+from datetime import timedelta
+
+import pytest
+from django.contrib.auth.models import User
+from freezegun import freeze_time
+from knox.models import AuthToken
+from rest_framework.test import APIClient
+
+from shared.models.linkage import CVEDerivationClusterProposal
+
+# A public (AllowAny) endpoint that returns 404 for non-existent objects,
+# used to probe authentication without needing real fixture data.
+PROBE_URL = "/api/v1/suggestions/999999/status"
+
+
+@pytest.mark.django_db
+def test_bearer_auth_valid_token(
+    user: User, make_token: Callable[..., tuple[AuthToken, str]]
+) -> None:
+    _, raw = make_token(user)
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION=f"Bearer {raw}")
+    response = client.get(PROBE_URL)
+    # 404 means the auth worked (no 401/403)
+    assert response.status_code == 404
+
+
+@pytest.mark.django_db
+def test_bearer_auth_invalid_token() -> None:
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION="Bearer invalidtoken")
+    response = client.get(PROBE_URL)
+    assert response.status_code == 401
+
+
+@pytest.mark.django_db
+def test_bearer_auth_expired_token(
+    user: User, make_token: Callable[..., tuple[AuthToken, str]]
+) -> None:
+    token_obj, raw = make_token(user)
+    assert token_obj.expiry is not None
+    with freeze_time(token_obj.expiry + timedelta(seconds=1)):
+        client = APIClient()
+        client.credentials(HTTP_AUTHORIZATION=f"Bearer {raw}")
+        response = client.get(PROBE_URL)
+    assert response.status_code == 401
+
+
+@pytest.mark.django_db
+def test_bearer_auth_revoked_token(
+    user: User, make_token: Callable[..., tuple[AuthToken, str]]
+) -> None:
+    _, raw = make_token(user)
+    AuthToken.objects.filter(user=user).delete()
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION=f"Bearer {raw}")
+    response = client.get(PROBE_URL)
+    assert response.status_code == 401
+
+
+@pytest.mark.django_db
+def test_bearer_auth_missing_header() -> None:
+    """No Authorization header → anonymous; AllowAny endpoints still respond."""
+    client = APIClient()
+    response = client.get(PROBE_URL)
+    assert response.status_code == 404
+
+
+@pytest.mark.django_db
+def test_bearer_auth_malformed_header() -> None:
+    client = APIClient()
+    client.credentials(HTTP_AUTHORIZATION="Token sometoken")
+    response = client.get(PROBE_URL)
+    # Not a Bearer header → treated as anonymous → AllowAny endpoint → 404
+    assert response.status_code == 404
+
+
+@pytest.mark.parametrize(
+    ("actor_fixture", "expected_status"),
+    [
+        ("user", 403),
+        ("staff", 200),
+    ],
+)
+@pytest.mark.django_db
+def test_bearer_auth_change_status_permissions(
+    request: pytest.FixtureRequest,
+    actor_fixture: str,
+    expected_status: int,
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    make_token: Callable[..., tuple[AuthToken, str]],
+) -> None:
+    """Bearer token: regular users get 403, staff can change status (200)."""
+    actor: User = request.getfixturevalue(actor_fixture)
+    suggestion = make_cached_suggestion()
+    _, raw = make_token(actor)
+    client = APIClient()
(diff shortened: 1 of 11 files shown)


src/project/settings.py
```diff
@@ -432,6 +432,10 @@ class AppSettings(BaseModel):
 REST_FRAMEWORK = {
     "DEFAULT_FILTER_BACKENDS": ["django_filters.rest_framework.DjangoFilterBackend"],
     "EXCEPTION_HANDLER": "api.exceptions.custom_exception_handler",
+    "DEFAULT_AUTHENTICATION_CLASSES": [
+        "api.authentication.BearerTokenAuthentication",
```
fricklerhandwerk: Why not
```suggestion
        "rest_framework.authtoken.authentication.TokenAuthentication",
```

https://www.django-rest-framework.org/api-guide/authentication/#setting-the-authentication-scheme
https://www.django-rest-framework.org/api-guide/authentication/#tokenauthentication

Then we don't need to implement the auth class ourselves.

src/shared/migrations/0091_add_api_token.py
fricklerhandwerk, line 1: Why not use `from rest_framework.authtoken.models import Token`?

src/webview/tests/test_token_views.py
```diff
@@ -0,0 +1,80 @@
+import pytest
+from django.contrib.auth.models import User
+from django.test import Client
```
fricklerhandwerk: Why not

```suggestion
from rest_framework.test import APIClient
```

You can access response data directly as dict there.

fricklerhandwerk (commented):
Nice, but we shouldn't reimplement our own auth stuff. `rest_framework` already ships that.

src/webview/tokens/views.py
```diff
@@ -0,0 +1,47 @@
...
+
+        if action == "revoke":
+            APIToken.objects.filter(user=request.user).delete()
+            return redirect(reverse("webview:tokens:manage"))
+
+        if action == "extend":
+            try:
+                token_obj = request.user.api_token
+                token_obj.extend()
+            except APIToken.DoesNotExist:
```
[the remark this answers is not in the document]
fricklerhandwerk:
```suggestion
            except APIToken.DoesNotExist:
                # Ignore invalid requests
```
Maybe satisfy the pedantic machine?

src/shared/templates/base.html
```diff
@@ -51,6 +51,7 @@ <h1>
             <a href="{% url 'webview:subscriptions:center' %}"
                id="subscriptions-navbar-icon"></a>
             {% notifications_badge user.profile.unread_notifications_count %}
+            <a href="{% url 'webview:tokens:manage' %}">API token</a>
```
fricklerhandwerk: Not sure we should put the link on the front page. If we decide to move it somewhere else later, it may be perceived as disruptive by those who got used to it.

At least we could make it more generic, and then also use that page later to display or link the reference documentation or whatever.

```suggestion
            <a href="{% url 'webview:tokens:manage' %}">API</a>
```

src/project/settings.py
florentc: But there's no support for expiry. The docs suggest (in the notes box at the beginning) to use https://github.com/jazzband/django-rest-knox
Or we can drop expiry and just keep manual revoke

src/shared/migrations/0091_add_api_token.py
florentc: No expiry

src/webview/tests/test_token_views.py
florentc: Done

src/shared/templates/base.html
florentc: Done

src/project/settings.py
fricklerhandwerk: That library does look good! Before we implement the UI on top of REST, we'd need to wrap its a bit for our views, and in any case implement a "revoke single token" method with its primitives, but otherwise it does exactly what we need.

florentc:
Reimplemented with `django-rest-knox`

src/api/tests/test_token_auth.py
```diff
@@ -0,0 +1,110 @@
...
+from django.contrib.auth.models import User
+from freezegun import freeze_time
+from knox.models import AuthToken
+from rest_framework.test import APIClient
+
+from shared.models.linkage import CVEDerivationClusterProposal
+
+pytest_plugins = ["shared.tests.conftest"]
+
+TOKEN_LIFETIME_DAYS = 30
```
fricklerhandwerk: Nit (ignore if you want): We can get this from settings, so it doesn't surprise as being arbitrary.

src/webview/tokens/views.py
```diff
@@ -0,0 +1,56 @@
...
+from django.contrib.auth.mixins import LoginRequiredMixin
+from django.contrib.auth.models import User
+from django.http import HttpRequest, HttpResponse
+from django.shortcuts import redirect
+from django.urls import reverse
+from django.utils import timezone
+from django.views.generic import TemplateView
+from knox.models import AuthToken
+
+TOKEN_LIFETIME_DAYS = 30
```
fricklerhandwerk: Artefact? Should use settings.

src/shared/tests/test_api_token.py
```diff
@@ -0,0 +1,72 @@
...
+TOKEN_LIFETIME_DAYS = 30
+
+
+def _create_token(user: User) -> tuple[AuthToken, str]:
+    return AuthToken.objects.create(  # type: ignore[return-value]
+        user=user, expiry=timedelta(days=TOKEN_LIFETIME_DAYS)
+    )
+
+
+@pytest.mark.django_db
```
fricklerhandwerk:
```suggestion
```
 This is not necessary because `user` already requests `db`. And if it's needed we follow the pattern where we add a fixture `db: None` (instead of a decorator) to make it uniform.

src/shared/tests/test_api_token.py
fricklerhandwerk, line 1: Hm, this looks like testing the library. Unless there's something we add on top (in which case we'd only test that), I guess let's drop it.

src/api/tests/test_token_auth.py
```diff
@@ -0,0 +1,110 @@
...
+
+# A public (AllowAny) endpoint that returns 404 for non-existent objects,
+# used to probe authentication without needing real fixture data.
+PROBE_URL = "/api/v1/suggestions/999999/status"
+
+
+def _create_token(user: User) -> tuple[AuthToken, str]:
```
fricklerhandwerk: Why not a fixture?

src/api/tests/test_token_auth.py
```diff
@@ -0,0 +1,109 @@
...
+
+import pytest
+from django.contrib.auth.models import User
+from freezegun import freeze_time
+from knox.models import AuthToken
+from rest_framework.test import APIClient
+
+from shared.models.linkage import CVEDerivationClusterProposal
+
+pytest_plugins = ["shared.tests.conftest"]
```
fricklerhandwerk: This should not be necessary anywhere. It exists in each app's `conftest.py` already and is this active automatically in every test of those apps.

src/webview/tests/test_token_views.py
```diff
@@ -0,0 +1,88 @@
...
+
+pytest_plugins = ["shared.tests.conftest"]
+
+TOKEN_LIFETIME_DAYS = 30
+
+
+def _create_token(user: User) -> tuple[AuthToken, str]:
+    return AuthToken.objects.create(  # type: ignore[return-value]
+        user=user, expiry=timedelta(days=TOKEN_LIFETIME_DAYS)
+    )
```
fricklerhandwerk: Stale

src/webview/tokens/views.py
```diff
@@ -0,0 +1,55 @@
...
+from django.shortcuts import redirect
+from django.urls import reverse
+from django.utils import timezone
+from django.views.generic import TemplateView
+from knox.models import AuthToken
+
+
+def _expiry() -> timedelta:
+    return settings.REST_KNOX["TOKEN_TTL"]
+
```
fricklerhandwerk: Why not 

```suggestion
expiry = settings.REST_KNOX["TOKEN_TTL"]

```

fricklerhandwerk (approved):

fricklerhandwerk (approved):

PR #1082, opened 2026-06-05
fix: Fix a bug in an old backward migration

Erethon:
Realistically, this code will never run but let's fix this bug so it doesn't show up in code scans.


fricklerhandwerk (approved):
Wow, how did that even run in the first place?

Erethon:
> Wow, how did that even run in the first place?

It never run, it's a backwards migration. But it's good we got rid of that (I introduced it), since it makes GitHub code scan happy.

PR #1085, opened 2026-06-07
feat: integrate django-prometheus for web serivce metrics

### LINKED ISSUE #1037: Web service metrics
Wire up https://github.com/django-commons/django-prometheus with our code and Grafana so we can observe production behavior better.

devnchill:
Closes #1037

### DIFF
--- a/contrib/grafana-dashboard.json
+++ b/contrib/grafana-dashboard.json
@@ -15,7 +15,7 @@
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "12.1.1"
+      "version": "13.0.2"
     },
     {
       "type": "datasource",
@@ -27,6 +27,12 @@
       "type": "panel",
       "id": "stat",
       "name": "Stat",
+      "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "timeseries",
+      "name": "Time series",
       "version": ""
     }
   ],
@@ -49,8 +55,8 @@
   "editable": true,
   "fiscalYearStartMonth": 0,
   "graphTooltip": 0,
-  "id": null,
   "links": [],
+  "liveNow": false,
   "panels": [
     {
       "collapsed": false,
@@ -60,8 +66,7 @@
         "x": 0,
         "y": 0
       },
-      "id": 4,
-      "panels": [],
+      "id": 17,
       "title": "Main stats",
       "type": "row"
     },
@@ -76,7 +81,6 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
@@ -117,7 +121,7 @@
         "textMode": "auto",
         "wideLayout": true
       },
-      "pluginVersion": "12.1.1",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -149,7 +153,6 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
@@ -190,7 +193,7 @@
         "textMode": "auto",
         "wideLayout": true
       },
-      "pluginVersion": "12.1.1",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -222,7 +225,6 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
@@ -263,7 +265,7 @@
         "textMode": "auto",
         "wideLayout": true
       },
-      "pluginVersion": "12.1.1",
+      "pluginVersion": "13.0.2",
       "targets": [
         {
           "datasource": {
@@ -295,7 +297,6 @@
           "color": {
             "mode": "thresholds"
           },
-          "mappings": [],
           "thresholds": {
             "mode": "absolute",
             "steps": [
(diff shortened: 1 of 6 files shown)


devnchill:
hello @fricklerhandwerk , I wanted to clear this point that I have to connect grafana just for us to see how metrics would show  up in grafana right ? I don't have to modify infra side for this , just the project side and then use nix to create test vm where grafana would be setup and visualize the metrics coming from the project ?

Erethon:
> hello @fricklerhandwerk , I wanted to clear this point that I have to connect grafana just for us to see how metrics would show up in grafana right ? I don't have to modify infra side for this , just the project side and then use nix to create test vm where grafana would be setup and visualize the metrics coming from the project ?

Hello! Technically you only need to setup Prometheus in a test VM to collect the metrics, no Grafana is needed. If you want to test the changes you made, I think adding a VM test or python test to check that the metrics endpoint (`/metrics`) returns one of the default django-prometheus metrics is more than enough. Spinning up Prometheus and Grafana for a test would be overkill in my opinion, but maybe something that some people would like to have in a dev environment.

fricklerhandwerk:
Yeah, having a way to at least manually check out of band what a given change is like would be good. I thought about maybe just adding instances to our clumsy NixOS dev module. It's dumb but it gets you the thing quickly.

But for this change, a check in the smoke test should be enough.

devnchill:
got it

devnchill:
ig i'll wait for #1131  to merge ? would help me in testing :laughing:

fricklerhandwerk:
In the meantime or later on I'd strongly recommend going through the contributing guide and exercising the workflows to obtain data locally. Such as "fetch some channels and run an evaluation". Because without data you don't have anything to look at in the dashboards.

Likely there are some papercuts there, such as for how to get a reasonable number of CVEs for testing. It would be such a great contribution if you extracted one such workflow as a separate guide like @DarshanCode2005 did in #1072. Right now these bits and pieces are scattered.

devnchill:
Makes sense, I'll work on that in the meantime :laughing:

devnchill:
hello @fricklerhandwerk , I first saved the new dashboard *as copy* so due to that the version i believe started counter from  `0` , i noticed the prev dashboard had `33` , so I am not sure if i have to replace current `version 4` with prev one , same goes for uid, it changes on each save i am assuming . i wasn't sure if changing id brings any breaking changes . if it does ig , we'll have to reuse old one ? 

Oh also from prev discussion it seemed that there were 2 ways to add test for this, one was through pytest and another was through nix test , atm I picked django , do I need to add test for nix vm too ? 

apart from that this pr had so much of learning experience lol , learned several new stuff because of this :smile:  .

fricklerhandwerk (approved):

fricklerhandwerk:
Since we're exposing directly from Django, the unit test is perfectly fine. I still need to get the Grafana dev setup working again (not sure what got messed up before I merged it, but it worked and now doesn't) to try it out.

Not sure the stray changes to the dashboard JSON are all needed though. @erethon opinions?

devnchill:
Hmm weird, Grafana dev setup was working for me couple hrs ago though, will try it again when I go home.

Regarding the dashboard, I thought you mentioned in initial issue that it would be nice to connect it to Grafana for quick lookup locally ..  So I thought this is what you meant from that ...  Mb for misunderstanding 😅

I'll remove dashboard changes when I reach home

fricklerhandwerk:
Oh the dashboard changes adding the data are fine, I mean all the unrelated stuff such as reformatting.

devnchill:
Oh my nvim is configured to format on save so I believe it used prettier for that when I saved the file

Erethon:
> Not sure the stray changes to the dashboard JSON are all needed though. @Erethon opinions?

The `plugin_version` changes look fine to me, I guess they're coming from a newer Grafana version. The nvim auto-formatting will be reverted in a future export, so I suggest @devnchill you re-export the dashboard without the auto-formatting.

contrib/grafana-dashboard.json
```diff
@@ -875,24 +850,258 @@
...
+      },
+      "pluginVersion": "13.0.2",
+      "targets": [
+        {
           "datasource": {
             "type": "prometheus",
             "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "builder",
+          "expr": "django_http_requests_latency_seconds_by_view_method_count",
```
Erethon, line 985: Please add a filter to distinguish between the staging/production instance or any other instances of `django_http_requests_latency_seconds_by_view_method_count`

contrib/grafana-dashboard.json
```diff
@@ -875,24 +850,258 @@
...
+      },
+      "pluginVersion": "13.0.2",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "builder",
+          "expr": "django_http_responses_total_by_status_total",
```
Erethon, line 1102: Please add a filter to distinguish between the staging/production instance or any other instances of `django_http_responses_total_by_status_total`.

Erethon (changes requested):
I've left two inline comments about the changes to the Grafana dashboard. If we don't care to have everything in the same page, we could re-use the dashboard from [this PR](https://github.com/django-commons/django-prometheus/pull/318) (I have no idea why it wasn't merged :().

contrib/grafana-dashboard.json
devnchill: okay :sweat_smile:

devnchill: hello @Erethon , I have a doubt , 
`"expr": "django_http_requests_latency_seconds_by_view_method_count{instance=\"$Instance\"}",`if i use this, then it would show empty results no?  because `label` is instance scraping the `sql_delta` metric and since sql and django run on different port, label would end up becoming different for both `sql` and `django` no? any advice ?

Issue #626, opened 2025-07-25
Track latest (successful) evaluations in Grafana


fricklerhandwerk:
In order to prevent things like https://github.com/Nix-Security-WG/nix-security-tracker/pull/625 in the future.

Show a table with the latest evaluation and latest *successful* evaluation for each channel.

PR #1093, opened 2026-06-09
feat: add latest completed evals

DarshanCode2005:
This PR is related to the issue #626 , I will try to implement the changes in order and small changes in particular.

### DIFF
--- a/src/shared/models/nix_evaluation.py
+++ b/src/shared/models/nix_evaluation.py
@@ -4,7 +4,7 @@ from django.contrib.postgres.indexes import BTreeIndex, GinIndex
 from django.contrib.postgres.search import SearchVectorField
 from django.core.exceptions import ObjectDoesNotExist
 from django.db import models
-from django.db.models import F, Q, Window
+from django.db.models import F, Window
 from django.db.models.functions import RowNumber
 from django.utils.translation import gettext_lazy as _
 from pgtrigger import UpdateSearchVector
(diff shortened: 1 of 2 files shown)


fricklerhandwerk:
How about adding a model manager instead? Then you can replace the access of latest evaluations in the same commit; we don't want to create dead code. If it's not exercised, how do you know (in Python) that it works?

DarshanCode2005:
> How about adding a model manager instead? Then you can replace the access of latest evaluations in the same commit; we don't want to create dead code. If it's not exercised, how do you know (in Python) that it works?

Moved the query to `NixEvaluationQuerySet.latest_completed_per_channel()` and wired it into automatic_linkage.py in the same commit. `test_linkage.py` covers the behavior.

src/shared/models/nix_evaluation.py
```diff
@@ -183,6 +185,22 @@ def is_rolling_release(self) -> bool:
         return self.channel_branch == settings.TRACKING_BRANCH
 
 
+class NixEvaluationQuerySet(models.QuerySet):
+    def latest_completed_per_channel(
+        self, channel_filter: Q | None = None
```
fricklerhandwerk: Instead of adding a parameter, the caller can just append filters or whatever to the returned query -- this is just as natural here, and a lot simpler because we don't need conditionals.

fricklerhandwerk: It would also make for a more obvious diff. I did a double-take because had I expected the lines to have simply moved.

fricklerhandwerk (changes requested):

src/shared/models/nix_evaluation.py
DarshanCode2005: callers can chain their own filters now.

fricklerhandwerk (approved):

PR #1096, opened 2026-06-09
fix: prevent overriding of credentials file everytime someone enters …

devnchill:
```sh
/nix/store/r6vn1bhg3zn9inbph9rv25n5cjxfvwca-shell-config-placeholder/bin/shell-config-placeholder: line 14: SECRET_KEY: cannot overwrite existing file
```
Because of `set -o noclobber`, I was receiving the error `SECRET_KEY: cannot overwrite existing file. shell-config-placeholder` previously created files in .credentials unconditionally every time someone entered the nix shell. This PR fixes it by first checking for their existence and only creating them when they are not present. This also prevents overriding file contents every time someone re-enters the shell, keeping the key stable across sessions.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -77,12 +77,12 @@ rec {
           mkdir -p "$nixpkgsDir"
           cd "$credentialsDir"
           set -o noclobber
-          python3 -c 'import secrets; print(secrets.token_hex(100))' > SECRET_KEY
-          echo bar > GH_CLIENT_ID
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
This is a duplicate of https://github.com/NixOS/nix-security-tracker/pull/1087, and that one is less repetitive. I suggest you take a look :)

devnchill:
Oops mb.I'll close this one

PR #1094, opened 2026-06-09
refactor: use latest_completed_evaluations in automatic_linkage

DarshanCode2005:
Depends on PR1.

PR #1097, opened 2026-06-09
refactor: generalize latest evaluation query per channel

DarshanCode2005:
Depends on #1093 and #1094.


DarshanCode2005:
only review after #1093 / #1094 merge; will rebase.

PR #1102, opened 2026-06-09
feat: add AdminRequiredMixin for staff-only views

DarshanCode2005:
From this PR onwards I will be implementing Staff-only status page related to #626

### DIFF
--- a/src/shared/auth/mixins.py
+++ b/src/shared/auth/mixins.py
@@ -1,0 +1,8 @@
+from django.contrib.auth.mixins import LoginRequiredMixin, UserPassesTestMixin
+
+from shared.auth.utils import isadmin
+
+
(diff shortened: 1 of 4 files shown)


DarshanCode2005:
After the previous changes are reviewed then I will rebase.

fricklerhandwerk:
What exactly do we need private views for?

DarshanCode2005:
> What exactly do we need private views for?

I hope that stack trace won't contain any sensitive information. Otherwise we can drop this.

fricklerhandwerk:
Unless we explicitly store the stack trace (which we currently don't), it can't show anywhere. Sysadmins currently get them by email. For metrics it's enough to record the events, which should happen via Prometheus (see a related issue from recently, I don't have a keyboard to look it up quickly).

PR #1103, opened 2026-06-09
feat: skip derivation with known vulnerabilities on CVE matching

### LINKED ISSUE #1063: Don't open reports for packages with `knownVulnerabilities`
see https://github.com/NixOS/nixpkgs/issues/526352#issuecomment-4588256471

devnchill:
Closes #1063 

- Checks within metadata of derivation for matching CVE id in known vulneliribilities 
- skips if they match 
- bump matching algorithm from `0` to `1` 

Not sure if it had to be 1 or something else ...


src/shared/listeners/automatic_linkage.py
```diff
@@ -112,6 +112,12 @@ def produce_linkage_candidates(
         .annotate(**annotations)
     )
     for drv in matches.iterator():
+        if (
```
fricklerhandwerk: Pretty sure we can do that right in the query.

fricklerhandwerk:
Don't forget tests! No way to be sure it does what it's supposed to if it doesn't run through a machine verification. And it makes review a lot easier: I can just look at the test and whether CI is green. Pulling the code and doing things manually takes time and is error-prone. And quick review means quick merge.

devnchill:
yeah writing test did occur but thought would be an overkill :sweat_smile: ....

devnchill:
I reworded the commit due to incorrect spelling :sweat_smile: , should I squash them ?

src/shared/listeners/automatic_linkage.py
```diff
@@ -103,7 +103,8 @@ def produce_linkage_candidates(
     matches = (
         NixDerivation.objects.exclude(
             # Test derivations are a Nixpkgs implementation detail and never represent software to be distributed.
-            attribute__startswith="tests.",
+            Q(attribute__startswith="tests.")
+            | Q(metadata__known_vulnerabilities__contains=[container.cve.cve_id])
```
fricklerhandwerk:
```suggestion
            attribute__startswith="tests.",
            metadata__known_vulnerabilities__contains=[container.cve.cve_id],
```

This should work just as well

fricklerhandwerk (approved):
LGTM modulo cosmetics, please run `nix-shell --run format` and squash.

src/shared/listeners/automatic_linkage.py
devnchill: wouldn't this do AND ?

fricklerhandwerk: Oh right, forget that...

devnchill:
rebased and squashed it

devnchill:
@fricklerhandwerk 
image
I recieved this error on running format however I beliee this is after rebase so likely due to https://github.com/NixOS/nix-security-tracker/commit/8f3802c38cb03bb8fb07ad7da33cb0517be64094 this commit . Should I create another pr to fix it ? 

```sh
/home/viena/dev/os-contribution/NixOs/nix-security-tracker/src/api/urls.py
  /home/viena/dev/os-contribution/NixOs/nix-security-tracker/src/api/urls.py:2:6 - error: Import "drf_spectacular.views" could not be resolved (reportMissingImports)
1 error, 0 warnings, 0 informations
```

fricklerhandwerk:
Is it possible you didn't restart your shell after rebasing onto main?

fricklerhandwerk:
I just realized, if we ignore those cases, we're throwing away good signals for matching quality. Finding the same CVE tells us how good the other criteria are. WDYT about instead, for example, matching and auto-dismissing with a new condition: already known? Although then we won't be able to track whether it's fixed. But if we auto-publish it without also creating a GitHub issue (and spamming maintainers who already know the problem was the original motivation for the change), we may accidentally add unrelated false-positive matches. And having it go past triagers requires a new publishing mode that skips the GitHub issue... nothing really ideal there.

IMO the best we can do cheaply is record the fact, so we may eventually deal with it. Throwing that information away will make that impossible. Opinions?

devnchill:
so match it first and then reject it later so that its atleast derivations are linked ?

devnchill:
> Is it possible you didn't restart your shell after rebasing onto main?

yeah i am using `direnv` (added .envrc to .git/info/exclude) , had to reload direnv

fricklerhandwerk:
> so match it first and then reject it later so that its atleast derivations are linked ?

We could in fact pick all derivations with that CVE in `knownVulnerabilities` and, if that set is nonempty, link them, and reject the CVE with the new rejection reason. Not sure exactly how to go about that efficiently though. Otherwise do the regular thing.

devnchill:
but what if a derivation matches cve by name and not by `knownvulnerability` ? if we reject cve purely on basis of known vulnerability , wouldn't it let derivation with that cve bypass too even though these doesn't has this cve listed as `knownvulnerability` in their metadata ?

fricklerhandwerk:
Are you saying there may be *more* derivations that match *only* the other criteria, and by auto-rejecting we'd be dropping them? This is of course possible in principle. That would be tricky to deal with, because we'd need to handle them separately somehow.

The problem is that currently it's very likely that additional matches are false positives anyway. So if we don't take the route to auto-reject those with `knownVulnerabilities` and ignore everything else, we'd need to present it to triagers for further review (option 3 from my original overview). Then after all false-positives are gone, if all that's left are the ones with `knownVulnerabilities` we know what to do again.

We'd then also need an option to manually reject as "already known" -- this should be easy, the mechanism exists and just needs another case.

devnchill:
Yeah I was thinking of such cases,like matching by other criteria and not by known vulnerability, I have no idea tbh onif it would be actually severe or just false positive but because you said there's a high chance that it would be a false positive, ig we could ignore them and follow the approach that you suggest above, the auto reject one ...

src/shared/listeners/automatic_linkage.py
```diff
@@ -189,11 +191,24 @@ def build_new_links(container: Container) -> bool:
         )
         return False
 
-    proposal = CVEDerivationClusterProposal.objects.create(
-        cve=container.cve,
-        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION,
+    has_known_vuln = any(
```
fricklerhandwerk: I think that's a fine compromise, but then let's actually only attach those drvs with known vulns. I.e. let's make this a filter, check in the conditional whether the result is nonempty, and in that case assign `drvs = with_known_vulns`.

That way there's nothing left for triagers to review. Otherwise, as discussed above, we'd be leaving things on the table without providing the means to clean them up manually.

fricklerhandwerk (approved):

src/shared/listeners/automatic_linkage.py
devnchill: done :smile:

devnchill:
rebasing on main

src/shared/listeners/automatic_linkage.py
```diff
@@ -89,12 +89,14 @@ def produce_linkage_candidates(
     matches = (
         NixDerivation.objects.exclude(
             # Test derivations are a Nixpkgs implementation detail and never represent software to be distributed.
-            attribute__startswith="tests.",
+            attribute__startswith="tests."
```
fricklerhandwerk:
```suggestion
            attribute__startswith="tests.",
```

No need to change that one

src/shared/listeners/automatic_linkage.py
```diff
@@ -89,12 +89,14 @@ def produce_linkage_candidates(
...
         )
         .filter(
             package_q | product_q,
             parent_evaluation__in=list(latest_complete_channels),
         )
+        # XXX(@devnchill): Pre-fetch metadata to avoid N+1 on known_vulnerabilities check
```
fricklerhandwerk:
```suggestion
```

This is a common pattern, no need to explain it

src/shared/listeners/automatic_linkage.py
```diff
@@ -175,16 +177,32 @@ def build_new_links(container: Container) -> bool:
...
         return False
 
-    proposal = CVEDerivationClusterProposal.objects.create(
-        cve=container.cve,
-        algorithm_version=CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION,
-    )
+    with_known_vuln = {
+        drv: flags
+        for drv, flags in drvs.items()
+        if drv.metadata and container.cve.cve_id in drv.metadata.known_vulnerabilities
```
fricklerhandwerk, line 182: This could literally be a filter via ORM, right? Then the conditional would be `with_known_vuln.exists()`

fricklerhandwerk (commented):
Some CI stuff still failing :) Almost there.

devnchill:
> Some CI stuff still failing :) Almost there.

yeah I was wondering if it was somehow due to my changes T_T

fricklerhandwerk:
Yeah I think you need to regenerate the migration.

devnchill:
rebased over main

fricklerhandwerk:
Great work!

PR #1105, opened 2026-06-10
fix: make evaluations unique per commit

adekoder:
Make evaluation unique per commit,  and added data migration to delete duplicate evaluations.


src/shared/migrations/0092_deduplicate_nixevaluations.py
```diff
@@ -0,0 +1,51 @@
...
+        canonical_id = dup["canonical_id"]
+
+        for eval_obj in NixEvaluation.objects.filter(
+            commit_sha1=commit_sha1
+        ).exclude(id=canonical_id):
+            DerivationClusterProposalLink.objects.filter(
+                derivation__parent_evaluation=eval_obj
+            ).delete()
+    
+            NixDerivation.objects.filter(parent_evaluation=eval_obj).delete()
```
fricklerhandwerk: Are we catching the metadata rows with this one, too? Did you try it locally? There will be ~10M rows to delete in a transaction here.

src/shared/migrations/0092_deduplicate_nixevaluations.py
```diff
@@ -0,0 +1,51 @@
...
+            eval_obj.delete()
+
+    # Delete proposals that have no derivations left after link cleanup.
+    CVEDerivationClusterProposal.objects.annotate(
```
fricklerhandwerk: This is in fact a failure condition. If it happens, our assumptions were wrong (duplicates are all matched the same, and thus after deleting them there should be no empty suggestions) and we need to abort here.

adekoder: i check in stage environment and we have 102 suggestions with no DRVs

```
In [3]: CVEDerivationClusterProposal.objects.annotate(
   ...:         drv_count=Count("derivations")
   ...:     ).filter(drv_count=0).count()
Out[3]: 102
```

src/shared/migrations/0092_deduplicate_nixevaluations.py
adekoder: Good catch

src/shared/migrations/0092_deduplicate_nixevaluations.py
fricklerhandwerk: Some of those will be auto-rejected due to some properties of the CVE without trying to match derivations to them, so that's not a criterion for deleting them.

adekoder: oh okay i guess it is normal to have suggestions with no drvs then, i will remove it

fricklerhandwerk: I'll check production once I get to the keyboard to make sure.

fricklerhandwerk (commented):
As discussed out of band, this needs to be a separate async command, and after that’s in prod we can deploy the uniqueness constraint from a follow-up PR.

src/shared/migrations/0092_deduplicate_nixevaluations.py
fricklerhandwerk: Yep, those with 0 drvs linked are all rejected, as expected.

PR #1107, opened 2026-06-11
feat: add ChannelEvaluationStatus dataclass

DarshanCode2005:
Adds `ChannelEvaluationStatus` in `evaluation_status.py` for #626. Fields only; `is_healthy` and `get_channel_evaluation_statuses()` follow in later PRs.


fricklerhandwerk:
The commit message/PR description really needs to say *why* this is needed.

I suppose it's to have one data structure for what a given channel is like in terms of evaluation? Then why don't we just add properties to the channel model?

PR #1115, opened 2026-06-12
refactor: single source of truth for tracked channel states

fricklerhandwerk:
Depends on https://github.com/NixOS/nix-security-tracker/pull/1122

Related: https://github.com/NixOS/nix-security-tracker/issues/864


DarshanCode2005:
* Could this PR include an updated test, mainly asserting asserting END_OF_LIFE channels don't produce matches, and STABLE/UNSTABLE channels do?
* a one-line comment explaining why STAGING and END_OF_LIFE are omitted would help future readers at one place.

fricklerhandwerk:
I removed the "staging" channel state altogether in https://github.com/NixOS/nix-security-tracker/pull/1122. 

Good catch on the test, the change broke an existing one anyway as it should be, since we're changing behavior here.

PR #1117, opened 2026-06-13
fix: prune package-attribute mappings orphaned by garbage collection

### LINKED ISSUE #1111: Garbage collect stale attribute paths attached to packages
When clustering derivations into packages, we're always accumulating new attribute paths. But once old derivations are garbage-collected (in the periodic management command or for other reasons), their attributes may become stale.

Add a check somewhere (for instance in the garbage-collection command, or maybe eagerly at clustering time) that will prune attribute paths in `PackageAttrpath` that don't correspond to derivations in `PackageDerivation`.

DarshanCode2005:
Package clustering registers `PackageAttrpath` rows as derivations are linked, but when derivations are garbage-collected their `PackageDerivation` rows are removed while the attrpath mappings can remain. This adds a batched prune helper and runs it as the final step of `garbage_collect`.

Closes #1111

### DIFF
--- a/src/shared/tests/test_garbage_collect.py
+++ b/src/shared/tests/test_garbage_collect.py
@@ -6,6 +6,7 @@ from io import StringIO
 import pytest
 from django.core.management import call_command
 
+from shared.management.commands.garbage_collect import DEFAULT_CUTOFF_DAYS
 from shared.models.cve import Container, CveRecord
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
@@ -22,6 +23,8 @@ from shared.models.nix_evaluation import (
     NixEvaluation,
     NixMaintainer,
 )
+from shared.models.package import Package, PackageAttrpath, PackageDerivation
+from shared.package_clustering import cluster_packages
 
 
 def test_cve_record_not_deleted_with_stale_proposal(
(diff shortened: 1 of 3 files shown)


src/shared/package_clustering.py
```diff
@@ -207,3 +207,26 @@ def package_from_attrs(drvs: list[dict]) -> dict[str, Package]:
             attrpath__in=[drv.attribute for drv in drvs]
         ).select_related("package")
     }
+
+
+def stale_package_attrpaths() -> QuerySet[PackageAttrpath]:
```
fricklerhandwerk: This could be on a model manager again.

src/shared/package_clustering.py
```diff
@@ -207,3 +207,26 @@ def package_from_attrs(drvs: list[dict]) -> dict[str, Package]:
...
+    live = PackageDerivation.objects.filter(
+        derivation__attribute=OuterRef("attrpath"),
+    )
+    return PackageAttrpath.objects.filter(~Exists(live))
+
+
+def prune_stale_package_attrpaths(batch_size: int = 10_000) -> int:
```
fricklerhandwerk: Since the GC command has a batching helper, you can just use it and pass it the right input.

src/shared/tests/test_garbage_collect.py
```diff
@@ -348,3 +350,38 @@ def test_when_there_is_duplicate_evaluation(
...
+def test_garbage_collect_prunes_stale_package_attrpaths(
+    make_channel: Callable[..., NixChannel],
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+) -> None:
+    channel = make_channel(state=NixChannel.ChannelState.END_OF_LIFE)
+    evaluation = make_evaluation(
+        channel=channel,
+        state=NixEvaluation.EvaluationState.COMPLETED,
+        age=timedelta(days=400),
```
fricklerhandwerk: Use the setting here, otherwise there's a chance this will break if the concrete value ever changes. (Unlikely of course, but this is easy and guarantees that it won't incur costs in the future.)

src/shared/tests/test_package_clustering.py
```diff
@@ -403,3 +407,38 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
     pkg = PackageAttrpath.objects.get(attrpath="shared").package
     assert PackageDerivation.objects.get(derivation=drv_a).package == pkg
     assert PackageDerivation.objects.get(derivation=drv_b).package == pkg
+
+
+def test_prune_stale_package_attrpaths_removes_orphans(
```
fricklerhandwerk: These duplicate the GC tests. You could move the "preserves linked ones" over and remove them here.

Also naming-wise, the GC operation is not the same as the clustering one, so already on those grounds the tests should live there.

fricklerhandwerk (approved):
Very good in terms of content, just some factoring nits!

PR #1125, opened 2026-06-15
chore: remove duplicate garbage collection of stale matches

fricklerhandwerk:


src/shared/management/commands/garbage_collect.py
```diff
@@ -100,14 +102,30 @@ def handle(self, *args: Any, **options: Any) -> None:
...
-            created_at__lt=cutoff,
-            status=CVEDerivationClusterProposal.Status.PENDING,
-            # No user input must be attached
-            maintainer_overlays__isnull=True,
-            package_overlays__isnull=True,
-            reference_url_overlays__isnull=True,
-        ).distinct()
+        candidates = (
+            CVEDerivationClusterProposal.objects.filter(
+                reduce(
```
adekoder: why do we need to use reduce for this, we can do this directly, much straightforward

``
```suggestion
      .filter(
       Q(created_at__lt=cutoff) | 
       Q(cve__date_published__lt=cutoff) | 
       Q(cve__date_published__isnull=True, cve__date_reserved__lt=cutoff)
  )
```

src/shared/tests/test_garbage_collect.py
```diff
@@ -442,3 +444,115 @@ def test_garbage_collect_dry_run_preserves_stale_package_attrpaths(
 
     assert PackageAttrpath.objects.filter(attrpath=drv.attribute).exists()
     assert "stale package attrpath" in out.getvalue().lower()
+
+
+def test_deletes_proposal_when_cve_published_date_is_old(
```
adekoder: The following tests can be parameterized since the only difference is the date_published

- test_preserves_proposal_when_cve_has_no_published_date_but_recent_reserved_date
- test_preserves_proposal_when_cve_published_date_is_recent
- test_deletes_proposal_when_cve_has_no_published_date_but_old_reserved_date
- test_deletes_proposal_when_cve_published_date_is_old

adekoder (dismissed):

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: Yeah right. I kinda got off the rails after not liking how that variant is formatted automatically. A list of things arranges uniformly, as opposed to a chain of operators.

Will revert.

PR #1127, opened 2026-06-16
feat: add per-channel evaluation status to Grafana

DarshanCode2005:
This PR adds observability for per-channel Nix evaluation health 
* SQL exporter query `channel_evaluation_status` in `infra/common.nix` exports `sql_channel_evaluation_status` with per-channel labels (channel, HEAD, latest state, commits) and timestamps
* Grafana table panel shows latest eval state and latest successful eval for each tracked channel

Table columns
| Column | Source |
| --- | --- |
| Channel | `channel_branch` |
| HEAD | `head_sha1_commit` |
| Latest state | raw enum (`WAITING`, `IN_PROGRESS`, `COMPLETED`, `CRASHED`, `FAILED`) |
| Latest commit | latest eval `commit_sha1` |
| Latest updated | latest eval `updated_at` |
| Latest elapsed | latest eval `elapsed` |
| Latest successful commit | newest `COMPLETED` eval |
| Latest successful updated | newest `COMPLETED` eval `updated_at` |

Channels filtered to DEPRECATED, BETA, STABLE, UNSTABLE (same as ADMISSIBLE_CHANNEL_STATES today).

Future work:
* pghistory on `NixEvaluation` state changes for waiting duration
* Alerts once thresholds decided

### DIFF
--- a/contrib/grafana-dashboard.json
+++ b/contrib/grafana-dashboard.json
@@ -28,6 +28,12 @@
       "id": "stat",
       "name": "Stat",
       "version": ""
+    },
+    {
+      "type": "panel",
+      "id": "table",
+      "name": "Table",
+      "version": ""
     }
   ],
   "annotations": {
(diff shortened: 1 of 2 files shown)


fricklerhandwerk (approved):
LGTM! @Erethon do we have a way to preview this somewhat quickly, short of spinning up our own Grafana?

Erethon:
> LGTM! @Erethon do we have a way to preview this somewhat quickly, short of spinning up our own Grafana?

Not really, it's a two part problem. We need the exporter to export the metrics and then you need a Prometheus instance to scrape them and Grafana to display them. But even having a Grafana instance we could easily add the dashboard [1], we would first need this change to be applied to staging so that metrics are exported. It's a chicken or the egg problem. 

My course of action to test this would be to deploy it manually on staging and see if the metrics look good after that. Having said that, running the SQL query manually on staging returns zero rows, but returns correctly on production! Looking more into it, it's because of https://github.com/NixOS/nix-security-tracker/pull/1122, the 0092 migration changed the `state` column of the NixChannel table, so this PR at the very least needs to be updated to reflect these changes.

Please note that having labels with high cardinality on Prometheus (as are the SHA hashes) will create problems down the road in Prometheus (look at the caution [here](https://prometheus.io/docs/practices/naming/#labels) and read more about it [here](https://www.robustperception.io/cardinality-is-key/)). Do we really need the specific SHAs for each channel? If not, I would suggest we drop them from the labels. If we do need those, we should do some math to figure out how many labels we expect to have, to figure out if it's too much for Prometheus (anything under ~10k should be fine in my experience) and/or see if the [new info function](https://prometheus.io/blog/2025/12/16/introducing-info-function/) coupled with info metrics and a more finely grained retention period for some of our metrics would be good enough.

[1]: I can add it to my personal one and share the URL for testing, that's easy.

Erethon (dismissed):
As mentioned [here](https://github.com/NixOS/nix-security-tracker/pull/1127#issuecomment-4723126783), this PR needs to be updated after the changes in #1122.

DarshanCode2005:
> > LGTM! @Erethon do we have a way to preview this somewhat quickly, short of spinning up our own Grafana?
> 
> Not really, it's a two part problem. We need the exporter to export the metrics and then you need a Prometheus instance to scrape them and Grafana to display them. But even having a Grafana instance we could easily add the dashboard [1], we would first need this change to be applied to staging so that metrics are exported. It's a chicken or the egg problem.
> 
> My course of action to test this would be to deploy it manually on staging and see if the metrics look good after that. Having said that, running the SQL query manually on staging returns zero rows, but returns correctly on production! Looking more into it, it's because of #1122, the 0092 migration changed the `state` column of the NixChannel table, so this PR at the very least needs to be updated to reflect these changes.
> 
> Please note that having labels with high cardinality on Prometheus (as are the SHA hashes) will create problems down the road in Prometheus (look at the caution [here](https://prometheus.io/docs/practices/naming/#labels) and read more about it [here](https://www.robustperception.io/cardinality-is-key/)). Do we really need the specific SHAs for each channel? If not, I would suggest we drop them from the labels. If we do need those, we should do some math to figure out how many labels we expect to have, to figure out if it's too much for Prometheus (anything under ~10k should be fine in my experience) and/or see if the [new info function](https://prometheus.io/blog/2025/12/16/introducing-info-function/) coupled with info metrics and a more finely grained retention period for some of our metrics would be good enough.
> 
> [1]: I can add it to my personal one and share the URL for testing, that's easy.

Thanks for the review. I also agree on the cardinality concern. I'll remove the SHA labels. Will push an update soon.

comment updated :+1:

contrib/grafana-dashboard.json
```diff
@@ -807,6 +813,246 @@
         "x": 0,
         "y": 11
       },
+      "id": 10,
+      "panels": [],
+      "title": "Channel evaluations",
```
fricklerhandwerk: I don't think we need an extra section here

contrib/grafana-dashboard.json
```diff
@@ -807,6 +813,246 @@
...
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "description": "Latest and latest-successful Nix evaluation per tracked channel (issue #626)",
```
fricklerhandwerk:
```suggestion
      "description": "Latest and latest-successful Nixpkgs evaluation per tracked channel",
```

fricklerhandwerk (changes requested):
I'm really sorry, I toally dropped the ball on reviewing this timely after getting dragged down with fixing the Grafana setup.


Image

It took me multiple attempts to figure out what the table is trying to convey. Part of that are some general issues with readability that we should fix before merging:
- The channel column label seems broken. Also it probably should be the *first* column since we're reading the table by channel.
- Can we align the timestamps to the right or make them display in a monospace font? Otherwise they're hard to parse.
- The `elapsed` field is only set after the evaluation ended, so most of the time it will be 0. It would make sense to compute that value, or if that's too involved, only show it if the evaluation is in one of the end states.
- Please show the latest evaluation's status too, that would make a lot easier to judge what's going on
- The "Latest at HEAD" column should be together with all the other columns about the latest evaluation. But do we even need it? Isn't the latest evaluation on a channel's HEAD commit by definition?

When we have that, it's a good start we can commit to.

The other part is more general and more about possible follow-ups: what are we supposed to do with that data? For instance, it would be worth knowing how old the latest complete evaluation is for any given channel, because this is something we really need to optimize for. Right now you need to compute that in your head because the timestamps are absolute. How about displaying that as a relative age and sorting the table by youngest? Another thing worth knowing would be how long queued evaluations have been waiting, since that tells us something about whether the queue may keep growing because we're not catching up. And having a chart of evaluation times (I guess grouped by status) would also be valuable.  A chart that tells us the delay between channel update and complete evaluation of the commit would be the interesting metric we could be improving on.  All those may be very different tables altogether, I don't know.

DarshanCode2005:
image

Hi @fricklerhandwerk I have done some changes and improvements. Let me know your thoughts, after that we can work on the follow-ups namely:
1> Relative age + sorting table
2> pghistory for to queue wait column/metrics
3> Charts and alerts....

infra/sql-exporter-queries.nix
```diff
@@ -49,4 +49,61 @@
...
+            latest_successful AS (
+              SELECT DISTINCT ON (channel_id)
+                channel_id, commit_sha1, updated_at
+              FROM shared_nixevaluation
+              WHERE state = 'COMPLETED'
+              ORDER BY channel_id, updated_at DESC
+            )
+            SELECT
+              c.channel_branch::text AS channel,
+              CASE l.state
```
fricklerhandwerk, line 80: This seems overly complicated. Why not pass through the string directly?

DarshanCode2005: Prometheus metric values must be floats. Since `latest_state` is exported as a value with timestamps, here encoding the enum numerically and mapping to Grafana. Using string would work but would change the series model and table transformations.

fricklerhandwerk (approved):

PR #1130, opened 2026-06-17
feat: update pending and draft suggestion for new evaluation

### LINKED ISSUE #618: Update pending and draft suggestions with new evaluations
Context: Currently package versions in suggestions are frozen to the point in time when they're created.

As a user of the security tracker, I want to see the latest state of a package in the suggestion view, in order to make informed triaging decisions.

## Acceptance criteria

Given an auto-created suggestion in untriaged or draft state,
When a new Nixpkgs commit is evaluated,
Then the package reference in that suggestion is updated to the latest evaluation.

## Implementation notes 

Once channel are updated, we need to:
- expire all pending suggestions and redo them
- iterate over rejected suggestions and accepted suggestions
	- do as if we were going to rematch things up
	- compare what is newly matched with what exist already
	- "replace" the M2M row for everything we can link (old derivation, new derivation)
	- add the new M2M rows
	- delete the old M2M we cannot match up
_warning_ : this means we need to develop a way to relate derivation in channel rev X and derivation in channel rev Y. (https://github.com/Nix-Security-WG/nix-security-tracker/issues/617#issue-3239673918)

(this problem also exist in some of the optimization work we do in Lix/Hydra for build analytics, I think what we can do is to track by `attrPath` and that's… probably the best we can do? i.e. if someone moves the derivation to another attribute path, bye bye, we lost it.) 

 _Originally posted by @RaitoBezarius in https://github.com/Nix-Security-WG/nix-security-tracker/issues/307#issuecomment-2489282984

adekoder:
Closes #618 

Description
When a channel evaluation completes, we update each suggestion's linked derivations
to point to the new evaluation's derivations, matched by attribute path. The
suggestion's cache is then marked stale so it rebuilds on next view, showing the
latest package versions.

**The net effect**: 
- every time a new commit is evaluated for a channel, suggestions in that channel automatically track the latest package versions. 
- A user opening a pending suggestion will always see the current version of each package and its
affected/unaffected status against the CVE version constraints.

### DIFF
--- a/src/shared/listeners/automatic_linkage.py
+++ b/src/shared/listeners/automatic_linkage.py
@@ -12,7 +12,7 @@ from dataclasses import dataclass, field
 
 import pgpubsub
 from django.conf import settings
-from django.db import models
+from django.db import models, transaction
 from django.db.models import (
     Case,
     Exists,
@@ -269,6 +269,73 @@ def build_new_links(container: Container) -> bool:
         )
 
     return True
+
+
+def refresh_suggestion_derivation_links(
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    """
+    Re-match a suggestion against the latest evaluations and reconcile its
+    derivation links, mirroring all rejection/match conditions from build_new_links.
+    """
+    container = (
+        suggestion.cve.container.filter(
+            affected__package_name__isnull=False,
+        ).first()
+        or suggestion.cve.container.first()
+    )
+
+    if container is None:
+        return
+
+    outcome = resolve_linkage_candidates(container)
+
+    with transaction.atomic():
+        try:
+            suggestion = CVEDerivationClusterProposal.objects.select_for_update().get(
+                pk=suggestion.pk,
+                status__in=[
+                    CVEDerivationClusterProposal.Status.PENDING,
+                    CVEDerivationClusterProposal.Status.ACCEPTED,
+                ],
+            )
+        except CVEDerivationClusterProposal.DoesNotExist:
+            return
+
+        update_fields = ["algorithm_version"]
+        suggestion.algorithm_version = (
+            CVEDerivationClusterProposal.CURRENT_ALGORITHM_VERSION
+        )
+
+        if outcome.rejection is not None:
+            suggestion.status = CVEDerivationClusterProposal.Status.REJECTED
+            suggestion.rejection_reason = outcome.rejection.reason
+            suggestion.rejection_match_count = outcome.rejection.match_count or None
+            suggestion.rejection_max_matches_limit = (
+                outcome.rejection.max_matches_limit or None
+            )
+            update_fields += [
+                "status",
+                "rejection_reason",
+                "rejection_match_count",
+                "rejection_max_matches_limit",
+            ]
+
+        suggestion.save(update_fields=update_fields)
+
+        DerivationClusterProposalLink.objects.filter(proposal=suggestion).delete()
+
+        if outcome.derivations:
+            DerivationClusterProposalLink.objects.bulk_create(
+                build_derivation_links(suggestion, outcome.derivations)
+            )
+
+    logger.info(
+        "Refreshed derivation links for suggestion %d (rejection_reason=%s, derivations=%s).",
+        suggestion.pk,
+        outcome.rejection.reason if outcome.rejection else None,
+        outcome.derivations.count() if outcome.derivations is not None else 0,
+    )
 
 
 @pgpubsub.post_insert_listener(ContainerChannel)
--- a/src/shared/listeners/package_clustering.py
+++ b/src/shared/listeners/package_clustering.py
@@ -2,8 +2,11 @@ import logging
 
 import pgpubsub
 
+from shared.cache_suggestions import cache_new_suggestions
 from shared.channels import NixEvaluationUpdateChannel
+from shared.listeners.automatic_linkage import refresh_suggestion_derivation_links
 from shared.models import NixDerivation, NixEvaluation
+from shared.models.linkage import CVEDerivationClusterProposal
 from shared.package_clustering import cluster_packages
 
 logger = logging.getLogger(__name__)
@@ -26,3 +29,24 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
         f"updated {result.packages_updated}, created {result.packages_created} packages, "
         f"updated {result.attrpaths_updated}, created {result.attrpaths_created} attrpaths."
     )
+
+    affected_suggestions = CVEDerivationClusterProposal.objects.filter(
+        derivations__parent_evaluation__channel=evaluation.channel,
+        status__in=[
+            CVEDerivationClusterProposal.Status.PENDING,
+            CVEDerivationClusterProposal.Status.ACCEPTED,
+        ],
+    ).distinct()
+
+    count = 0
+    for suggestion in affected_suggestions:
+        refresh_suggestion_derivation_links(suggestion)
+        cache_new_suggestions(suggestion)
+        count += 1
+
+    if count:
+        logger.info(
+            "Updated derivation links and rebuilt caches for %d suggestion(s) after evaluation %s",
+            count,
+            evaluation,
+        )
--- a/src/shared/tests/test_linkage.py
+++ b/src/shared/tests/test_linkage.py
@@ -5,7 +5,10 @@ import pytest
 from django.test import override_settings
 
 from shared.cache_suggestions import cache_new_suggestions
-from shared.listeners.automatic_linkage import build_new_links
+from shared.listeners.automatic_linkage import (
+    build_new_links,
+    refresh_suggestion_derivation_links,
+)
 from shared.models.cve import Container, Tag
 from shared.models.linkage import (
     CVEDerivationClusterProposal,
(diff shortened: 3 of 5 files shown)


src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +29,33 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+
+
+def _update_suggestions_after_clustering(evaluation: NixEvaluation) -> None:
+    affected_suggestions = (
+        CVEDerivationClusterProposal.objects.filter(
+            derivations__parent_evaluation__channel=evaluation.channel,
+        )
+        .exclude(
+            status=CVEDerivationClusterProposal.Status.PUBLISHED,
+        )
```
fricklerhandwerk:
```suggestion
        .filter(
            status=CVEDerivationClusterProposal.Status.PENDING,
            status=CVEDerivationClusterProposal.Status.ACCEPTED,
        )
```

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +29,33 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+        .distinct()
+    )
+
+    count = 0
+    for suggestion in affected_suggestions:
+        update_derivation_links_for_channel(suggestion, evaluation)
+        count += 1
+
+    if count:
+        # Mark caches stale, rebuilt lazily on next view via ensure_fresh_cache()
```
fricklerhandwerk: Style nit: I prefer not to use/repeat code symbols in comments, because that's just another opportunity to go stale on refactorings. Also the value of comments to me is getting a rough idea of what's going on, expressed in more-or-less real-world concepts rather than code specifics. Keeping this distance between code and comments makes both more robust and more useful, in my experience. 

```suggestion
        # Mark caches stale, rebuilt lazily on next view
```

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +29,33 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+        )
+        .distinct()
+    )
+
+    count = 0
+    for suggestion in affected_suggestions:
+        update_derivation_links_for_channel(suggestion, evaluation)
+        count += 1
+
+    if count:
```
fricklerhandwerk, line 47: The counting seems unnecessary. If `affected_suggestions` is empty, the update query will do nothing anyway.

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +29,33 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+        # Mark caches stale, rebuilt lazily on next view via ensure_fresh_cache()
+        CachedSuggestions.objects.filter(proposal__in=affected_suggestions).update(
```
fricklerhandwerk: Since this is running out of band, how about we just update the cache in-place? A good preliminary refactoring would be adding `cache_new_suggestions` as a method (could be a pass-through to the standalone procedure for starters), and then we'd just iterate over them each one separately for now.

Then users won't incur a random-seeming response delay, and there won't be a time interval where the pending suggestion list shows empty. (Fixing that UI symptom requires a different line of work that's currently out of scope.)

src/shared/tests/test_linkage_update.py
```diff
@@ -0,0 +1,123 @@
...
+    old_eval = make_evaluation()
+    new_eval = make_evaluation()
+
+    old_drv = make_drv(evaluation=old_eval)
+    new_drv = make_drv(evaluation=new_eval, attribute=old_drv.attribute)
+
+    suggestion = make_suggestion(
+        container=cve, drvs={old_drv: ProvenanceFlags.PACKAGE_NAME_MATCH}
+    )
+
```
fricklerhandwerk: Maybe for extra paranoia, check that the number of links is also 1 before you run the update, and that what's linked is `old_drv`?

src/shared/tests/test_linkage_update.py
```diff
@@ -0,0 +1,123 @@
...
+    unstable_ch = make_channel(channel_branch="nixos-unstable")
+    unrelated_eval = make_evaluation(channel=unstable_ch)
+    update_derivation_links_for_channel(suggestion, unrelated_eval)
+
+    links = DerivationClusterProposalLink.objects.filter(proposal=suggestion)
+    assert links.count() == 1
+    assert links.get().derivation == drv
+
+
+def test_cache_marked_stale_after_clustering(
```
fricklerhandwerk: Maybe instead let's do a playwright test that updates the drv version, and that's what shows after the update. Then the implementation detail of how the cache is updated doesn't matter.

src/shared/tests/test_linkage_update.py
```diff
@@ -0,0 +1,123 @@
...
+    cache_new_suggestions(suggestion)
+
+    assert not CachedSuggestions.objects.get(proposal=suggestion).data_stale
+
+    _update_suggestions_after_clustering(new_eval)
+
+    assert CachedSuggestions.objects.get(proposal=suggestion).data_stale
+
+
+def test_published_suggestions_not_updated(
```
fricklerhandwerk: WDYT about adding a list of suggestion states eligible for updates, like we do for `TRACKED_CHANNEL_STATES`? Then the test could be similar as for that, where we pick some state, assert it's not in that list, and then do the postcondition check that nothing happened.

src/shared/linkage_update.py
```diff
@@ -0,0 +1,64 @@
...
+    suggestion: CVEDerivationClusterProposal,
+    new_evaluation: NixEvaluation,
+) -> None:
+    """
+    Replace derivation links for a suggestion with the corresponding derivations
+    from new_evaluation, matched by attribute path.
+
+    Links whose attribute path no longer exists in the new evaluation are removed.
+    """
+    old_links = list(
```
fricklerhandwerk: I suspect we can keep `old_links` a queryset, and later just `old_links.delete()`. See further comments. But I may be wrong here...

src/shared/linkage_update.py
```diff
@@ -0,0 +1,64 @@
...
+            attribute__in=attr_to_link.keys(),
+            parent_evaluation=new_evaluation,
+        )
+    }
+
+    to_delete_pks = [link.pk for link in old_links]
+    to_create = [
+        DerivationClusterProposalLink(
+            proposal=suggestion,
+            derivation=new_drvs[attr],
```
fricklerhandwerk: I don't think this works, because many drvs could have the same attribute, and you'd get them essentially in random order in that dict.

src/shared/linkage_update.py
```diff
@@ -0,0 +1,64 @@
...
+
+    attr_to_link = {link.derivation.attribute: link for link in old_links}
+
+    new_drvs = {
+        drv.attribute: drv
+        for drv in NixDerivation.objects.filter(
+            attribute__in=attr_to_link.keys(),
+            parent_evaluation=new_evaluation,
+        )
+    }
```
fricklerhandwerk:
```suggestion
    new_drvs = NixDerivation.objects.filter(
        attribute__in=old_links.value_list("derivation__attribute", flat=True).distinct(),
        parent_evaluation=new_evaluation,
    )
```

But then we can't just bulk-relink them, because we somehow need to set the right the provenance flags...

Also in light of the plan for making matching stricter (e.g. filtering out versions), we should make a bit better provisions for new derivations not getting matched to the CVE any more. How about we essentially rematch `new_drvs` against the given suggestion's CVE using `produce_linkage_candidates()`?

Then drop the old links and add the new ones as we do in `build_new_links`. We may need to take another look how to do that cleanly, and also think about stuff like:
- what if all new derivations fail the matching? do we auto-reject the CVE?
- what if we hit the `knownVulnerabilities` case? we'd need to extract that so it's handled in both code paths

fricklerhandwerk (commented):
Good stuff, overall! I think we can get around adding a field to the cache model by doing an eager update. I'd really like to have the cache layer die off slowly instead of expanding it with new bells and whistles.

I also gave potential issues with concurrent updates via our `Ignore` buttons a brief thought. Seems like we're fine, since we're currently handling packages by attribute. If a maintainer disappears from one version to another, an ignore/restore request just fails and the next page load will show the correct thing. @florentc any other worries regarding that?

The relinking part needs more thought though. Right now we'd be dropping derivations somewhat randomly. A somewhat robust approach would be actually rematching them using the same code as for the original matching, and handling the fallout. A really robust approach would be matching and re-matching via packages instead, but that's probably way too much for now.

I'll set it to draft until we figure out all the corners of rematching.

src/shared/listeners/package_clustering.py
fricklerhandwerk: Ah and what happens when we crash mid-way? Do we just wait for the next evaluation to hit?

src/shared/linkage_update.py
```diff
@@ -0,0 +1,67 @@
...
+)
+
+logger = logging.getLogger(__name__)
+
+
+def refresh_suggestion_derivation_links(
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    """
+    Re-evaluate a suggestion against the latest evaluations and reconcile its
```
fricklerhandwerk:
```suggestion
    Re-match a suggestion against the latest evaluations and reconcile its
```

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +29,28 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
         f"updated {result.packages_updated}, created {result.packages_created} packages, "
         f"updated {result.attrpaths_updated}, created {result.attrpaths_created} attrpaths."
     )
+
+    _update_suggestions_after_clustering(evaluation)
```
fricklerhandwerk: I don't think we should extract that. This was super confusing to read. What's wrong with calling `cluster_after_evaluation()` after `make_evaluation()` in the test directly? Or even folding the call into `make_evaluation()` directly?

src/shared/models/linkage.py
```diff
@@ -79,6 +79,10 @@ class RejectionReason(models.TextChoices):
             "known_vulnerability",
             _("already tracked in derivation metadata"),
         )
+        NO_MATCHES = (
+            "no_matches",
+            _("no matching derivations found"),
```
```suggestion
            _("no matching packages found"),
```

We can already anticipate that we'll be matching packages Soon™

src/shared/tests/test_linkage.py
```diff
@@ -133,7 +139,13 @@ def test_link_product_or_package_name(
         # Check the whole data pipeline by also caching the suggestion
         cache_new_suggestions(link.proposal)
     else:
-        assert not match
+        assert match
```
fricklerhandwerk: Better assert the match after it's made in L134.

src/shared/listeners/automatic_linkage.py
```diff
@@ -102,34 +200,7 @@ def produce_linkage_candidates(
     return matches
 
 
-def build_new_links(container: Container) -> bool:
```
fricklerhandwerk, line 228: I have a really hard time following this refactoring. I think I see why you made it: we need to extract the "matching outcome" to be independent from creating a new suggestion. But right now it's almost impossible to see whether or what behavior changed. From the tests we can vaguely infer that it's pure, but I find it safer to have a separate commit with a smaller diff as far as that's possible. For instance, I don't think it's strictly required to extract `get_filtered_affected`.

src/shared/linkage_update.py
fricklerhandwerk, line 1: I think this can go into `automatic_linkage.py`, it's about the same business

src/shared/listeners/automatic_linkage.py
```diff
@@ -24,14 +25,111 @@
...
+    - rejection_reason    set when the container warrants a rejected proposal
+    - derivations=None    no links should be attached (rejection without derivations)
+    - derivations=non-empty  links to attach
+    - rejection_match_count / rejection_max_matches_limit  populated for MAX_MATCHES_EXCEEDED
+    """
+
+    rejection_reason: CVEDerivationClusterProposal.RejectionReason | None = None
+    derivations: models.QuerySet | None = field(default=None)
+    rejection_match_count: int = 0
+    rejection_max_matches_limit: int = 0
```
fricklerhandwerk: I think those are quite evident
```suggestion
    derivations: models.QuerySet | None = field(default=None)
    rejection_reason: CVEDerivationClusterProposal.RejectionReason | None = None
    # Populated when when match limit exceeded
    rejection_match_count: int = 0
    rejection_max_matches_limit: int = 0
```

Alternatively, for clarity create an extra data type just for the rejection case:

```python
    @dataclass
    class Rejection:
        reason = CVEDerivationClusterProposal.RejectionReason
        # Populated when when match limit exceeded
        match_count: int = 0 
        max_matches_limit: int = 0
```

... and then

```python
    derivations: models.QuerySet | None = field(default=None)
    rejection = Rejection | None = None
```

Since we're talking about it, @DarshanCode2005 we may want another reason variant for the weird case we encountered where a comment is supplied and which currently keeps the reason as `None`. Then we can confidently say the `reason` field is always set if the suggestion is to be rejected.

src/shared/listeners/automatic_linkage.py
fricklerhandwerk: Would it be a viable alternative to return the suggestion as an object and then update the right fields in the database entry through it? Then we wouldn't need an intermediate data type.

fricklerhandwerk (commented):
Cool, some good catches there. Primarily I think it needs some more clarity because it's such a load-bearing change, so still style discussion to be done.

src/shared/listeners/package_clustering.py
adekoder: Not sure i understand this comment on this line?

src/shared/listeners/package_clustering.py
adekoder: this is for log purposes not for performing the updates.

src/shared/listeners/package_clustering.py
fricklerhandwerk: Ah it seems you removed the call to `_update_suggestions_after_clustering()` from the tests already. Then we can just skip this and the following `def _update...` line. I'll push a commit.

adekoder: not sure i understand... you want to push a commit for ?

fricklerhandwerk: See https://github.com/NixOS/nix-security-tracker/pull/1130/commits/ed9a2571d59f8abee377fb2b4b4cb9e5e6cc587b

fricklerhandwerk:
I just realized we may get into trouble with concurrency again: Imagine we're running a post-eval re-matching while a user changes the state of a suggestion, e.g. to REJECTED or PUBLISHED.

adekoder:
> I just realized we may get into trouble with concurrency again: Imagine we're running a post-eval re-matching while a user changes the state of a suggestion, e.g. to REJECTED or PUBLISHED.

Ah true, Good Catch, we need to lock the row in post eval

fricklerhandwerk:
Pushing a fix.

fricklerhandwerk (approved):
I think this is good now. Do you want to squash and merge when tests are green?

PR #1132, opened 2026-06-17
feat: Boilerplate of new Preact-powered frontend

### LINKED ISSUE #1119: Dynamic UI prototype
Set up the infrastructure for implementing UI views with a stateful client.

florentc:
Closes https://github.com/NixOS/nix-security-tracker/issues/1119

Lays down the boilerplate for a new SPA web UI (Preact + Vite + TypeScript) that coexists with the existing Django views.
It's served on a dedicated /ui-v2/ route, authenticates via the existing session cookies, and talks to the REST API
through a type-safe, auto-generated client.

This PR is boilerplate + the migration of a proof of concept toy feature (toggling auto-subscribe). Building out the rest of
the UI and broader API improvements will be done is subsequent PRs.

## Frontend app (frontend/)

 - Preact + Vite + TypeScript project
 - Biome for linting + formatting
 - Orval generates a TanStack-Query API client from the drf-spectacular OpenAPI schema (/api/schema). Generated code is put in src/api/generated/ (git-ignored).
 - Server-injected runtime config via window.__CONFIG__ (debug, showDemoDisclaimer, production, revision) with dev 
fallbacks.

## Django integration

 - django-vite: Django serves the HTML shell (ui_v2.html); Vite serves JS/CSS assets. VITE_DEV_MODE switches between the Vite dev server and the built assets (minified js and css).
 - Catch-all route on /ui-v2/*, enabling client-side routing. /api and /accounts 
continue to hit Django (specified in vite config)
 - New GET /api/v1/me endpoint returning the current user (username, avatar, admin/committer flags) for cookie-based auth in the UI.
 - API cleanup to produce a cleaner OpenAPI schema (at least around the toy feature this PR handles).

## Nix & tooling

 - nix/frontend.nix builds the frontend as a derivation
 - Biome pre-commit hook runs from frontend/ so its config is treated as the project root.
 - Playwright-based test suite for the new UI under src/webview/tests/ui_v2/ (app-loads + auto-subscribe toggle), with 
centralized route constants and an auto-detecting build fixture. See src/webview/tests/ui_v2/README.md.

## How to run locally

     manage runserver
     cd frontend && npm run dev

Then go to http://127.0.0.1:8000/ui-v2/


frontend/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+Start Django and the Vite dev server:
+
+```console
+# Terminal 1: Django (from repo root, inside nix-shell)
+manage runserver
+
+# Terminal 2: Vite asset server (from frontend/)
+cd frontend
+npm run dev
```
fricklerhandwerk: Yeah I wonder if we should just reintroduce the Procfile instead...

frontend/README.md
```diff
@@ -0,0 +1,189 @@
...
+- **Always fresh**: nix-shell entry ensures dependencies are installed; you regenerate on demand
+- **No stale code risk**: if you forget to regenerate, TypeScript will catch mismatches at compile time
+
+### First time setup / after pulling
+
+If `src/api/generated/` is empty and you need it:
+
+```console
+# Ensure Django is running, then:
+npm run generate-api
```
fricklerhandwerk: Sigh, every time I see this sort of manual dependency chain I get more convinced it's time for a native service management in shells. (Has nothing to do with how you did this, just a rant on the side.)

Can't we hack this together to be slightly more ergonomic? When I try this at my keyboard, I bet half of it will fail because I forgot steps. Maybe some of them can be automated?

frontend/README.md
```diff
@@ -0,0 +1,189 @@
...
+
+## Linting and formatting
+
+This project uses [Biome](https://biomejs.dev/) for both linting and formatting.
+
+> **Important**: The `biome` binary comes from `pkgs.biome` in the Nix shell, NOT from npm.
+> This avoids FHS/glibc compatibility issues that Nix environments have with native npm binaries.
+> Do **not** add `@biomejs/biome` to `package.json`.
+
+### Commands
```
fricklerhandwerk: Do we really need this explicitly in the shell? It's run as part of `format` already.

frontend/biome.json
```diff
@@ -0,0 +1,29 @@
+{
```
fricklerhandwerk, line 1: This should probably go into pre-commit.nix

frontend/src/components/footer/Footer.module.css
fricklerhandwerk, line 1: Why can't we reuse the base template and send this static information statically?

Also we'd be duplicating it for an unknown amount of time...

frontend/src/components/header/AuthStatus.tsx
```diff
@@ -0,0 +1,38 @@
+import { LogInIcon, LogOutIcon, UserIcon } from "lucide-preact";
+import { Link } from "wouter-preact";
+import { LOGIN_URL, logout, useAuth } from "@/hooks/useAuth";
+import styles from "./AuthStatus.module.css";
+
+export function AuthStatus() {
+  const { user, isAuthenticated, isLoading } = useAuth();
+
+  if (isLoading) {
+    return <span>Loading...</span>;
```
fricklerhandwerk, line 10: I noticed that with short latencies the loading state flashes very briefly. Is there a way to configure the library to only show it when it's been actually toggled long enough?

frontend/src/config.ts
```diff
@@ -0,0 +1,47 @@
+/**
+ * Server-injected application configuration.
+ *
+ * In production, Django injects a `window.__CONFIG__` object in the HTML template
```
fricklerhandwerk: Django or Django-vite?

And do we really need the boilerplate duplication of defaults? Seems very unhealthy... can't we make it such that if everything is wired the values will be there, otherwise it's clearly broken and fails tests?

frontend/vite.config.ts
```diff
@@ -0,0 +1,28 @@
...
+  build: {
+    manifest: true,
+    rollupOptions: {
+      input: "src/main.tsx",
+    },
+  },
+  server: {
+    port: 5173,
+    // Absolute origin so HMR asset URLs work when Django serves the HTML from :8000
+    origin: "http://localhost:5173",
```
fricklerhandwerk, line 26: All this is development only? The hard coding hurts my brain a bit.

nix/frontend.nix
```diff
@@ -0,0 +1,20 @@
+{ pkgs }:
```
fricklerhandwerk: Expand all direct children of pkgs here, this would be idiomatic

src/api/issues/views.py
```diff
@@ -8,6 +8,13 @@
 
 from shared.models import NixpkgsIssue
 
+# FIXME(@florentc): The inline Serializer class causes drf-spectacular to generate a schema
```
fricklerhandwerk, line 11: We should merge those fixes as preparatory work for this PR

src/project/settings.py
```diff
@@ -469,6 +470,9 @@ class AppSettings(BaseModel):
 
 # Disable regular signup but allow GitHub auth
 SOCIALACCOUNT_ONLY = True
+SOCIALACCOUNT_LOGIN_ON_GET = (
+    True  # Skip intermediate "Continue" page, redirect to GitHub immediately
+)
```
fricklerhandwerk: Why not...?

```suggestion
# Skip intermediate "Continue" page, redirect to GitHub immediately
SOCIALACCOUNT_LOGIN_ON_GET = True
```

src/project/settings.py
```diff
@@ -492,6 +496,54 @@ class AppSettings(BaseModel):
 
 STATIC_URL = "static/"
 
+# django-vite: Vite serves JS/CSS assets, Django serves the HTML shell.
```
fricklerhandwerk: I need to look at this more carefully. Some of that complexity is absolutely warranted, but my gut feel is this will be extremely brittle and it should be figured out somewhat properly.

fricklerhandwerk: Generally, I suppose some of this can be moved into the fancy pydantic-based config parser above.

src/webview/templates/ui_v2.html
```diff
@@ -0,0 +1,28 @@
...
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Nixpkgs security tracker</title>
+    {% vite_hmr_client %}
+    {% vite_asset 'src/main.tsx' %}
+  </head>
+  <body>
+    <!--
```
fricklerhandwerk, line 12: Clever!

src/webview/tests/ui_v2/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+        )
+
+    # <dist>/.vite/manifest.json -> <dist>
+    assets_dir = manifest_path.parent.parent
+
+    settings.DJANGO_VITE = {
+        "default": {
+            "dev_mode": False,
+            "dev_server_host": "localhost",
+            "dev_server_port": 5173,
```
fricklerhandwerk: Would be great to have a single source of truth for this

src/webview/tests/ui_v2/conftest.py
```diff
@@ -0,0 +1,124 @@
...
+    server-rendered pages. The new UI is a client-rendered SPA: with JavaScript disabled
+    it renders nothing, so that parametrization is meaningless here. Overriding
+    ``browser_context_args`` (without referencing ``no_js``) drops the parametrization for
+    everything under this package.
+    """
+    return {"java_script_enabled": True}
+
+
+@pytest.fixture(scope="session", autouse=True)
+def _frontend_assets() -> Generator[None]:
```
fricklerhandwerk, line 30: This is extremely clever, in a rarely occurring good way. I liked that idea a lot when you proposed it. Impressive that it works and is not a pile of impenetrable garble.

This is by far the best compromise between static builds and live reloads I've ever seen.

fricklerhandwerk (commented):
Didn't look at it too closely for the obvious reasons, but design wise this is very impressive.

I guess the next step is cutting down on the LLM noise, in particular the mostly redundant documentation. Some bits here and there are valuable, in particular the rationale. But mostly it just describes what the code does way too literally, that can go away. What I would like for documentation is linking to authoritative sources on the pile of new tools. Just the sheer number of names makes my head spin.


The other thing would be tying it together more firmly, so we're not sprinkling hard coded values over multiple places.

And otherwise generally reducing the diff to what's needed for now (In particular I strongly suggest removing the base template duplication. We can still figure out how to do it better later.)

frontend/src/components/header/AuthStatus.tsx
florentc: - if we don't show "Loading...", it means we show nothing until the data is ready, and when it's ready the component renders and we'll see it flash anyway
- the "flash", a.k.a. fetching the user info from the API, is only done on page reload when you arrive, press F5, or open a new tab. If you navigate inside the web app, this is in cache, not invalidated, and not re-rendered. Go from user settings to homepage by clicking on the avatar / main title and you'll see there is no reload nor flash.


This behavior will be present in many places in the UI in the context of dynamic loading of pieces of data. I think it's noticeable here because there is no pretty loading state yet, it's a raw "Loading..." text.

I propose to use "skeletons" during loading states: placeholders that occupy roughly the same space and shape as the loaded component. That kind of thing:



I have not yet chosen an implementation yet, hence the "Loading..." here and there, but it was on my todo list: deciding on a consistent way to represent loading state.

Here, that would mean a circle the same size as the loaded avatar with a subtle loading animation. This would reduce the visual noise.

frontend/src/config.ts
florentc: This is just to display the staging banner and the revision in the footer. The values are fed by Django (not Django-vite) through the ui-v2 empty-shell template. It is wired and the values will be there.

The alternative would be to have an API endpoint from which to obtain the information. Maybe it's better in fact. That way, it's available to API users as well, and everything that feeds UI v2 comes consistently from API endpoints.

frontend/vite.config.ts
florentc: It's development only. Vite runs on 5173 by default. The only time you'll run Vite is during development for live reload and debugging. In production and tests, it uses built assets.

src/api/issues/views.py
florentc: I decided to remove them from the scope of this PR because the issue endpoints will have to be reworked anyway later. They are already no longer suitable for the multi-CVE per issue feature. I wanted to just care about the API we do use in this first iteration of UI v2 and avoid fixing the issue API in this PR.

frontend/src/components/footer/Footer.module.css
florentc: If I'm not mistaken you'd want to render the header and footer with Django, and then inject the dynamic Preact web app in the middle.

I decided against this for several reasons:
- This prevents us from having any dynamism and benefits from the preact UI in the header and footer. For instance rendering the user info, implementing fancier menus in the future etc.
- This will be more convenient to maintain. Everything is rendered by the preact app and that's it. All the styling is done there too in consistent way.
- It brings nothing.

I think the way to go would be to have API endpoints serving the information in question. That way all is consistent. All the info comes from API, and frontend renders everything.

See https://github.com/NixOS/nix-security-tracker/pull/1132#discussion_r3435968092 it's linked

frontend/src/api/.gitignore
```diff
@@ -0,0 +1,3 @@
+# Generated by Orval — do not edit manually.
+# Regenerate with: npm run generate-api (requires Django running on :8000)
```
fricklerhandwerk:
```suggestion
```

Reads like `.gitignore` is generated manually. And the regeneration instruction is redundant with some other doc.

frontend/src/api/client.ts
```diff
@@ -0,0 +1,109 @@
...
+ * 1. Serializes query params (Orval passes them as an object, not a URL string)
+ * 2. Injects X-CSRFToken header on unsafe methods (Django CSRF requirement)
+ * 3. Serializes JSON request bodies
+ * 4. Parses JSON responses and returns typed data (Promise<T>)
+ * 5. Throws ApiError on non-2xx (so TanStack Query marks the query as failed)
+ *
+ * The ApiError class exposes `.status` so consumers can discriminate errors
+ * (e.g., useAuth treats 401 as "not logged in" rather than a failure).
+ *
+ * REFERENCED BY: orval.config.ts → output.override.mutator
```
fricklerhandwerk: Looks like this could drift?
```suggestion
```

CONTRIBUTING.md
```diff
@@ -47,6 +47,19 @@ List all available [management commands](https://docs.djangoproject.com/en/6.0/r
 manage help
 ```
 
+### Start the service
```
fricklerhandwerk: This needs to go to the new quickstart guide

frontend/src/config.ts
```diff
@@ -0,0 +1,32 @@
+/**
+ * Server-injected application configuration.
```
fricklerhandwerk, line 2:
```suggestion
 * Server-injected application configuration.
 * FIXME(@florentc): Expose this through an API endpoint so we don't have to hard-code it.
```

frontend/README.md
```diff
@@ -0,0 +1,28 @@
+# Frontend ([Preact](https://preactjs.com/) / [Vite](https://vite.dev/) UI)
+
+The new frontend for the Nixpkgs security tracker, served at `/ui-v2/`. It
```
fricklerhandwerk: Please reformat as one line per sentence

nix/schema.nix
```diff
@@ -0,0 +1,61 @@
+# Generates the OpenAPI schema (schema.yaml) from the Django REST API offline,
```
fricklerhandwerk:
```suggestion
# Generates the OpenAPI schema from the Django REST API offline,
```

I'll eventually add this to the contributing guide, but I think it's a very bad idea to mention literal code tokens in in-code documentation. I strongly prefer talking about the code in general terms (and even more preferrably: only for explanation, never description), from experience this is the best handle against drift.

frontend/README.md
```diff
@@ -0,0 +1,28 @@
...
+## API client generation ([Orval](https://orval.dev/))
+
+The [TanStack Query](https://tanstack.com/query) client in `src/api/generated/`
+is generated (and gitignored) from the OpenAPI schema using Orval. The schema is the
+single source of truth. Regenerate it when the API changes.
+
+From the live schema (with Django running) run `npm run generate-api`.
+
+For offline/CI generation from a saved schema file, use
+`npm run generate-api:local` (reads `./schema.yaml`); the Nix frontend build does
```
fricklerhandwerk:
```suggestion
`npm run generate-api:local` (reads a schema generated with [drf-spectacular](https://github.com/tfranzel/drf-spectacular)); the Nix frontend build does
```

fricklerhandwerk: In fact we could simply expose that built file if somebody needs it (why would you do something *without* Nix?). So let's just remove the paragraph for now.

frontend/README.md
```diff
@@ -0,0 +1,28 @@
...
+this automatically.
+
+`orval-transformer.ts` patches the schema before generation. Ideally, this
```
fricklerhandwerk:
```suggestion
The schema is [fixed up before it's generated](./orval-transformer.ts). Ideally, this
```

src/project/settings.py
```diff
@@ -205,6 +205,21 @@ def default_server_email(self) -> Self:
                 self.SERVER_EMAIL = self.DEFAULT_FROM_EMAIL
             return self
 
+        VITE_MANIFEST_PATH: Path | None = Field(
+            description="""
+            Built-asset manifest. In production this is set to the Nix-built path via the
+            DJANGO_VITE_MANIFEST_PATH env var; locally it defaults to the dev build output.
```
fricklerhandwerk:
```suggestion
            Manifest of built assets.
            In production this is set to the Nix-built path via environment variable;
            locally it defaults to the one built with `npm` in the development environment.
```

fricklerhandwerk (commented):
Reviewed together. Just some style nits, the volume and density of comments is very good now. I'll do another short pass later we'll merge today.

src/webview/tests/ui_v2/routes.py
```diff
@@ -0,0 +1,8 @@
+"""Route definitions for UI v2 tests.
```
fricklerhandwerk:
```suggestion
"""
Route definitions for UI v2 tests.
```

I wonder whether we can encode such stuff in the linter settings...

src/webview/tests/ui_v2/routes.py
```diff
@@ -0,0 +1,8 @@
...
+Mirrors the client-side routes declared in ``frontend/src/App.tsx``
+To be kept in sync.
```
fricklerhandwerk:
```suggestion
Mirrors the client-side routes from the frontend App.
# FIXME(@fricklerhandwerk): Read the routes from a single source of truth.
```

src/webview/tests/ui_v2/conftest.py
```diff
@@ -0,0 +1,118 @@
+"""Test setup for the new Preact/Vite UI served at ``/ui-v2/``.
+
+These tests reuse the legacy Playwright fixtures from ``webview/tests/conftest.py``
+(``logged_in_as``, ``as_staff``, ``mock_oauth_login``) and add the one thing the SPA
+needs that server-rendered pages don't: its built JavaScript/CSS assets served during
+the test ``live_server`` run.
+
+See ``README.md`` in this directory for the full rationale, the three frontend-serving
+cases, and the build auto-detection truth table.
```
fricklerhandwerk:
```suggestion
"""
Test setup for the new Preact/Vite UI

These fixtures build the JavaScript/CSS assets served by the test server.
This allows running tests from the source tree without rebuilding the Nix derivation
or starting the local development server.
```

src/webview/tests/ui_v2/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+
+@pytest.fixture
+def browser_context_args() -> dict[str, Any]:
+    """Always run the SPA with JavaScript enabled.
+
+    The legacy suite parametrizes a ``no_js`` fixture to test progressive enhancement of
+    server-rendered pages. The new UI is a client-rendered SPA: with JavaScript disabled
+    it renders nothing, so that parametrization is meaningless here. Overriding
+    ``browser_context_args`` (without referencing ``no_js``) drops the parametrization for
+    everything under this package.
```
fricklerhandwerk:
```suggestion
    """
    Always run the SPA with JavaScript enabled.

    The legacy suite parametrizes fixture to test progressive enhancement of
    server-rendered pages, but that's not needed here.
```

src/webview/tests/ui_v2/conftest.py
```diff
@@ -0,0 +1,118 @@
...
+    The only environment that cannot build is exactly the one where freshness is already
+    guaranteed, so there is no stale-asset risk and no manual flag is needed.
+
+    Why reconfigure at runtime rather than via env vars: the whole ``webview`` suite shares
+    one process, so settings are imported once (with ``dev_mode`` defaulting to ``DEBUG``),
+    and the built manifest only exists *after* the build runs. So this fixture forces
+    ``dev_mode=False``, points ``manifest_path`` at the chosen build, registers the build
+    directory under the ``vite`` static prefix (so ``live_server``'s static handler serves
+    ``/static/vite/...``), and resets django-vite's cached loader plus the staticfiles
+    finder cache so the changes take effect.
```
fricklerhandwerk:
```suggestion
```

I think the top comment is enough

src/webview/tests/ui_v2/README.md
fricklerhandwerk, line 1: I think we can drop this altogether, it doesn't seem to add anything to what's already said

CONTRIBUTING.md
```diff
@@ -327,6 +340,17 @@ Interact with the involved virtual machines in a test:
 $(nix-build -A tests.driverInteractive)/bin/nixos-test-driver
 ```
 
+### Tests for the new frontend (`/ui-v2/`)
```
fricklerhandwerk: Not convinced we need this. There's nothing extra one has to do to test the new UI, so why repeat the same instruction?

fricklerhandwerk (approved):
👍 modulo remaining docslop

fricklerhandwerk:

PR #1135, opened 2026-06-18
correct backtick formatting in CONTRIBUTING.md

devnchill:
there was one extra backtick causing incorrect rendering of `head_sha1_commit`


fricklerhandwerk:
I think you need to retry and only do the desired change.

devnchill:
> I think you need to retry and only do the desired change.

yeah , i bymistake created this branch out of web-metrics branch so it ended up using the web metrics commit :sweat_smile:

devnchill:
https://github.com/NixOS/nix-security-tracker/pull/1136 this is the correct one

PR #1138, opened 2026-06-20
feat: auto-reject CVEs exceeding MAX_MATCHES limit

DarshanCode2005:
### Summary

Previously, if automatic linkage found more than MAX_MATCHES derivations, the CVE was skipped and only a warning was logged. With this change, those cases are now handled consistently with other auto-rejection scenarios.


src/shared/models/linkage.py
```diff
@@ -71,6 +71,10 @@ class RejectionReason(models.TextChoices):
             "hardware_only_cpe",
             _("hardware only"),
         )
+        MAX_MATCHES_EXCEEDED = (
+            "max_matches_exceeded",
+            _("max allowed matches exceeded"),
```
```suggestion
            _("max. allowed matches exceeded"),
```

```suggestion
            _("maximum allowed matches exceeded"),
```

src/shared/models/linkage.py
```diff
@@ -214,6 +221,16 @@ def clean(self) -> None:
                     "in_issue_draft": "A suggestion can only be in an issue draft when it is accepted"
                 }
             )
+        if self.rejection_reason == self.RejectionReason.MAX_MATCHES_EXCEEDED:
```
fricklerhandwerk: This should go into the above conditional where the status is "rejected". And since we're at it, it should be an elif block, and the else else throws so we're sure to capture incomplete case coverage as an error.

fricklerhandwerk (approved):
Very good! Just some small style notes.

src/shared/models/linkage.py
```diff
@@ -201,6 +208,40 @@ def clean(self) -> None:
...
+                    )
+            elif self.rejection_reason in (
+                self.RejectionReason.EXCLUSIVELY_HOSTED_SERVICE,
+                self.RejectionReason.NOT_IN_NIXPKGS,
+                self.RejectionReason.HARDWARE_ONLY_CPE,
+                self.RejectionReason.KNOWN_VULNERABILITY,
+            ):
+                pass
+            elif self.rejection_reason is None:
+                pass
```
fricklerhandwerk: It was in fact a bit surprising why we would have a None rejection reason here again, but yeah, we only had the None + no comment case so far.

```suggestion
                None,  # When comment is set
                self.RejectionReason.EXCLUSIVELY_HOSTED_SERVICE,
                self.RejectionReason.NOT_IN_NIXPKGS,
                self.RejectionReason.HARDWARE_ONLY_CPE,
                self.RejectionReason.KNOWN_VULNERABILITY,
            ):
                pass
```

Not sure this is the least surprising way of writing it though.

fricklerhandwerk (approved):

fricklerhandwerk:
Nice, please squash.

Issue #1002, opened 2026-04-17
Application metrics


fricklerhandwerk:
We have a lot of things that are performance-sensitive, but we're not tracking any of them:
- Cache regeneration
- Nixpkgs evaluation
- CVE ingestion
- Garbage collection
- Matching algorithm performance
- probably more

Store the metrics somewhere instead of logging them into the bitbucket, and set up Grafana dashboards to make them visible.

This issue is primarily about setting up the wiring to collect and display application-level metrics in the first place. It's not about any particular metric, so a PR should merely make samples for how to collect them and can pick a straightforward one.

DarshanCode2005:
Hi Sir @fricklerhandwerk, could you please assign this issue to me? I’m currently away attending a developer conference, but I’ll be back in a few days and can then take a look at it.

PR #1139, opened 2026-06-21
feat: expose proposal false-positive ratio

DarshanCode2005:
Towards #1002

Adds a false_positive_ratio SQL exporter metric: triaged proposal rejection rate (rejected / (accepted + rejected), with published counted as accepted). Users can see matching triage quality without running manual SQL.

### DIFF
--- a/infra/README.md
+++ b/infra/README.md
@@ -57,6 +57,6 @@ Moreover, a [Postgres Exporter](https://github.com/prometheus-community/postgres
 
 Metrics are scraped hourly from the tracker database and exposed as `sql_<name>` in Prometheus.
 
-| Metric | Definition |
-|--------|------------|
+| Metric                 | Definition                                                                                                                                                                                                                                                                                            |
+| ---------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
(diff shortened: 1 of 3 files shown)


infra/README.md
```diff
@@ -52,3 +52,11 @@ Then commit `secrets/name_of_secret.age` as usual.
 A [Prometheus Node Exporter](https://github.com/prometheus/node_exporter) is running exposing host specfic metrics. These are scraped by the nixos.org [Prometheus](https://prometheus.nixos.org/graph) and are also available under [Grafana](https://grafana.nixos.org/d/rYdddlPWk/node-exporter-full?orgId=1&from=now-24h&to=now&timezone=browser&var-datasource=default&var-job=node&var-node=tracker.security.nixos.org:9100&var-diskdevices=%5Ba-z%5D%2B%7Cnvme%5B0-9%5D%2Bn%5B0-9%5D%2B%7Cmmcblk%5B0-9%5D%2B&refresh=1m).
 
 Moreover, a [Postgres Exporter](https://github.com/prometheus-community/postgres_exporter) and a [SQL exporter](https://github.com/justwatchcom/sql_exporter) run in the host exposing Postgres and application specific metrics. A dashboard for these metrics is [here](https://grafana.nixos.org/d/beo2uotj65lvkb/nix-security-tracker?orgId=1&from=now-6h&to=now&timezone=browser&var-Instance=tracker.security.nixos.org:9237).
+
+### Application metrics
```
fricklerhandwerk, line 56: If you want to point people to the source of just the matching quality metrics, consider extracting just those into a separate file and point to that file. This is of course conditional on that these metrics are somewhat special. If they're not, it's enough to just point people to the general metrics file (and maybe rename it accordingly so it can be found more easily just by browsing the source).

fricklerhandwerk (approved):
Looks alright, I need to fix the local Grafana setup to be able to test it though.

PR #1141, opened 2026-06-22
feat: collect performance metrics for regenerating cached suggestions

DarshanCode2005:
Related to #1002

### Problem

Performance-sensitive batch jobs (cache regeneration, CVE ingestion, garbage collection etcetra) only log durations and counts to stdout/logs. Operators have no Grafana view of whether these jobs are slowing down or processing less data over time. The existing SQL exporter gives aggregate counts and recency (`sql_delta`, `sql_matching`), but not job durations or throughput.

### What this PR does

Establishes the **end-to-end wiring** for application-level metrics from batch jobs to Grafana. Per issue scope, this is wiring + **one sample metric**, not full instrumentation.

### Pipeline
regenerate_cached_suggestions -> cache_regeneration.prom -> node_exporter textfile collector -> Prometheus -> Grafana

### DIFF
--- a/contrib/grafana-dashboard.json
+++ b/contrib/grafana-dashboard.json
@@ -884,6 +884,149 @@
       ],
       "title": "Database Size",
       "type": "stat"
+    },
+    {
+      "collapsed": false,
+      "gridPos": {
+        "h": 1,
+        "w": 24,
+        "x": 0,
+        "y": 17
+      },
+      "id": 20,
+      "panels": [],
+      "title": "Application metrics",
+      "type": "row"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "description": "Duration of the last cache regeneration run (oneshot batch job)",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": 0
+              }
+            ]
+          },
+          "unit": "s"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 5,
+        "w": 6,
+        "x": 0,
+        "y": 18
+      },
+      "id": 21,
+      "options": {
+        "colorMode": "none",
+        "graphMode": "area",
+        "justifyMode": "auto",
+        "orientation": "auto",
+        "percentChangeColorMode": "standard",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "showPercentChange": false,
+        "textMode": "auto",
+        "wideLayout": true
+      },
+      "pluginVersion": "12.1.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "expr": "sectracker_cache_regeneration_duration_seconds{job=\"node\", instance=\"$Instance\"}",
+          "legendFormat": "{{mode}}",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "Cache regeneration duration",
+      "type": "stat"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${DS_PROMETHEUS}"
+      },
+      "description": "Suggestions regenerated in the last cache regeneration run",
+      "fieldConfig": {
+        "defaults": {
+          "color": {
+            "mode": "thresholds"
+          },
+          "mappings": [],
+          "thresholds": {
+            "mode": "absolute",
+            "steps": [
+              {
+                "color": "green",
+                "value": 0
+              }
+            ]
+          }
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 5,
+        "w": 6,
+        "x": 6,
+        "y": 18
+      },
+      "id": 22,
+      "options": {
+        "colorMode": "none",
+        "graphMode": "area",
+        "justifyMode": "auto",
+        "orientation": "auto",
+        "percentChangeColorMode": "standard",
+        "reduceOptions": {
+          "calcs": [
+            "lastNotNull"
+          ],
+          "fields": "",
+          "values": false
+        },
+        "showPercentChange": false,
+        "textMode": "auto",
+        "wideLayout": true
+      },
+      "pluginVersion": "12.1.1",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "expr": "sectracker_cache_regeneration_suggestions{job=\"node\", instance=\"$Instance\"}",
+          "range": true,
+          "refId": "A"
+        }
+      ],
+      "title": "Cache regeneration suggestions",
+      "type": "stat"
     }
   ],
   "schemaVersion": 41,
--- a/default.nix
+++ b/default.nix
@@ -120,6 +120,7 @@ rec {
               };
             in
             if git ? dirtyRev then "${git.shortRev}-dirty" else git.shortRev;
+          METRICS_TEXTFILE_DIR = toString ./. + "/prometheus-metrics";
         };
       };
 
@@ -164,6 +165,7 @@ rec {
         ln -sf ${sources.nixos-logo} src/webview/static/nixos-logo.svg
 
         mkdir -p $CREDENTIALS_DIRECTORY
+        mkdir -p ${toString ./.}/prometheus-metrics
         # TODO(@fricklerhandwerk): move all configuration over to pydantic-settings
         touch .settings.py
         export USER_SETTINGS_FILE=${builtins.toString ./.settings.py}
--- a/infra/common.nix
+++ b/infra/common.nix
@@ -88,7 +88,17 @@ in
   services.prometheus.exporters.node = {
     enable = true;
     openFirewall = true;
+    enabledCollectors = [ "textfile" ];
+    extraFlags = [
+      "--collector.textfile.directory=${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR}"
+    ];
   };
+
+  services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR = "/var/lib/nix-security-tracker/metrics";
+
+  systemd.tmpfiles.rules = [
+    "d ${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR} 2750 nix-security-tracker ${config.services.prometheus.exporters.node.user} -"
+  ];
 
   services.prometheus.exporters.postgres = {
     enable = true;
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -348,7 +348,11 @@ in
           ];
           wantedBy = [ "multi-user.target" ];
 
-          serviceConfig.Type = "oneshot";
+          serviceConfig = {
+            Type = "oneshot";
+            # Make performance metrics file, produced as a side effect, readable by Prometheus node exporter
+            UMask = "0027";
+          };
           script = ''
             wst-manage backfill_package_clustering
             wst-manage regenerate_cached_suggestions
--- a/src/project/settings.py
+++ b/src/project/settings.py
@@ -100,6 +100,12 @@ class Settings(BaseSettings):
             By default, in the root of this Git repository.
             """
         )
+        METRICS_TEXTFILE_DIR: DirectoryPath | None = Field(
+            description="""
+            Directory for Prometheus textfile metrics consumed by node_exporter.
+            """,
+            default=None,
+        )
         CHANNEL_MONITORING_URL: HttpUrl = Field(
             description="""
             URL from which to fetch the current channel structure.
--- a/src/shared/management/commands/regenerate_cached_suggestions.py
+++ b/src/shared/management/commands/regenerate_cached_suggestions.py
@@ -5,8 +5,10 @@ from typing import Any
 
 from django.core.management.base import BaseCommand
 from django.db.models import Exists, OuterRef, Q
+from prometheus_client import CollectorRegistry, Gauge
 
 from shared.cache_suggestions import cache_new_suggestions
+from shared.metrics import write_metrics_textfile
 from shared.models.cached import CachedSuggestions
 from shared.models.linkage import CVEDerivationClusterProposal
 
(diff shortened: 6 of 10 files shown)


src/shared/metrics.py
```diff
@@ -0,0 +1,50 @@
...
+            labelnames=labelnames,
+            registry=registry,
+        )
+        if labels:
+            gauge.labels(**labels).set(value)
+        else:
+            gauge.set(value)
+    prom_path = path if path.suffix == ".prom" else path.with_suffix(".prom")
+    write_to_textfile(str(prom_path), registry)
+    os.chmod(prom_path, 0o644)
```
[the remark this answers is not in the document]
DarshanCode2005: -_-

contrib/grafana-dashboard.json
Erethon, line 1: The dashboard contains a template variable called `Instance` that allows us to select which instance's (production/staging/some-future-one) metrics to view. You should filter based on that variable on the newly added panels, otherwise the queries will fail since Prometheus will return two time-series (one for prod and one for staging).

The diff would look something like this:
```
-         "expr": "sectracker_cache_regeneration_duration_seconds{job=\"node\"}",
+         "expr": "sectracker_cache_regeneration_duration_seconds{job=\"node\", instance=\"$Instance\"}",
```
this should be done for both panels.

src/shared/management/commands/regenerate_cached_suggestions.py
```diff
@@ -61,9 +66,18 @@ def handle(self, *args: Any, **options: Any) -> None:
...
-        # FIXME(@fricklerhandwerk): Do structured logging for actual metrics so we can keep track of how we're doing.
-        duration = (time.time() - start,)
+        duration = time.time() - start
+
+        write_batch_job_metrics(
+            Path(settings.METRICS_TEXTFILE_DIR),
+            "cache_regeneration",
+            {
+                "sectracker_cache_regeneration_duration_seconds": duration,
+                "sectracker_cache_regeneration_suggestions_total": float(count),
```
Erethon: These metrics were registered as Gauges earlier (and rightfully so). However, `_total` is a suffix that Prometheus will add automatically on Counters and adding it to Gauges is against its [naming conventions](https://prometheus.io/docs/instrumenting/writing_exporters/#naming) (plus it's confusing as a user seeing a `_total` for something that's not a total).

src/shared/metrics.py
```diff
@@ -0,0 +1,50 @@
...
+
+
+def write_batch_job_metrics(
```
Erethon: I don't see much point in having this function, the work is already done in `write_textfile_gauges`. I would remove this completely.

src/shared/metrics.py
```diff
@@ -0,0 +1,50 @@
...
+
+from prometheus_client import CollectorRegistry, Gauge, write_to_textfile
+
+
+def write_textfile_gauges(
+    path: Path,
+    gauges: Mapping[str, float],
+    *,
+    labels: Mapping[str, str] | None = None,
+    help_text: str | None = None,
```
Erethon: `help_text` was never used in `write_batch_job_metrics` (which I recommend we remove in another comment). Having said that, it should also be a `Mapping` since we want different help_texts for each metric, otherwise all metrics would get the same help text.

Erethon (dismissed):
I've left an early review. Besides the inline comments, can you also fix the python formatting so that CI is happy please?

infra/common.nix
fricklerhandwerk:
```suggestion
{ config, pkgs, lib, ... }:
```

infra/common.nix
```diff
@@ -85,6 +85,8 @@ in
   services.prometheus.exporters.node = {
     enable = true;
     openFirewall = true;
+    enabledCollectors = [ "textfile" ];
+    extraFlags = [ "--collector.textfile.directory=/var/lib/nix-security-tracker/metrics" ];
```
fricklerhandwerk: Roughly:

```suggestion
    extraFlags = [ "--collector.textfile.directory=${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR}" ];
```

That way it can't go wrong any more

src/project/settings.py
```diff
@@ -330,6 +330,9 @@ class AppSettings(BaseModel):
     Path(BASE_DIR / ".." / "nixpkgs-evaluation-logs").resolve()
 )
 CVE_CACHE_DIR: str = str(Path(BASE_DIR / ".." / "cve-cache").resolve())
+METRICS_TEXTFILE_DIR: str = str(
+    Path(BASE_DIR / ".." / "prometheus-metrics").resolve()
```
fricklerhandwerk: Add this to `.gitignore`. And please generally move new config parameters into the above Pydantic model (we discussed in sync what it would look like roughly).

src/shared/metrics.py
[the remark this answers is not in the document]
fricklerhandwerk: This can be worked around by adding e.g. prometheus to the right group at the configuration level so it can read from that directory without messing with permissions. Please don't forget to also make it work in the dev setup (I can test and fix it up, but the code change should be there at least in spirit -- ideally without duplicating the mechanism).

src/shared/management/commands/regenerate_cached_suggestions.py
DarshanCode2005: Okay, got it. Removing _total

src/shared/metrics.py
DarshanCode2005: Okay , removed this and updated the function call with the `write_textfile_gauges`

src/shared/metrics.py
DarshanCode2005: Ahh! Got it.

src/shared/metrics.py
```diff
@@ -0,0 +1,39 @@
...
+            labelnames=labelnames,
+            registry=registry,
+        )
+        if labels:
+            gauge.labels(**labels).set(value)
+        else:
+            gauge.set(value)
+    prom_path = path if path.suffix == ".prom" else path.with_suffix(".prom")
+    write_to_textfile(str(prom_path), registry)
+    os.chmod(prom_path, 0o644)
```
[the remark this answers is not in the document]
fricklerhandwerk: The fix is to instead give the service and prometheus the same groups to read from the files. Some trickery is probably needed to make it also work in the development shell

src/shared/metrics.py
```diff
@@ -0,0 +1,44 @@
...
+        )
+        if labels:
+            gauge.labels(**labels).set(value)
+        else:
+            gauge.set(value)
+    prom_path = path if path.suffix == ".prom" else path.with_suffix(".prom")
+    write_to_textfile(str(prom_path), registry)
+    file_mode = os.environ.get("METRICS_TEXTFILE_FILE_MODE")
+    if file_mode is not None:
+        os.chmod(prom_path, int(file_mode, 8))
```
fricklerhandwerk: I still don't think this permission handling should even be here. Need to check more closely, but intuitively all that should be set up before that. And we really shouldn't read from an environment variable ad hoc. The (currently undocumented) rule is that all configuration goes through `settings.py`.

default.nix
```diff
@@ -117,6 +117,7 @@ rec {
               };
             in
             if git ? dirtyRev then "${git.shortRev}-dirty" else git.shortRev;
+          METRICS_TEXTFILE_DIR = toString ./. + "/prometheus-metrics";
```
fricklerhandwerk, line 123: Don't we need a shell hook to make sure the directory exists?

src/project/settings.py
```diff
@@ -100,6 +100,14 @@ class DjangoSettings(BaseModel):
             By default, in the root of this Git repository.
             """
         )
+        METRICS_TEXTFILE_DIR: Path = Field(
+            description="""
+            Directory for Prometheus textfile metrics consumed by node_exporter.
+            """,
+            default_factory=lambda: (
+                Path(__file__).resolve().parent.parent.parent / "prometheus-metrics"
+            ),
```
fricklerhandwerk:
```suggestion
```

Since we're setting it in `configuration.nix` and `default.nix`, is there any compelling reason to have a default in the first place? I always try to avoid defaults for things that will break at runtime or invisibly if you don't set them right. It already happened that such a value was not set in `default.nix` and someone miscalculated or forgot to update the number of directories to go up...

src/project/settings.py
```diff
@@ -534,3 +542,4 @@ class AppSettings(BaseModel):
 CSRF_COOKIE_SECURE = not DEBUG  # noqa: F405 # pyright: ignore [reportUndefinedVariable]
 
 Path(EVALUATION_LOGS_DIRECTORY).mkdir(exist_ok=True)
+Path(METRICS_TEXTFILE_DIR).mkdir(parents=True, exist_ok=True)  # noqa: F405 # pyright: ignore [reportUndefinedVariable]
```
fricklerhandwerk: You're supposedly already doing that in `configuration.nix` for the deployment case. I suggest adding an equivalent line to the shell hook to `default.nix`. Application settings really shouldn't perform side effects, IMO.

fricklerhandwerk: PS: Everything below expanding the Pydantic value and here can be considered messy legacy from prototype days. Don't copy what that stuff does, think for yourself if there's no precedent.

@florentc and I had a discussion about this too the other day, and the preliminary result was that we should probably have a self-contained pre-processing step for configuration values that need it, possibly in the form of another Pydantic value that takes `DJANGO_SETTINGS` as input. It really annoys me that we don't natively have the equivalent of `let ... in ...` here in Python, because really that's what we want to express.

src/shared/tests/test_metrics.py
```diff
@@ -0,0 +1,49 @@
...
+        },
+        labels={"mode": "all"},
+        help_texts={
+            "sectracker_cache_regeneration_duration_seconds": (
+                "Duration of last cache regeneration run"
+            ),
+            "sectracker_cache_regeneration_suggestions": (
+                "Suggestions regenerated in last run"
+            ),
+        },
```
fricklerhandwerk: How about changing the interface such that the invocation keeps together what belongs together rather than requiring the caller to correctly apply string identifiers?

```suggestion
        {
            "sectracker_cache_regeneration_duration_seconds": {
                "help_text": "Duration of last cache regeneration run",
                 "value": 42.0,
             },
            "sectracker_cache_regeneration_suggestions": {
                "help_text": "Suggestions regenerated in last run",
                "value": 7.0,
            },
        },
       labels={"mode": "all"},
```

Or even better, make a helper data type that makes mistyping field names a lot more unlikely:
```suggestion
        {
            "sectracker_cache_regeneration_duration_seconds": Metric(
                help_text="Duration of last cache regeneration run",
                value=42.0,
             ),
            "sectracker_cache_regeneration_suggestions": Metric(
                help_text="Suggestions regenerated in last run",
                value=7.0,
            ),
        },
        labels={"mode": "all"},
```

Although at that point I wonder why we're not creating `Gauge()` directly here? I don't think we need any extra convenience that will infer details if we don't supply them. Just make the caller supply the details.

src/shared/metrics.py
DarshanCode2005: Hmm, agreed. I am removing the chmod and metrics_text-file_mode entirely, So I am moving production permissions to infra-side.

default.nix
DarshanCode2005: Okay , adding mkdir -p for prometheus-metrics in the shell hook, that way the directory exists before any manage command runs in the dev shell.

src/project/settings.py
DarshanCode2005: Okay, while fixing the CI issue, I was a bit confused , added this while fixing this issue , but forgot to remove :cry:

src/project/settings.py
DarshanCode2005: Alright, happy to leave a follow-up for a cleaner preprocessing step; didn't want to expand scope of this wiring PR.

src/shared/tests/test_metrics.py
DarshanCode2005: refactored as suggested, write_textfile_gauges now only takes a CollectorRegistry and writes it to the textfile callers register fully configured gauge instances, with name, help text, labels and value all in one place. Also dropped the parallel dicts.

nix/configuration.nix
```diff
@@ -184,6 +184,7 @@ in
         EVALUATION_LOGS_DIRECTORY = mkDefault "/var/log/nix-security-tracker/evaluation";
         LOCAL_NIXPKGS_CHECKOUT = mkDefault "/var/lib/nix-security-tracker/nixpkgs-repo";
         CVE_CACHE_DIR = mkDefault "/var/lib/nix-security-tracker/cve-cache";
+        METRICS_TEXTFILE_DIR = mkDefault "/var/lib/nix-security-tracker/metrics";
```
fricklerhandwerk: This needs to go to the other `settings =` -- sorry, this settings jungle is a mess, I'm in the process of cleaning it up so this sort of friction doesn't occur in the future https://github.com/NixOS/nix-security-tracker/pull/1166

fricklerhandwerk (approved):
LGTM now! @Erethon final words?

fricklerhandwerk:
Apart from formatting this seems correct now. I can't merge it yet because I ran into an unrelated problem introduced by https://github.com/NixOS/nix-security-tracker/pull/1130, which makes it impossible to regenerate the cache and try it out the metrics locally.

infra/common.nix
```diff
@@ -85,8 +90,18 @@ in
...
+    enabledCollectors = [ "textfile" ];
+    extraFlags = [
+      "--collector.textfile.directory=${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR}"
+    ];
   };
 
+  services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR = "/var/lib/nix-security-tracker/metrics";
+
+  systemd.tmpfiles.rules = [
+    "d ${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR} 2750 nix-security-tracker node-exporter -"
```
fricklerhandwerk:
```suggestion
    "d ${config.services.nix-security-tracker.settings.METRICS_TEXTFILE_DIR} 2750 nix-security-tracker ${config.services.prometheus.exporters.node.user} -"
```

Let's be even more explicit here

DarshanCode2005:
Took a while to debug , but did it : )

The issue was that after @fricklerhandwerk refactored `write_metrics_textfile()` to accept a filename (`name: str`) and use `METRICS_TEXTFILE_DIR` from settings, the tests weren't updated. They were still passing a full `Path` object (`tmp_path / "sample"`). Since `tmp_path` is an absolute path, `pathlib` ignored `METRICS_TEXTFILE_DIR` entirely, so the tests wrote directly to the temporary directory instead of using the configured metrics directory. As a result, the tests still passed but weren't actually verifying the production behavior.

To fix this, the tests now use `override_settings(METRICS_TEXTFILE_DIR=tmp_path)` so they're isolated and don't depend on the environment. They also pass simple string filenames like `"sample"` and `"cache_regeneration"`, which matches how the function is used in production, for example in `regenerate_cached_suggestions.py`. The implementation of `src/shared/metrics.py` itself didn't change. Only the tests were updated to correctly reflect the refactor.

nix/configuration.nix
```diff
@@ -348,7 +348,10 @@ in
           ];
           wantedBy = [ "multi-user.target" ];
 
-          serviceConfig.Type = "oneshot";
+          serviceConfig = {
+            Type = "oneshot";
+            UMask = "0027";
```
fricklerhandwerk, line 354:
```suggestion
            # Make performance metrics file, produced as a side effect, readable by Prometheus node exporter
            UMask = "0027";
```

fricklerhandwerk:
Unfortunately I can't test this locally for dumb reasons (related https://github.com/NixOS/nix-security-tracker/issues/1045#issuecomment-4611421058), but this is well-tested and I trust it will work and at worst need small fixups in the future.

PR #1146, opened 2026-06-23
feat: fetch channel information from Hydra

fricklerhandwerk:
Further towards https://github.com/NixOS/nix-security-tracker/issues/864

Since the source of truth for latest evaluations and builds is in Hydra,
from now on we're only using the infrastructure team's `channel.nix`
for information about the release state of channels.


src/project/settings.py
DarshanCode2005: fetch_all_channels no longer uses it. So this could be removed or deprecated with a comment

src/shared/git.py
```diff
@@ -4,17 +4,32 @@
...
 logger = logging.getLogger(__name__)
 
 
+def get_head_sha1(url: AnyUrl, branch: str) -> str:
+    result = subprocess.run(
+        ["git", "ls-remote", str(url), f"refs/heads/{branch}"],
+        capture_output=True,
+        text=True,
+        check=True,
+    )
```
DarshanCode2005, line 27: `timeout` would be nice.

src/shared/hydra.py
```diff
@@ -0,0 +1,58 @@
...
+        self.base_url = base_url.rstrip("/")
+
+    def get_jobset(self, jobset_path: str) -> Jobset:
+        """
+        Fetches jobset metadata for the given `<project>/<jobset>` path.
+        """
+        resp = requests.get(
+            f"{self.base_url}/jobset/{jobset_path}",
+            headers={"Accept": "application/json"},
+        )
```
DarshanCode2005, line 113: `timeout` would be nice.

src/project/settings.py
fricklerhandwerk: Good catch

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -1,81 +1,51 @@
...
-        print("Parallel fetches results", results)
+        for channel_name, channel in sorted_channels:
+            jobset_path = jobset_from_job(channel.job)
+            jobset = client.get_jobset(jobset_path)
+            branch_name = jobset.input_branch(
+                "nixpkgs", default=settings.TRACKING_BRANCH
+            )
+
+            if channel.variant == channel.Variant.SMALL:
+                commit_sha1 = get_head_sha1(settings.GIT_CLONE_URL, branch_name)
```
fricklerhandwerk: To be more honest, we should maybe actually take the actual evaluated commit here? At least that's what one would expect, but then again what is done now will get the *latest available* commit, which is closer to maintainers' ground truth.

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -1,81 +1,51 @@
...
-            pprint(branch_info | {"channel_branch": channel.name})
-            NixChannel.objects.update_or_create(
-                branch_info, channel_branch=channel_branch
-            )
-
-        repo = GitRepo(
-            settings.LOCAL_NIXPKGS_CHECKOUT,
-            stderr=sys.stderr.fileno(),
+        client = default_client()
+        # Small channels first so the other variants can inherit their branch commit.
```
fricklerhandwerk: Should maybe note here that those are closest to `master`.

src/shared/management/commands/fetch_all_channels.py
```diff
@@ -1,81 +1,51 @@
...
+                )
+            else:
+                branch = NixpkgsBranch.objects.get(name=branch_name)
+
+            ch, created = NixChannel.objects.update_or_create(
+                channel_branch=channel_name,
+                defaults={
+                    "release_branch": branch,
+                    "state": channel.status,
+                    "head_sha1_commit": branch.head_sha1_commit,
```
fricklerhandwerk: In fact this is wrong, because we need to make a clear distinction between channel and its associated release branch. In practice users will never take the tip of master for `nixpkgs-unstable`, because then they'd have to build a lot of stuff locally. But the point of channels for us is to show what's shipped so when a vuln is fixed on a channel, that commit is what users should at least upgrade to.

src/shared/release_channels.py
```diff
@@ -0,0 +1,31 @@
...
+
+from shared._release_channels import (  # type: ignore[reportMissingImports]
+    channels as _raw,  # Populated at build time from `github:NixOS/infra//channels.nix`
+)
+from shared.models.nix_evaluation import NixChannel
+
+ChannelState = NixChannel.ChannelState
+
+
+class Channel(BaseModel):
```
fricklerhandwerk: We may want to stuff Hydra output directly into the database model and simply run model-level validation on it.

fricklerhandwerk:
Reviewed with @florentc

src/shared/release_channels.py
fricklerhandwerk: Actually no, that would be tight coupling...

fricklerhandwerk:
This took quite a while. Touching the concept just a bit required a ton of rework in detail. But also, doing this diligently revealed how many moving parts there actually are, which we were generously papering over previously.

I was particularly paranoid about testing things, since we'll have fun if something goes wrong in a live deployment. That produced most of the change, because making it both robust and sustainable is not really straightforward.

There are were so many loose ends there, spend days chasing them, and now I just realized the integration test is probably still broken. Though right now I'm stuck figuring out why `nix-shell --run format` passes but `nix-build -A git-hooks` wants to reorder imports...

PR #1150, opened 2026-06-25
fix: Require POST for social login instead of GET

Erethon:
The [docs of allauth](https://docs.allauth.org/en/dev/socialaccount/configuration.html) specify:
> For security considerations, it is strongly recommended to require POST requests.

This will add an extra step where the user needs to click on a "Continue" button before they can login. I'm opening this PR to discuss if this is a UX decision we made consciously or something that we happened to include. If it's the former, then I'll edit this PR to add a comment to the setting, because it caught me by surprise when looking at the code.

### DIFF
--- a/src/project/settings.py
+++ b/src/project/settings.py
@@ -486,8 +486,6 @@ SITE_ID = 1
 
 # Disable regular signup but allow GitHub auth
 SOCIALACCOUNT_ONLY = True
-# Skip intermediate "Continue" page, redirect to GitHub immediately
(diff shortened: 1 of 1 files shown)


fricklerhandwerk:
IIRC the problem @florentc was that the API-based UI client doesn't eat the HTML response well. But we can use https://docs.allauth.org/en/latest/headless/index.html in that case.

Erethon:
Closing in favor of #1200

PR #1154, opened 2026-06-28
GET /notifications

devnchill:
Related to #1024 (https://github.com/NixOS/nix-security-tracker/issues/1024#issuecomment-4804941025) 

> notifications/ seems like a pretty standalone thing to tackle @devnchill, and I can implement the front side of things after. API only

- Adds REST API endpoint for GET `notifications` 
- Tests for GET `notifications`

### DIFF
--- a/src/api/notifications/views.py
+++ b/src/api/notifications/views.py
@@ -1,0 +1,132 @@
+from enum import StrEnum
+from typing import Any
+
+from django.db.models import QuerySet
+from drf_spectacular.utils import extend_schema, extend_schema_field
+from rest_framework import serializers, viewsets
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.permissions import IsAuthenticated
+from rest_framework.request import Request
+from rest_framework.response import Response
+
+from api.suggestions.serializers import SuggestionPackageSerializer
+from shared.cache_suggestions import CachedSuggestion
+from webview.models import Notification, SuggestionNotification, TextNotification
+
+
+class NotificationType(StrEnum):
+    TEXT = "text"
+    SUGGESTION = "suggestion"
+
+
+class NotificationSerializer(serializers.ModelSerializer):
+    is_obsolete = serializers.SerializerMethodField()
+    message = serializers.SerializerMethodField()
+    matching_maintained_packages = serializers.SerializerMethodField()
+    matching_subscribed_packages = serializers.SerializerMethodField()
+    suggestion_id = serializers.SerializerMethodField()
+    title = serializers.SerializerMethodField()
+    type = serializers.SerializerMethodField()
+
+    class Meta:
+        model = Notification
+        fields = [
+            "created_at",
+            "id",
+            "is_obsolete",
+            "is_read",
+            "message",
+            "matching_maintained_packages",
+            "matching_subscribed_packages",
+            "suggestion_id",
+            "title",
+            "type",
+        ]
+
+    def get_is_obsolete(self, notif: Notification) -> bool:
+        if isinstance(notif, TextNotification):
+            return False
+        return not (
+            self.get_matching_maintained_packages(notif)
+            or self.get_matching_subscribed_packages(notif)
+        )
+
+    def get_message(self, notif: Notification) -> str | None:
+        if isinstance(notif, TextNotification):
+            return notif.message
+        return None
+
+    @extend_schema_field(serializers.DictField(child=SuggestionPackageSerializer()))
+    def get_matching_maintained_packages(
+        self, notif: Notification
+    ) -> dict[str, Any] | None:
+        username = self.context["request"].user.username
+        if isinstance(notif, SuggestionNotification):
+            cached = CachedSuggestion.model_validate(notif.suggestion.cached.payload)
+            return {
+                pname: SuggestionPackageSerializer(pkg).data
+                for pname, pkg in cached.packages.items()
+                if any(m.github == username for m in pkg.maintainers)
+            }
+        return None
+
+    @extend_schema_field(serializers.DictField(child=SuggestionPackageSerializer()))
+    def get_matching_subscribed_packages(
+        self, notif: Notification
+    ) -> dict[str, Any] | None:
+        if isinstance(notif, SuggestionNotification):
+            cached = CachedSuggestion.model_validate(notif.suggestion.cached.payload)
+            subscribed_attrs = set(
+                self.context["request"].user.profile.package_subscriptions
+            )
+            return {
+                attr: SuggestionPackageSerializer(cached.packages[attr]).data
+                for attr in cached.packages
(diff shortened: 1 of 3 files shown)


src/api/notifications/views.py
```diff
@@ -0,0 +1,40 @@
...
+
+    def get_title(self, obj: Notification) -> str:
+        return obj.title
+
+
+class NotificationPagination(PageNumberPagination):
+    page_size = 20
+    page_size_query_param = "per_page"
+    max_page_size = 100
+
```
DarshanCode2005, line 107: the viewset has no pagination_class and REST has no global pagination, and the tests assume flatlist , either you could wire it up and update tests or remove until pagination is needed.

DarshanCode2005:
tests are nice :heart: 

suggestion: you should implement tests for GET `/notifications/:id` example:
1> retrieve own notification - 200
2> retrieve another user's notification - 404
3> retrieve non-existent id - 404

DarshanCode2005:
for notification frontend will most likely require message for `TextNotification` you could confirm it with @florentc sir.

DarshanCode2005:
@devnchill you could add @extend_schema for OPENAPI.

Also, `test_token_auth.py` probes auth on other endpoints, but notifications have no equivalent , a small test would document that API-key clients can use this endpoint.

src/api/notifications/views.py
devnchill: forgot to add it  T_T , i will wire it up and update tests , thanks for pointing that out :sweat_smile:

devnchill:
> tests are nice ❤️
> 
> suggestion: you should implement tests for GET `/notifications/:id` example: 1> retrieve own notification - 200 2> retrieve another user's notification - 404 3> retrieve non-existent id - 404

thanks :smile: , yeah I was aware about missing test for that but thought it would it would be best practice to write tests for `/notifications/:id`  in separate pr , this was meant just for GET /notification ... and /notification:id is kind of sideeffect of ReadOnly viewset :sweat_smile: ... @florentc  @fricklerhandwerk   any thoughts ....

devnchill:
resolved merge conflicts and rebased :sweat_smile:

florentc:
Thanks for the PR and review. We'll indeed want the text content in the serialized text notifications. As for the suggestion notifications, we'll want the suggestion id so that the frontend or the user can query this suggestion later. I'm currently working on the suggestions API including the suggestion serializer so for now in this PR let's only return the suggestion_id. If needed in the future we can inline the suggestion in the response.

florentc (commented):
I'd like the API endpoint to also return, as structured data, the information presented in the current web UI in case the notification is about a suggestion:
- which packages match what the user maintains (attribute names)
- which packages match what the user subscribed to (attribute names)
- whether the notification is obsolete (boolean)

You may find webview tests that illustrate those cases.

devnchill:
alr :laughing:

src/api/urls.py
```diff
@@ -4,13 +4,15 @@
...
 from api.me import CurrentUserView
+from api.notifications.views import NotificationViewSet
 from api.server_info import ServerInfoView
 from api.subscriptions.views import SubscriptionsViewSet
 from api.suggestions.views import SuggestionViewSet
 from api.tokens.views import TokenManagementView
 
 v1_router = routers.DefaultRouter(trailing_slash=False)
 v1_router.register(r"issues", NixpkgsIssueViewSet)
+v1_router.register(r"notifications", NotificationViewSet, basename="notification")
```
florentc: Why do we need `basename` and "notification" without s?

src/api/tests/test_token_auth.py
```diff
@@ -105,3 +108,25 @@ def test_bearer_auth_change_status_permissions(
...
+    print(f"content={response.content[:500]}")
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data["count"] == 0
+
+
+@pytest.mark.django_db
+def test_bearer_auth_notifications_no_token() -> None:
+    client = APIClient()
+    response = client.get(NOTIFICATIONS_URL)
+    assert response.status_code == status.HTTP_401_UNAUTHORIZED
```
florentc: What is the purpose of these tests in test_token_auth? Doesn't seem relevant at first sight.

florentc (changes requested):
I don't see the data I talked about in my previous message. The list of matched packages (aka what triggered the notification). It has been introduced in the webview in #867 

The rest of the api has extend_schema annotations to name and document endpoints in the openAPI spec. Please fit the convention.

devnchill:
> I don't see the data I talked about in my previous message. The list of matched packages (aka what triggered the notification). It has been introduced in the webview in #867
> 
> The rest of the api has extend_schema annotations to name and document endpoints in the openAPI spec. Please fit the convention.

Yeah i'm working on it , was lil confused on the payload of cached suggestion

src/api/urls.py
devnchill: image
when i was going through the docs, i found out the basename is required if there is no queryset attr , since notification uses `get_queryset` method , I added it, although I see that it `notifications` would be better

src/api/tests/test_token_auth.py
devnchill: iirc it was in response to darsh's suggestion , he suggested to add a small example just to show that it would work for api key based clients :sweat_smile:

florentc: No need to repeat that test for various protected endpoints IMO. We have not added new cases for every new endpoints and this one doesn't seem to break any pattern. Besides, if we were to do it (which I don't think is very useful) we'd rather deduplicate (e.g. parameterize a single test) since the logic is always the same.

florentc (approved):
The policy on this repo is to squash & merge but this can't be enforced automatically as we use the merge queue. When ready please squash.

src/api/urls.py
```diff
@@ -4,13 +4,15 @@
...
 from api.me import CurrentUserView
+from api.notifications.views import NotificationViewSet
 from api.server_info import ServerInfoView
 from api.subscriptions.views import SubscriptionsViewSet
 from api.suggestions.views import SuggestionViewSet
 from api.tokens.views import TokenManagementView
 
 v1_router = routers.DefaultRouter(trailing_slash=False)
 v1_router.register(r"issues", NixpkgsIssueViewSet)
+v1_router.register(r"notifications", NotificationViewSet, basename="notifications")
```
florentc: Why a `basename` here?

devnchill: i answered it here https://github.com/NixOS/nix-security-tracker/pull/1154#discussion_r3657918182 , I don't really know if there is a way to escape this T_T . I noticed that `SuggestionViewSet` has both `get_queryset` and `queryset` attr however i'm not sure what to do for Notification case. 

Also some viewsets are singular,some plural . I think it would be nice to have some convention regarding it..

PR #1155, opened 2026-06-29
UI v2: Port all user settings features from legacy UI

florentc:
This ports all legacy UI features related to user settings:
- package subscriptions
- auto subscribe 
- email notification
- individual package subscription pages
- api tokens

This includes implementing the missing API endpoints in Django, testing the API, implementing the frontend, testing the frontend.

There is a slight difference from legacy in terms of user facing routes: all are prefixed by `/user` out of consistency. Individual package subscription pages are under `/user/subscriptions/packages/{package_name}` (`packageS` plural by convention).

This PR also brings styling improvements and some future proofing (in particular, revisions to the utility classes, and removal of ad-hoc CSS in favor of those classes).

This uses some UI components from the ArkUI component library (light unstyled components). Namely the switches, the menu, and the tabs. There are many shenanigans related for instance to keyboard navigation, or accessibility, that ArkUI components implement correctly (compared to reinventing the wheel ourselves).

This also introduces, out of topic (can be removed from the PR), a brief summary on the homepage about which features are currently ported on UI v2, so users know what to expect. A link invites them to go back to the legacy UI.


src/api/subscriptions/views.py
```diff
@@ -15,6 +19,25 @@ class SubscriptionsViewSet(viewsets.GenericViewSet):
...
+        )
+
+    class PackageSubscriptionStatusSerializer(serializers.Serializer):
+        subscribed = serializers.BooleanField(
+            help_text="Whether the authenticated user is subscribed to this package.",
+        )
+
+    class AddPackageSubscriptionSerializer(serializers.Serializer):
+        package_name = serializers.CharField(
+            help_text="Nix attribute name of the package to subscribe to (e.g. 'firefox', 'python311Packages.requests').",
```
fricklerhandwerk:
```suggestion
            help_text="Nixpkgs attribute name of the package to subscribe to (e.g. 'firefox', 'python311Packages.requests').",
```

src/api/subscriptions/views.py
```diff
@@ -15,6 +19,25 @@ class SubscriptionsViewSet(viewsets.GenericViewSet):
     class AutoSubscribeSerializer(serializers.Serializer):
         enabled = serializers.BooleanField()
 
+    class EmailNotificationsSerializer(serializers.Serializer):
+        enabled = serializers.BooleanField()
+
+    class PackageSubscriptionsSerializer(serializers.Serializer):
+        packages = serializers.ListField(
+            child=serializers.CharField(),
+            help_text="Package attribute names the user has manually subscribed to.",
```
fricklerhandwerk:
```suggestion
            help_text="Package attribute names the user has manually subscribed to",
```

I think generally these are not sentences...

src/api/subscriptions/views.py
```diff
@@ -50,3 +73,140 @@ def auto_subscribe_to_maintained_packages(self, request: Request) -> Response:
...
+                {"packages": profile.package_subscriptions}
+            )
+            return Response(serializer.data, status=status.HTTP_200_OK)
+        elif request.method == "POST":
+            input_serializer = self.AddPackageSubscriptionSerializer(data=request.data)
+            input_serializer.is_valid(raise_exception=True)
+            validated: dict[str, Any] = input_serializer.validated_data  # type: ignore[assignment]
+            package_name = str(validated["package_name"]).strip()
+
+            if not NixDerivation.objects.filter(attribute=package_name).exists():
```
fricklerhandwerk: We may want to use the package data model here already, and instead query from the much smaller table `PackageAttribute` (which has only ~100k entries as opposed to ~100M)

frontend/src/routes/PackageSubscriptionPage.tsx
fricklerhandwerk, line 1: I think we should skip this for now. If we want to have users send around the subscription page for a package, we still have the original UI, and for the new UI we probably want to just add the toggle as a widget to the per-package page.

src/api/subscriptions/views.py
```diff
@@ -50,3 +73,140 @@ def auto_subscribe_to_maintained_packages(self, request: Request) -> Response:
...
+        description="Get the current email notifications preference.",
+    )
+    @extend_schema(
+        methods=["PUT"],
+        operation_id="setEmailNotifications",
+        description="Update the email notifications preference.",
+    )
+    @action(
+        detail=False,
+        methods=["get", "put"],
```
fricklerhandwerk:
```suggestion
        methods=["GET", "PUT"],
```

Pick one consistently

frontend/src/hooks/useTokenStatus.ts
fricklerhandwerk, line 1: IIUC this is a fixup for what the generated schema does with the token API. Can we merge this into the Orval post-processing instead?

frontend/src/components/ui/Menu.module.css
```diff
@@ -0,0 +1,50 @@
...
+}
+
+.content {
+  background: white;
+  color: black;
+  border: 1px solid var(--gray);
+  border-radius: 0.4em;
+  box-shadow: 0 4px 12px rgba(0, 0, 0, 0.15);
+  padding: 0.25em 0;
+  min-width: 12em;
```
fricklerhandwerk, line 26: Random note, probably too late: Shouldn't we generally use root sizing, so we can never run into trouble when nesting components?

```suggestion
  min-width: 12rem;
```

fricklerhandwerk (approved):

frontend/src/components/header/AuthStatus.tsx
```diff
@@ -1,38 +1,52 @@
...
-        ) : (
-          <UserIcon className={`${styles.avatar} ${styles.placeholder}`} />
-        )}
-      </Link>
-      <button type="button" onClick={logout} className={styles.logoutButton} title="Logout">
-        <LogOutIcon />
-      </button>
-    </div>
+    <Menu
+      trigger={<Avatar avatarUrl={user.avatar_url} username={user.username} />}
```
fricklerhandwerk: This currently breaks our little convention to textually annotate interactive elements. How about we annotate this with "Account" (since this is constant space as opposed to the login name)? Or at least the login name?

fricklerhandwerk:

PR #1157, opened 2026-06-29
feat: evaluate tips of release branches

### LINKED ISSUE #864: Also evaluate release branches
Needs refinement:

Currently we're only evaluating [channel branches](https://nix.dev/concepts/faq.html#which-channel-branch-should-i-use), but often those already contain fixes to a given CVE by the time they're matched against.

Since triaging vulnerabilities is primarily targeting maintainers, we also want to evaluate the corresponding release branches. (Nixpkgs users later will need to know in which channel branches a given fix landed.)

We probably want to display package versions for both release and channel branches, and probably also have a way to specify the commit in which a fix was provided (there's a separate issue for that but it may need further refinement, since commits will differ per channel for e.g. backports).

fricklerhandwerk:
Closes #864

This also makes an incremental move to displaying new evaluations based on
branches as such, while preserving display of channel-oriented evaluations.


fricklerhandwerk:
I'll hold off rebasing this until https://github.com/NixOS/nix-security-tracker/pull/1211 is merged.

PR #1159, opened 2026-06-29
refactor: extract linkage decision logic and always create proposals

adekoder:
Introduce Rejection and LinkageOutcome dataclasses to model the result of evaluating a container's linkage candidates, and extract resolve_linkage_candidates() to separate decision logic from persistence in build_new_links().

build_new_links() now always creates a CVEDerivationClusterProposal, including for NO_MATCHES, previously a silent return False. Add NO_MATCHES as an explicit RejectionReason to make these cases traceable.


fricklerhandwerk (approved):
Nice cleanup!

PR #1160, opened 2026-06-30
feat: obtain server info in new UI through an API endpoint

florentc:
Server info (production vs staging, debug, revision) used to be relayed to UI v2 through the Django template. This introduces an API endpoint to obtain the info. It simplifies the Django template and brings consistency, along with another usable API endpoint.


fricklerhandwerk (approved):
As discussed we should collapse the production and debug flags, since their distinction has become obsolete and looking at what's going on is confusing. But that's for another change.

fricklerhandwerk:

PR #1161, opened 2026-07-01
fix: read CVE problemTypes via the schema's plural 'descriptions' key

### LINKED ISSUE #715: CVE problem type descriptions aren't ingested
The field name ([reference documentation](https://cveproject.github.io/cve-schema/schema/docs/#oneOf_i0_containers_cna_problemTypes)) has a typo, it must be `descriptions` not `description`:

https://github.com/NixOS/nix-security-tracker/blob/aca950262fed92687e90b26ea32d8d7c2819024a/src/shared/fetchers.py#L262

This is a quick fix for new CVEs, but we probably want to backfill existing ingested data, and that requires more machinery that currently doesn't exist. At the very least we need to avoid creating new `CveRecord` and `Container` entries:

https://github.com/NixOS/nix-security-tracker/blob/aca950262fed92687e90b26ea32d8d7c2819024a/src/shared/fetchers.py#L215-L216

https://github.com/NixOS/nix-security-tracker/blob/aca950262fed92687e90b26ea32d8d7c2819024a/src/shared/fetchers.py#L254

CodeBlackwell:
## Summary

`make_container()` in `src/shared/fetchers.py` read
`problem.get("description", [])` when flattening `problemTypes` entries, but
the CVE JSON 5.0 schema keys this field `descriptions` (plural) — see the
[schema reference](https://cveproject.github.io/cve-schema/schema/docs/#oneOf_i0_containers_cna_problemTypes).
The singular key never matches real CVE data, so `.get()` always fell back
to `[]`, meaning **every ingested CVE ends up with zero problem types**
regardless of what the source record actually contains, as reported in #715.

## Fix

One-line key fix: `problem.get("description", [])` → `problem.get("descriptions", [])`.

While investigating, I also checked the concern (raised alongside #715) that
flattening loses each problem's `cweId`/`type` metadata. Per the actual CVE
5.0 schema, each entry in `problemTypes[].descriptions[]` already carries
`lang`, `description`, `cweId`, and `type` together — exactly the flat shape
`make_problem_type()` expects — so the metadata isn't actually lost; this
single key fix is sufficient on its own.

## Tests

Added `test_make_container_ingests_problem_types_with_cwe_metadata` to
`src/shared/tests/test_fetchers.py`, which builds a `problemTypes` payload
matching the real schema shape and asserts both that a `ProblemType` record
is created at all, and that its `cwe_id`/`_type`/`description` fields are
populated correctly. `pyright` clean, all 4 tests in
`shared/tests/test_fetchers.py` pass.

Fixes #715

🤖 Found and fixed via an automated bug-discovery sweep (code-review-graph +
adversarial multi-agent verification), with human review gating every step
from source selection through this PR.

### DIFF
--- a/src/shared/fetchers.py
+++ b/src/shared/fetchers.py
@@ -248,7 +248,7 @@ def make_container(
     problems = [
         desc
         for problem in data.get("problemTypes", [])
-        for desc in problem.get("description", [])
+        for desc in problem.get("descriptions", [])
     ]
     obj.problem_types.set(map(make_problem_type, problems))
     obj.references.set(map(make_reference, data.get("references", [])))
--- a/src/shared/tests/test_fetchers.py
+++ b/src/shared/tests/test_fetchers.py
@@ -1,5 +1,5 @@
-from shared.fetchers import make_metric
-from shared.models.cve import Metric
+from shared.fetchers import make_container, make_metric
+from shared.models.cve import Container, CveRecord, Metric, Organization
 
 
 def test_make_metric_none(
@@ -29,3 +29,59 @@ def test_make_metric_fallback_v3(
     assert metric
     assert metric.format == Metric.Format.V30
     assert metric.vector_string == cvss_v3_metric[Metric.Format.V30]["vectorString"]
+
+
+def _bare_cve_record() -> CveRecord:
+    org, _ = Organization.objects.get_or_create(
+        uuid="11111111-1111-1111-1111-111111111111", short_name="test-org"
+    )
+    return CveRecord.objects.create(cve_id="CVE-2025-0002", assigner=org)
+
+
+# A real CVE Record problemTypes entry per the CVE JSON 5.0 schema: an array of
+# objects that each hold a "descriptions" array, whose items carry lang,
+# description, cweId, type and (optionally) references.
+_PROBLEM_TYPES = [
+    {
+        "descriptions": [
+            {
+                "lang": "en",
+                "description": "Cross-site Scripting (XSS)",
+                "cweId": "CWE-79",
+                "type": "CWE",
+            }
+        ]
+    }
+]
+
+
+def test_make_container_ingests_problem_types_with_cwe_metadata(
+    db: None,
+) -> None:
+    """A CVE container's problemTypes must be ingested at all (the schema
+    keys the array by the plural 'descriptions', not 'description') *and*
+    each ingested ProblemType must keep the cweId/type/description metadata
+    carried on its schema-shaped entry.
+
+    Both symptoms come from the same flatten step (fetchers.py, building the
+    `problems` list from `data["problemTypes"]`): reading the wrong key
+    collapses the list to empty (zero problem types), while dropping fields
+    off the flattened entries would null out cwe_id/_type/description on
+    whatever *is* created. A fix that only addresses one half leaves this
+    single test red.
+    """
+    cve = _bare_cve_record()
+    data = {
+        "providerMetadata": {"orgId": "22222222-2222-2222-2222-222222222222"},
+        "problemTypes": _PROBLEM_TYPES,
+    }
+
+    container = make_container(data, _type=Container.Type.CNA, cve=cve)
+
+    assert container.problem_types.count() == 1
+    problem_type = container.problem_types.get()
+    assert problem_type.cwe_id == "CWE-79"
+    assert problem_type._type == "CWE"
+    assert list(problem_type.description.values_list("value", flat=True)) == [
+        "Cross-site Scripting (XSS)"
+    ]


devnchill:
are you ai ? nvm missed bottom section

CodeBlackwell:
@devnchill Ai Engineer trying to build my own human in the loop assisted system. Any feedback 100% accepted and immediately considered in all future submission. I am human, I would like to actually contribute and help.

devnchill:
Oh, I just got surprised that you can automate such tasks like finding issues , solving it and even creating prs for fix

fricklerhandwerk:
Yeah but that's trivial these days if you pay LLM hosting providers enough. Just wish it into existence.

The problem is, this change is extremely noisy, it takes quite some effort to see why it's correct. It may sound dumb, but I've experienced it so often that I'll write it down: having done code reviews for years, by now I see from a distance, by the jaggedness of the code, whether it makes sense, whether it's friendly to readers with certain prerequisites, whether it's done with care.

Surely it can be improved automatically to some degree by telling the LLM to imitate the surrounding style better. But my (unprovable, sorry) intuition is, the underlying problem has something to do with conceptual crispness and compression. A person who got a clear understanding of both the problem at hand and its context, as well as a good theory of mind of the readers of the code, will produce changes that look and feel different, and are little effort to review. We may get there with LLMs eventually (at what cost though?) but we're currently not there.

For example, a few weeks of intense collaboration, with @florentc we've done what I previously thought was impossible, and made 1000LOC changes go through review in an hour of video call *with confidence*. Those changes took substantial effort to produce, to be sure, and even more effort to exchange the tacit knowledge, assumptions, boundary conditions, etc. that allowed shaping it such that both of us understand that those 1000LOC don't actually contain all that much information, which would have taken reams of text and unfathomable tedium to write down in a useful fashion.

I've now written quite a message, but that's still a fraction of the effort I feel would be needed to make that patch convincing.

So all I can do here is ask contributors to follow what I consider the most important idea in the [guidelines](https://github.com/NixOS/nix-security-tracker/blob/main/CONTRIBUTING.md#contribution-culture): help maintainers to take responsibility for your change. Your code may be a gift, but accepting it has an ongoing cost which you may not notice because you don't bear it if you submit and leave.

Maintainers could surely help by making the requirements more explicit! I know this would reduce the barrier to entry, I've been working on technical documentation of all sorts of Nix things professionally for years. But that's a very different business; undoubtedly useful, especially to beginners, but not a priority in this project (in fact almost all projects). And breaking down everything to the granularity at which the machine can come up with the correct solution that is legible and maintainable... **is** producing the solution. The bandwidth for that is fundamentally limited, even with machine assistance.

So, @codeblackwell what can we do here? I'm pretty unmotivated to review the change in this state. It ignores a bunch of patterns already in use, which prevents it from fitting in nicely. And pointing that out in each instance is quite the chore. I'd be jumping on it if I get a signal that you actually engaged with the project to some nontrivial extent, because then I can reasonably assume it's worth spending the time because someone seems to be serious about it. You will notice this from other contributions to this repository.


For example, you could do the legwork to identify and document some such patterns in a place and in a way that is maintainable and fits in with the style and structure of existing documentation (essentially answering yourself the question "but why doesn't it currently fit in and what would be needed for it to fit in?"). Maintainable means: won't need to be touched in the foreseeable future. That's a great beginner task! I haven't seen LLMs being able to do that particular thing on their own, because they tend to just repeat what's in the code instead of describing properties that aren't immediately visible from it but shape it. Often such additions end up being one or two sentences, and still can take hours to produce. I will deterministically reciprocate by doing reviews that explain in detail how the results of honest effort may be improved within that framework of correctness and maintainability. And I'll add related permanent documentation myself where it seems fit, so the next person doesn't have to endure the same ordeal of lecturing but instead get a compressed instruction.

How about that?

CodeBlackwell:
@fricklerhandwerk Deal.

Again, The feedback is invaluable. Here's my "counter-proposal" if you will - I'll:

1. Mark this PR as draft for now.
2. Actually spend time in the repo. Merged PRs, the review threads around them, the existing docs. I want to figure out which patterns my change bulldozed past and why they're there. Not what the code does, but the stuff that shapes it that you can't see from any single file.
3. Write that up as a docs contribution that fits what's already there. You set the bar at "won't need to be touched in the foreseeable future," so that's the bar.
4. Then, and only then, come back to this change and rework it inside that framework. Or close it, if what I learn says it deserves closing.

One thing I want to try on top of that, and feel free to shoot it down now. 

You wrote that breaking everything down to the granularity where the machine produces a legible, maintainable solution is producing the solution. 

I think you're **_mostly right_**, and the "**mostly**" is what I want to poke at. While I'm doing step 3 anyway, I'd like to also draft an `AGENTS.md` for the repo: the same patterns and boundary conditions, restated as constraints for anyone contributing with machine assistance. I agree, it will not (yet) give a model the conceptual crispness you're describing, I know that. But some real fraction of what made my patch annoying to review was mechanical (ignoring house patterns, wrong idioms, general jaggedness), and that fraction does respond to written instructions. 

If it buys anything, reviewers get to spend their effort on the part that genuinely needs a human - and less on slop. If it buys nothing, you reject it in one line and the human docs stand on their own.

If you already know where the pattern documentation should live, point me there. Otherwise I'll propose a spot and you can redirect me.

And thanks for taking the time to write all that instead of just closing the PR. I get the asymmetry.


______________________
@devnchill Yes it is relatively trivial to speak wishes unto most "Electric Prayer Wheels" (Hilarious btw) and get some 'meh' to decent results less across the board - Scaling with how much money you dump at it. I've been an engineer for 10 years, so my experience pre-dates LLMs as does my pride in my work. Genuinely I Believe in the power of combining true Engineering with LLMs to automated and accelerate everything, and want to be part of that. I built [crg-debug](https://github.com/CodeBlackwell/crg-debug) (Code-Review-Graph-Debug).

Not to get promotional: a Claude Code plugin that builds a knowledge graph of the codebase, hunts bugs with parallel finder agents, and adversarially verifies each finding so only confirmed, test-backed bugs survive (TDD + Tree of thought adversarial validation) to become PRs. (Under Dev and open to feedback/contribution). To the point of price - It starts with trying to solve it with the smallest model possible, then escalates up a predefined model ladder.

fricklerhandwerk:
I disagree on `AGENTS.md` on the same premise I would disagree with tracking someone's `.vscode/settings.json`. It's your local tooling, and you should avoid tracking it in `.git/info/exclude`. An LLM can read the same documentation as everyone else just fine.

@CodeBlackwell your response reads super weird. I have yet to see anyone with 10 years professional experience who talks like that or behaves like that in public spaces. If you actually wrote this yourself and mean it, then sorry for the misunderstanding, maybe that's just the way you talk. Otherwise, please stop sending spam now. The fix is already merged and was a one-liner after discussing the correct solution for a possible follow-up: https://github.com/NixOS/nix-security-tracker/pull/1162/changes/afeb0e851a77073f4f785fd0f081829165e85869

PR #1162, opened 2026-07-01
Fix CVE problem type descriptions and backfill CVEs missing it 

### LINKED ISSUE #715: CVE problem type descriptions aren't ingested

devnchill:
Resolves #715 
- New management command `backfill_cve_data` that downloads the latest CVE bundle and re-runs `make_cve()` for every existing `CveRecord` in the database, backfilling their problem types
-  Extracted common download/bundle-extract/cache logic from `ingest_bulk_cve.py` into reusable functions
- prevent duplicate containers per CVE
- makes reingestion idempotent 
- existing container gets refreshed instead of getting skipped

### DIFF
--- a/src/shared/fetchers.py
+++ b/src/shared/fetchers.py
@@ -248,7 +248,7 @@ def make_container(
     problems = [
         desc
         for problem in data.get("problemTypes", [])
-        for desc in problem.get("description", [])
+        for desc in problem.get("descriptions", [])
     ]
     obj.problem_types.set(map(make_problem_type, problems))
     obj.references.set(map(make_reference, data.get("references", [])))


src/shared/models/cve.py
```diff
@@ -383,6 +383,12 @@ def __str__(self) -> str:
         return self.cve.cve_id
 
     class Meta:  # type: ignore[override]
+        constraints = [
+            models.UniqueConstraint(
+                fields=["cve", "_type"],
```
fricklerhandwerk: This doesn't seem right. The spec says there can me [multiple ADP containers](https://cveproject.github.io/cve-schema/schema/docs/#oneOf_i0_containers_adp)

src/shared/migrations/0097_unique_container_per_cve_type.py
```diff
@@ -0,0 +1,36 @@
+# Generated by Django 5.2.14 on 2026-07-01 14:49
+
+from django.db import migrations, models
+
+
+def deduplicate_containers(apps: object, schema_editor: object) -> None:
+    """Remove duplicate containers keeping the oldest (lowest id) per (cve, _type)."""
+    Container = apps.get_model("shared", "Container")
+    db_alias = schema_editor.connection.alias
```
fricklerhandwerk: I may be missing something here and would be happy to be enlightened why you did that, but I don't think we need that. What would it be for in this case?

src/shared/cve_download.py
fricklerhandwerk, line 1: Extracting the logic of dealing with those bundle files is a good idea! And a good preparation for making the feature. But then let's do it properly. Since all of this stuff is stateful, how about encapsulating it in an object? Then we can even make it such that the consumer doesn't need to care about implementation details, but simply do something like

```python
cve_archive = CVEarchive(settings.CVE_ARCHIVE_URL)
for cve_json in cve_archive.get(from=last_updated, to=date.today()).json_files:
   ...
```

I'm not super sure this extraction is absolutely worth doing, because what's going on is not *that* complicated. So it could just as well be distraction/procrastination for the actual thing. But it's not wrong, surely if done well it would be a reasonable QoL improvement for contributors doing changes in the future.

src/shared/cve_download.py
```diff
@@ -0,0 +1,71 @@
...
+        with zipfile.ZipFile(tmp_arc) as z_arc:
+            logger.info("Extracting the first archive to cves.zip")
+            z_arc.extractall(path=tmp_dir)
+
+        with zipfile.ZipFile(f"{tmp_dir}/cves.zip") as z_arc:
+            logger.info("Extracting the second archive to cves")
+            z_arc.extractall(path=data_cache_dir)
+
+
+def ensure_cve_cache(force_download: bool = False) -> str:
```
fricklerhandwerk: I'm not sure this directory handling is a worthy thing on its own. The original code is from the very early days and very ad hoc. Maybe it will essentially dissolve when refactoring the business logic into a class?

src/shared/management/commands/backfill_cve_data.py
```diff
@@ -0,0 +1,64 @@
...
+        parser.add_argument(
+            "--force-download",
+            action="store_true",
+            help="Ignore the local data cache and download the bundle again.",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        cve_data_cache_dir = ensure_cve_cache(force_download=options["force_download"])
+        cve_list = list_cve_json_files(cve_data_cache_dir)
+        existing_ids = set(models.CveRecord.objects.values_list("cve_id", flat=True))
```
fricklerhandwerk: That will be a rather long list (~350k). Not great, not terrible. But why not reuse the pattern from garbage collection where we slice the query at chunk size? Then there's no need to "report chunk-wise" since the chunking is already in place.

src/shared/management/commands/backfill_cve_data.py
```diff
@@ -0,0 +1,64 @@
...
+            cve_id = name.removesuffix(".json")
+
+            if cve_id not in existing_ids:
+                skipped += 1
+                continue
+
+            with open(j_cve) as fc:
+                data = json.load(fc)
+
+            record = models.CveRecord.objects.filter(cve_id=cve_id).first()
```
fricklerhandwerk: Hm, assuming we do have duplicates at that point, what do we do with the duplicate? Shouldn't we first ensure at the DB level that there are no duplicates?

fricklerhandwerk (commented):
I didn't get into much depth yet, also since this is still a draft, but took the time to check the general direction. It seems that you may be trying to do too many things at once:
- Fix the typo in the ingestion
- Refactor the handling of downloaded files
- Manage uniqueness/deduplication
- Figure out repeated ingestion

All of them are surely worth doing, but they're steeply increasing in difficulty. Especially to get deduplication right, one needs to do extensive testing to avoid breaking data that's referred to by user input (i.e. all the triaged stuff).

I suspect it would help to reduce the scope and do things in a deliberate sequence:
- Think about a testing strategy for ingestion

  Hand-crafted examples don't scale, they're extremely error-prone, and hard to verify in review. For example, what if we could somehow massage [Hypothesis](https://hypothesis.readthedocs.io/en/latest/) into generating test cases based on the [JSON schema](https://cveproject.github.io/cve-schema/)?
- Test the code paths we actually need

  That way, any subsequent change stays safe.

- Throw away ingestion code paths we're not using

  Would reduce the stuff to deal with *a lot*, and also save notable disk space.

- Throw away data we're not using, deduplicate messy relations

  This can be expected to be super tedious, because essentially it needs to be done relation by relation, starting at the leaves. Honestly, I recommend not touching that as a volunteer, this is exactly what people get paid for to do as a job. @adekoder did an extensive analysis a few months ago, this will take substantial work because one will run into a huge number of micro-decisions with potentially annoying consequences.
  
So for now, maybe just do the fix and if you're interested let's iterate on the refactoring. What will be very valuable  in any case (both for the project and as a learning experience) would be prototyping the property testing.

src/shared/models/cve.py
devnchill: Oh, I didn't realize there could be multiple ADP containers per CVE. I was  trying to prevent duplicate containers from building up on re-ingestion, but I see now that making `(cve, _type)` unique would break ADPs.

src/shared/migrations/0097_unique_container_per_cve_type.py
devnchill: coz I changed the `create` to `update_or_creat` and since at that time time I didn't realize CVE can have multiple ADP's at that i added that unique constrain ? so this would help in remove duplicates ? *although it's wrong now coz of multiple ADP for same cve id T_T*

src/shared/cve_download.py
devnchill: Ig i'll drop it for this pr due to scope issue , will do follow up prs to refactor it. the main reason behind it was due to backfill reusing the same logic from `bulk ingest`

src/shared/cve_download.py
devnchill: I see

src/shared/management/commands/backfill_cve_data.py
devnchill: yeah that sounds better

devnchill:
@fricklerhandwerk thanks for the detailed review , ig my thinking was: since I'm already touching the ingestion code to fix the typo, I might as well fix the duplicate containers problem I noticed , the old `create()` was piling up rows every re-ingestion. But I didn't realize ADP containers could repeat per CVE, so the unique constraint was wrong, and I see now that deduplication is a much deeper problem than I thought.

ig for this pr i'll drop the last commit and just fix typo and then if you are okay with it,can try to work on those points you mentioned :thinking: 

The property testing with Hypothesis against the CVE JSON schema sounds like a great learning opportunity ,I'd like to try prototyping that in a follow-up.

Does scaling it back to just the typo fix sounds okay?

fricklerhandwerk:
Yeah that's how it goes. After many years of doing that professionally, I still struggle to avoid scope explosion on a fix to some problem and at the same time manage the backlog of new problems discovered on the way...


The actual typo fix could indeed be a one-liner, as an exception even without a test, because testing essentially opens the big rabbit hole. Please document the fix against reference documentation in the commit message and leave a concise justification why it's not tested.

Looking forward to the property testing prototype! I'd be interested to work on that myself for the same reason, but I have other more pressing priorities so I'd be happy to review and discuss what you find.

devnchill:
understood :smile:

fricklerhandwerk (approved):

PR #1163, opened 2026-07-01
chore: replace all shell invocations with `exec`

fricklerhandwerk:
This prevents shell injection attacks with strings we take from upstream.
Those were theoretical anyway, since it would have required breaking the much more valuable target of core NixOS infrastructure.

The change has one advantage: Should the commands ever need to be changed by one argument, the diff will be very tidy.

Context: LLMs running in a loop are being pointed at this project. Hopefully this will quiet them down so we don't get even more redundant reports.

### DIFF
--- a/src/shared/git.py
+++ b/src/shared/git.py
@@ -68,14 +68,14 @@ class GitRepo:
 
     async def execute_git_command(
         self,
-        cmd: str,
+        *cmd: str,
         stdout: int | IO[Any] | None = None,
         stderr: int | IO[Any] | None = None,
     ) -> asyncio.subprocess.Process:
         final_stdout = stdout or self.stdout or asyncio.subprocess.PIPE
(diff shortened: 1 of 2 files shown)


Erethon (approved):
LGTM, thnx!

src/shared/git.py
```diff
@@ -91,14 +91,31 @@ async def clone(
...
         else:
-            clone_process = await asyncio.create_subprocess_shell(
-                f"git clone --depth=1 --bare --progress {repo_clone_url} {self.repo_path}",
+            clone_process = await asyncio.create_subprocess_exec(
+                *[
+                    "git",
+                    "clone",
+                    "--depth=1",
+                    "--bare",
+                    "--progress",
```
LeSuisse, line 115: Should not we also have a `--` section between the options and the parameters to be sure there is no confusion?

See https://snyk.io/blog/argument-injection-when-using-git-and-mercurial/ for some examples of argument injections

PR #1164, opened 2026-07-02
feat: create cve to package matching along side cve to drv matching

adekoder:
This PR introduces the groundwork to full migrate from cve - derivation matching to cve - package matching,
every time a proposal is matched, it also records which packages were involved, one row per distinct package. Derivation linking is left completely intact for now so nothing in the UI or API breaks.

what this PR is doing right now is 
- start recording package links alongside derivation links
- next step after this will be update views and consumers to read from package links instead, then  drop derivation linking entirely.

Notes: We need to run the `backfill_proposal_package_links` to create the package matching for already link package drvs


src/shared/tests/test_backfill_proposal_package_links.py
```diff
@@ -0,0 +1,122 @@
...
+    link = PackageClusterProposalLink.objects.get(proposal=suggestion)
+    assert link.package == pkg
+    assert link.provenance_flags == ProvenanceFlags.PACKAGE_NAME_MATCH
+
+
+def test_backfill_skips_unclustered_drv(
+    drv: NixDerivation,
+    suggestion: CVEDerivationClusterProposal,
+) -> None:
+    """Derivations with no PackageDerivation are silently skipped; no package link is created."""
```
fricklerhandwerk: I'd much prefer if we force-cluster the missing derivations during linking while we're at it. Otherwise when we switch to displaying the packages in a match, the results will be different.

src/shared/tests/test_linkage.py
```diff
@@ -284,3 +286,83 @@ def test_skip_known_vulnerability(
...
+    link = PackageClusterProposalLink.objects.get(proposal=proposal)
+    assert link.package == pkg
+    assert link.provenance_flags == ProvenanceFlags.PACKAGE_NAME_MATCH
+
+
+def test_unclustered_drv_produces_no_package_links(
+    make_container: Callable[..., Container],
+    make_drv: Callable[..., NixDerivation],
+) -> None:
+    """Derivations not yet assigned to a package are skipped without error."""
```
fricklerhandwerk, line 498: Same as above, let's just force-cluster them while matching.

src/shared/models/linkage.py
```diff
@@ -503,3 +504,29 @@ class DerivationClusterProposalLink(models.Model):
     # TODO: how to design the integrity here?
     # we probably want to add a fancy check here.
     provenance_flags = models.IntegerField()
+
+
+class PackageClusterProposalLink(models.Model):
+    """
+    A link between a Package and a CVEDerivationClusterProposal.
+    """
+
```
```suggestion
```

Not needed IMO

fricklerhandwerk (commented):
Sweet! I'd only go for immediate consistency here to avoid trouble down the road.

src/shared/tests/test_backfill_proposal_package_links.py
adekoder: but this process is handled by the post eval process, you want us to do that here as well ?

fricklerhandwerk: We can't tell how far the process is at that point, right? And the way this is built, once a suggestion is linked with packages, we won't touch it again, so there's no way to recover of its incomplete.

src/shared/management/commands/backfill_proposal_package_links.py
```diff
@@ -0,0 +1,63 @@
...
+                )
+            )
+            .distinct()
+        )
+        total = qs.count()
+        self.stdout.write(f"Backfilling package links for {total} proposals...")
+
+        created = 0
+        for proposal in qs.iterator(chunk_size=500):
+            package_flags: dict[int, int] = {}
```
fricklerhandwerk: Each suggestion should probably be handled atomically, since a background job may be killed mid-way by e.g. a deployment.

src/shared/tests/test_backfill_proposal_package_links.py
adekoder: Valid point... i will include it in the flows

fricklerhandwerk (approved):
Nice!

PR #1165, opened 2026-07-02
fix: Quadruple max_locks_per_transaction

### LINKED ISSUE #1153: Increase lock limit
We're getting

```log
out of shared memory
HINT: You might need to increase max_locks_per_transaction
```

once in a while. This is surely because of  the row lock introduced by https://github.com/NixOS/nix-security-tracker/pull/1035.

@Erethon what's a good setting?

Erethon:
The default for this setting is 64, let's quadruple it to allow for bigger batch sizes.

@fricklerhandwerk @adekoder did you do any fine-tuning on that 10k rows batch size when testing the PR? I believe lowering that would also help, but I don't know how it would impact speed.

Closes #1153

### DIFF
--- a/infra/production.nix
+++ b/infra/production.nix
@@ -78,6 +78,7 @@ in
         max_parallel_workers_per_gather = "4";
         max_parallel_workers = "8";
         max_parallel_maintenance_workers = "4";
+        max_locks_per_transaction = "256";
       };
       authentication = ''
         local all all trust
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -135,6 +135,14 @@ in
         PRODUCTION = cfg.production;
         STATIC_ROOT = "/var/lib/nix-security-tracker/static/"; # trailing slash is required!
         VITE_MANIFEST_PATH = "${cfg.frontend}/.vite/manifest.json";
+        PACKAGE_CLUSTERING_BATCH_SIZE =
+          let
+            parallelism = cfg.maxJobProcessors + 1; # account for periodic backfill
+            # fall back to implicit Postgres defaults
+            connections = cfg.services.postgresql.settings.max_connections or 100;
+            locks = cfg.services.postgresql.settings.max_locks_per_transaction or 64;
+          in
+          connections * locks / parallelism * 4 / 5; # add some margin for other transactions
         REVISION =
           (builtins.fetchGit {
             url = ../.;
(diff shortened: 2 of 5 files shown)


fricklerhandwerk:
I ran it with default Postgres settings locally a couple of times, without other concurrent things going on. Larger batch sizes (e.g. 20k, 50k) would error out on their own in the beginning when each batch would produce new packages/attributes. But I didn't do any rigorous measurement, so I don't even know how that 64 figures into a batch size of 10k.

infra/production.nix
```diff
@@ -71,6 +71,7 @@ in
         max_parallel_workers_per_gather = "4";
         max_parallel_workers = "8";
         max_parallel_maintenance_workers = "4";
+        max_locks_per_transaction = "256";
```
fricklerhandwerk, line 81: How about we move that into right into `configuration.nix` and say why it's need to be like that? I don't think this setting is in any way specific to any deployment.

fricklerhandwerk: Discussed with @Erethon:
- The number of advisory locks is <num connections> x <max_locks> (200x64 = 12400)
  - That explains why running batches locally didn't allow for more than 10k
- Bumping to 256 will give us  >50k slots, which should be plenty at that batch size
- We should move it to `configuration.nix` because it's an application-level requirement

fricklerhandwerk:
@Erethon I added a commit to avoid operators having to think about the setting if they don't want to, please take a look.

nix/configuration.nix
```diff
@@ -135,6 +135,14 @@ in
         PRODUCTION = cfg.production;
         STATIC_ROOT = "/var/lib/nix-security-tracker/static/"; # trailing slash is required!
         VITE_MANIFEST_PATH = "${cfg.frontend}/.vite/manifest.json";
+        PACKAGE_CLUSTERING_BATCH_SIZE =
+          let
+            parallelism = cfg.maxJobProcessors + 1; # account for periodic backfill
+            # fall back to implicit Postgres defaults
+            connections = cfg.services.postgres.settings.max_connections or 100;
```
Erethon: s/postgres/postgresql/?

nix/configuration.nix
```diff
@@ -135,6 +135,14 @@ in
...
+            locks = cfg.services.postgres.settings.max_locks_per_transaction or 64;
```
Erethon: s/postgres/postgresql/?

PR #1168, opened 2026-07-03
perf: only evaluate small channels

fricklerhandwerk:
Towards #864

Small channels are tracking their release branches most closely.
This will reduce evaluation turnaround time and triaging noise.

We don't need other channel states at the moment, since we're currently only communicating with package maintainers about the state of development.
The state of what's shipped is a future concern for communicating with package consumers.

The ideal solution for that is to evaluate only the tips of all release branches, but doing that cleanly and displaying the result faithfully requires more invasive changes.


src/shared/migrations/0097_nixchannel_variant.py
```diff
@@ -0,0 +1,38 @@
...
+def populate_variant(apps, schema_editor):
+    NixChannel = apps.get_model("shared", "NixChannel")
+    for channel in NixChannel.objects.all():
+        if channel.channel_branch.endswith("-small"):
+            channel.variant = "small"
+        elif channel.channel_branch.endswith("-darwin"):
+            channel.variant = "darwin"
+        elif channel.channel_branch.startswith("nixos-"):
+            channel.variant = "primary"
+        channel.save(update_fields=["variant"])
```
florentc, line 15: It seems to me we introduce redundancy in the model:
1. With no guarantee at the model level through constraint that both channel name suffix and variant field will keep matching after the migration for new channels. If I understand correctly it is done  in `channels.nix` for future channels right? It's only when new NixOS releases happen that we'll change it so the risk of introducing a mismatch then is unlikely? 
2. Couldn't we introduce a variant `@property` in the channel model that relies on checking the branch suffix? Would it be a big performance hit?

fricklerhandwerk: The suffix is not the authoritative source for the channel variant, it's indeed `channel.nix`. This migration is supposed to reflect that. But since we can't meaningfully look into the past, we're using the heuristic here for a last time. So I think doing it this way is actually right.

florentc: Makes sense :+1:

PR #1169, opened 2026-07-03
fix: regeneration of cache entries for non-matching CVEs

fricklerhandwerk:
This was made necessary after https://github.com/NixOS/nix-security-tracker/pull/1130 started picking up on previously discarded CVEs, that now would go through caching. Blocks testing #1141

### DIFF
--- a/src/shared/tests/test_suggestion_caching.py
+++ b/src/shared/tests/test_suggestion_caching.py
@@ -200,6 +200,24 @@ def test_ensure_fresh_cache_noop_when_fresh(
     )
 
 
+# FIXME(@fricklerhandwerk): [ref:invalid-cves] This tests a thing that shouldn't be possible in the first place; fix the underlying problem.
+def test_cache_skipped_without_title_or_description(
+    make_container: Callable[..., Container],
+) -> None:
+    container = make_container(
+        title=None, description=None, package_name=None, product=None
+    )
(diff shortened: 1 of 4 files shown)


src/shared/tests/test_suggestion_caching.py
```diff
@@ -200,6 +200,23 @@ def test_ensure_fresh_cache_noop_when_fresh(
     )
 
 
+def test_cache_skipped_without_title_or_description(
+    make_container: Callable[..., Container],
+) -> None:
+    container = make_container(
+        title=None, description=None, package_name=None, product=None
```
fricklerhandwerk, line 208: Interesting. How is this possible, since I actually made that field not-None?

fricklerhandwerk:
I'll merge this ~soon so we can resume deploying to production again.

src/webview/tests/test_issues.py
```diff
@@ -52,7 +53,11 @@ def test_publish_gh_issue_empty_title(
 
     container = make_container(title=title, description=description)
     accepted_suggestion = make_suggestion(
-        container=container, status=CVEDerivationClusterProposal.Status.ACCEPTED
+        drvs={
```
fricklerhandwerk, line 56: This is "related" because the changed code path is exercised by the test. I noticed that we're not using `drv` but probably should, as we do in similar setups.

PR #1172, opened 2026-07-07
feat: Suggestion details through API and new UI

florentc:
This brings suggestion details pages to the new UI (`ui-v2/suggestions/by-id/{suggestion_id}`).

It adds relevant endpoints in the API and tests.


src/api/suggestions/views.py
```diff
@@ -41,35 +51,82 @@ def to_representation(self, instance: CVEDerivationClusterProposal) -> dict:
...
 
-class SuggestionViewSet(viewsets.GenericViewSet):
+class SuggestionViewSet(RetrieveModelMixin, viewsets.GenericViewSet):
     queryset = CVEDerivationClusterProposal.objects.all()
-    permission_classes = [IsAuthenticated, CanEditSuggestion]
+    serializer_class = SuggestionSerializer
+
+    def get_permissions(self) -> list:
+        if (
+            getattr(self, "action", None) == "status"
```
fricklerhandwerk: Why do we need to check for the particular operation?

src/api/suggestions/views.py
```diff
@@ -41,35 +51,82 @@ def to_representation(self, instance: CVEDerivationClusterProposal) -> dict:
...
+            and getattr(self.request, "method", None) == "POST"
```
fricklerhandwerk:
```suggestion
            and getattr(self.request, "method", None) != "GET"
```

Since all mutations probably need auth...

src/api/suggestions/views.py
```diff
@@ -41,35 +51,82 @@ def to_representation(self, instance: CVEDerivationClusterProposal) -> dict:
...
+        ):
+            return [IsAuthenticated(), CanEditSuggestion()]
+        return super().get_permissions()
```
fricklerhandwerk: Do we really need that? `RetrieveModelMixin` does not seem to carry any permissions, no? I think it can just be

```
return getattr(self.request, "method", None) != "GET" and [IsAuthenticated(), CanEditSuggestion()]
```

src/api/suggestions/views.py
```diff
@@ -41,35 +51,82 @@ def to_representation(self, instance: CVEDerivationClusterProposal) -> dict:
...
+        methods=["get", "post"],
         url_path="status",
         serializer_class=SuggestionStatusSerializer,
-        permission_classes=[AllowAny],
     )
     def status(self, request: Request, pk: int) -> Response:
+        if request.method == "GET":
+            instance = self.get_object()
+            return Response(self.get_serializer(instance).data)
+        # POST: change status
```
fricklerhandwerk: Let's do this explicitly
```suggestion
        if request.method == "POST":
```
and fail on `else:`

src/api/suggestions/views.py
```diff
@@ -41,35 +51,82 @@ def to_representation(self, instance: CVEDerivationClusterProposal) -> dict:
...
+        serializer_class=ActivityLogEntrySerializer,
+        permission_classes=[AllowAny],
+    )
+    def activity_log(self, request: Request, pk: int) -> Response:
         instance = self.get_object()
-
-        instance.change_status(**serializer.validated_data)
-
-        return Response(self.get_serializer(instance).data)
+        raw_events = fetch_suggestion_events([instance.pk]).get(instance.pk, [])
```
fricklerhandwerk, line 131: We may want a FIXME here to fold all this stuff into the model

src/api/tests/test_suggestion_detail.py
```diff
@@ -0,0 +1,116 @@
...
+    assert "title" in data
+    assert "description" in data
+    assert "packages" in data
+    assert "original_packages" in data
+    assert "ignored_packages" in data
+    assert "metrics" in data
+    assert "affected_products" in data
+    assert "categorized_maintainers" in data
+    assert "categorized_url_references" in data
+    assert "in_issue_draft" in data
```
fricklerhandwerk:
```suggestion
    assert data["cve_id"] == suggestion.cve.cve_id
```

Maybe just leave that one for good measure, since you'll be testing the relevant fields in the future when you do actions

src/webview/tests/ui_v2/test_suggestion_detail.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_suggestion_detail_loads(
+    live_server: LiveServer,
+    page: Page,
+    cached_suggestion: CVEDerivationClusterProposal,
+) -> None:
+    """The suggestion detail page loads and shows the CVE ID."""
+    page.goto(live_server.url + SUGGESTION_DETAIL + f"/{cached_suggestion.pk}")
+    expect(page.get_by_text("CVE-2025-0001", exact=False)).to_be_visible()
```
fricklerhandwerk:
```suggestion
    expect(page.get_by_text(cached_suggestion.cve.cve_id, exact=False)).to_be_visible()
```

Otherwise we magically have to know it's right. Same in the following

src/webview/tests/ui_v2/test_suggestion_detail.py
```diff
@@ -0,0 +1,103 @@
...
+
+
+def test_suggestion_detail_shows_status(
```
fricklerhandwerk, line 22: The following ones amount to one big test... maybe consider putting them all into one.

src/api/suggestions/serializers.py
```diff
@@ -0,0 +1,279 @@
...
+from shared.models.linkage import CVEDerivationClusterProposal
+
+_CVSS_PARSERS = {
+    Metric.Format.V40: (CVSS4, constants4.METRICS_ABBREVIATIONS),
+    Metric.Format.V31: (CVSS3, constants3.METRICS_ABBREVIATIONS),
+    Metric.Format.V30: (CVSS3, constants3.METRICS_ABBREVIATIONS),
+}
+
+
+def _compute_cvss_fields(
```
fricklerhandwerk: This looks like a copy of the work we're doing in `severity_badge()` in `viewutils.py`. Please extract the relevant part and reuse it here.

src/api/suggestions/serializers.py
```diff
@@ -0,0 +1,279 @@
...
+    metric_dict: dict,
+) -> tuple[float | None, str | None, list[dict] | None]:
```
fricklerhandwerk: This should be a namedtuple maybe? And maybe alias the float/str to something that makes sense for a reader.

src/api/suggestions/serializers.py
```diff
@@ -0,0 +1,279 @@
...
+                    "label": f"{abbreviations[k]} ({k})",
+                    "value": f"{parsed.get_value_description(k)} ({v})",
+                }
+                for k, v in parsed.metrics.items()
+            ]
+            return float(score), severity.upper(), human_readable
+    return None, None, None
+
+
+class MetricHumanReadableItemSerializer(serializers.Serializer):
```
fricklerhandwerk, line 16: These names are really long. How about we just strip the `Serializer` suffix and access them qualified at the call sites?

src/api/suggestions/serializers.py
```diff
@@ -0,0 +1,279 @@
...
+class ActivityLogEntrySerializer(serializers.Serializer):
+    """A single activity log entry. The `action` field is the discriminator.
+
+    - "create": rejection_reason may be set (auto-dismissed at creation)
+    - "insert" | "update": status_value + optional rejection_reason
+    - "package.restore" | "package.ignore": package_names
+    - "maintainer.add" | "maintainer.delete" | "maintainer.ignore" | "maintainer.restore": maintainers
+    - "reference.ignore" | "reference.restore": references
+    """
+
```
fricklerhandwerk:
```suggestion
```

src/api/suggestions/serializers.py
```diff
@@ -0,0 +1,279 @@
...
+    timestamp = serializers.DateTimeField()
+    username = serializers.CharField(allow_null=True)
+    status_value = serializers.CharField(allow_null=True, default=None)
+    rejection_reason = serializers.CharField(allow_null=True, default=None)
+    package_names = serializers.ListField(child=serializers.CharField(), default=list)
+    maintainers = MaintainerSerializer(many=True, default=list)
+    references = ActivityLogReferenceSerializer(many=True, default=list)
+
+
+def folded_event_to_dict(event: FoldedEventType) -> dict:
```
fricklerhandwerk, line 221: As discussed, we should probably delegate the folding to the UI.

fricklerhandwerk (approved):
Cool, just some small nits! (reviewed together in sync)

fricklerhandwerk:

PR #1175, opened 2026-07-08
fix: don't collect replication stats

fricklerhandwerk:
Since the last Nixpkgs bump, `prometheus-postgres-exporter` has a buggy query that produces:

    pq: column \"slot_name\" does not exist

This change disables triggering the offending query.


Erethon (approved):
Tested locally with latest prometheus-postgres-exporter from Nixpkgs and option seems to exist, so lgmt.

PR #1178, opened 2026-07-09
feat: live editing comments in suggestions in new UI

florentc:
- API endpoints to get and edit the comment of a suggestion
- API tests
- Comment edition in new UI: no need to explicitly save, the edited comment is saved automatically with some debouncing and there is visual feedback (yellow: edited, green: saved, red: error)

### DIFF
--- a/frontend/src/components/suggestions/Comment.tsx
+++ b/frontend/src/components/suggestions/Comment.tsx
@@ -1,11 +1,70 @@
+import { useEffect, useRef, useState } from "preact/hooks";
+import { useCommentMutation } from "@/hooks/useComment";
+import styles from "./Comment.module.css";
+
+type SaveState = "idle" | "pending" | "saving" | "saved" | "error";
+
 type Props = {
-  comment: string;
+  suggestionId: number;
+  comment: string | null;
+  canEdit: boolean;
 };
 
-export function Comment({ comment }: Props) {
+const DEBOUNCE_SAVE_MS = 500;
+const DEBOUNCE_CLEAR_SAVED_FEEDBACK_MS = 2000;
+
+export function Comment({ suggestionId, comment, canEdit }: Props) {
+  const [value, setValue] = useState(comment ?? "");
+  const [saveState, setSaveState] = useState<SaveState>("idle");
+  const debounceRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const savedTimeoutRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
+  const mutation = useCommentMutation(suggestionId);
+
+  // Sync if the prop changes externally (e.g. parent re-fetches)
+  useEffect(() => {
+    setValue(comment ?? "");
+  }, [comment]);
+
+  function handleChange(e: Event) {
+    const next = (e.target as HTMLTextAreaElement).value;
+    setValue(next);
+    setSaveState("pending");
+
+    if (debounceRef.current) clearTimeout(debounceRef.current);
+    if (savedTimeoutRef.current) clearTimeout(savedTimeoutRef.current);
+
+    debounceRef.current = setTimeout(() => {
+      setSaveState("saving");
+      mutation.mutate(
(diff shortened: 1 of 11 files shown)


src/api/suggestions/serializers.py
```diff
@@ -13,6 +13,17 @@
...
 
+class SuggestionCommentSerializer(serializers.Serializer):
+    """Serializer for reading or updating a suggestion comment."""
+
+    comment = serializers.CharField(
+        allow_null=True,
+        allow_blank=True,
+        required=True,
+        help_text="Free-text comment. Set to empty string to clear.",
+    )
```
DarshanCode2005, line 25: the model also has max_length , so overlong comments may pass validation , similarly the limit could be added on frontend as well

frontend/src/components/suggestions/Comment.tsx
```diff
@@ -1,11 +1,63 @@
...
+          onSuccess: () => {
+            setSaveState("saved");
+            savedTimeoutRef.current = setTimeout(() => setSaveState("idle"), 2000);
+          },
+          onError: () => {
+            setSaveState("error");
+          },
+        },
+      );
+    }, 500);
```
fricklerhandwerk: Let's make it a named constant to make it easier to find

PS: It's perfectly okay to have it hard coded, because for us changing a config value is exactly the same as changing a constant.

frontend/src/components/suggestions/Comment.tsx
```diff
@@ -1,11 +1,63 @@
...
+    if (savedTimeoutRef.current) clearTimeout(savedTimeoutRef.current);
+
+    debounceRef.current = setTimeout(() => {
+      setSaveState("saving");
+      mutation.mutate(
+        { id: suggestionId, data: { comment: next } },
+        {
+          onSuccess: () => {
+            setSaveState("saved");
+            savedTimeoutRef.current = setTimeout(() => setSaveState("idle"), 2000);
```
fricklerhandwerk: Same, please name it

frontend/src/components/suggestions/Comment.tsx
```diff
@@ -1,11 +1,63 @@
...
+          },
+          onError: () => {
+            setSaveState("error");
```
fricklerhandwerk, line 50: 🤔 Intuitively it would be good to know what's going on, but I guess the UI won't allow most invalid states, and if the backend is down people will notice in other ways.

frontend/src/components/suggestions/Suggestion.tsx
```diff
@@ -30,6 +31,9 @@ export function Suggestion({ suggestion }: Props) {
     categorized_url_references,
   } = suggestion;
 
+  const { user } = useAuth();
+  const canEdit = !!(user?.is_committer || user?.is_admin);
```
fricklerhandwerk: Double negation `??`


```suggestion
  const canEdit = user?.is_committer || user?.is_admin;
```

src/api/suggestions/permissions.py
fricklerhandwerk, line 1: Let's put it back, no point having it in a separate file right now

src/api/tests/test_suggestion_comment.py
```diff
@@ -0,0 +1,100 @@
...
+
+def test_patch_comment_clears_with_empty_string(
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+    committer: User,
+) -> None:
+    suggestion = make_cached_suggestion(comment="foo")
+    client = APIClient()
+    client.force_login(committer)
+    response = client.patch(url(suggestion.pk), {"comment": ""}, format="json")
+    assert response.status_code == 200
```
fricklerhandwerk, line 85: Noted by @florentc in review: Maybe we don't need to return the value on PATCH.

Yeah sure. IMO it doesn't matter.

fricklerhandwerk (approved):
Nice, some good improvements to overall sanity there

fricklerhandwerk:

PR #1183, opened 2026-07-12
feat: collect performance metrics for garbage collection and CVE delta ingest

DarshanCode2005:
### Problem

Oneshot batch jobs besides cache regeneration only log durations and counts. Operators have no Grafana view of whether garbage collection or CVE delta ingest are slowing down or processing less data over time.

This PR reuses the textfile metrics wiring from #1141 for 2 more jobs.

Pipeline (Same):

job -> *.prom -> node_exporter textfile collector -> Prometheus -> Grafana

Related to #1002 . Follows #1141

### DIFF
--- a/contrib/grafana-dashboard.json
+++ b/contrib/grafana-dashboard.json
@@ -1292,13 +1292,13 @@
     "list": [
       {
         "current": {},
-        "definition": "label_values(sql_delta,instance)",
+        "definition": "label_values(up, instance)",
         "label": "Instance",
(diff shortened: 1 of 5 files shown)


DarshanCode2005:
image

fricklerhandwerk:
Please let's make these charts so we can see what's happening over time. If it doesn't work in a small-ish change, please add one metric as a chart in a separate PR and then we can repurpose this one to repeat the pattern for the other metrics.

PR #1184, opened 2026-07-12
feat: expose worker Prometheus metrics for matching and eval ingest

DarshanCode2005:
### Description

Related to #1002. Follows #1141 and #1183.

### Problem

Oneshot batch jobs already export last-run gauges through the textfile collector (`job="node"`). Long-running listen workers (matching and evaluation ingest) only log durations, so operators have no live Prometheus/Grafana visibility into these hot paths.

### What this PR does

Adds live `prometheus_client` HTTP `/metrics` endpoints for listen workers and instruments the key execution paths.

Exports:

* `sectracker_matching_duration_seconds`
* `sectracker_matching_candidates`
* `sectracker_nix_evaluation_batch_ingest_duration_seconds`
* `sectracker_nix_evaluation_batch_ingest_attributes`

Also:

* Adds `METRICS_HTTP_PORT`; metrics server starts only in `listen --worker` processes
* Exposes worker metrics on `:9252` and evaluator multiprocess metrics on `:9253`
* Adds firewall ruls and Grafana panels under **Application metrics** using the `sectracker-worker` and `sectracker-evaluator` jobs

### DIFF
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -181,9 +181,27 @@ in
       type = types.int;
       default = 2;
     };
+    metricsWorkerPort = mkOption {
+      description = ''
+        TCP port for the main listen worker Prometheus /metrics endpoint (matching).
+      '';
+      type = types.port;
+      default = 9252;
+    };
+    metricsEvaluatorPort = mkOption {
+      description = ''
+        TCP port for the evaluator multiprocess Prometheus /metrics sidecar.
+      '';
+      type = types.port;
+      default = 9253;
+    };
   };
 
   config = mkIf cfg.enable {
+    networking.firewall.allowedTCPPorts = [
+      cfg.metricsWorkerPort
+      cfg.metricsEvaluatorPort
+    ];
     environment.systemPackages = [ wstExternalManageScript ];
     services = {
       # TODO(@fricklerhandwerk): move all configuration over to pydantic-settings
@@ -332,7 +350,14 @@ in
           ];
           wantedBy = [ "multi-user.target" ];
 
+          environment = {
+            PROMETHEUS_MULTIPROC_DIR = "/var/lib/nix-security-tracker/prometheus-multiproc";
+          };
+
           script = ''
+            mkdir -p "$PROMETHEUS_MULTIPROC_DIR"
+            # prometheus_client multiprocess mode requires an empty directory at start.
+            find "$PROMETHEUS_MULTIPROC_DIR" -mindepth 1 -delete
             # Before starting, crash all the in-progress evaluations.
             # This will prevent them from being stalled forever, since workers would not pick up evaluations marked as in-progress.
             wst-manage crash_all_evaluations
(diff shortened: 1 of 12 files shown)


nix/configuration.nix
```diff
@@ -343,6 +368,31 @@ in
...
+              cfg.settings
+              // {
+                METRICS_HTTP_PORT = cfg.metricsEvaluatorPort;
+              }
+            );
+          };
+
+          script = ''
+            mkdir -p "$PROMETHEUS_MULTIPROC_DIR"
+            wst-manage serve_worker_metrics
```
fricklerhandwerk, line 392: My hunch is this could be an ad hoc more-or-less-inline  Python HTTP server here, which we put behind the proxy. Seems kind of wrong to weave this into the application code.

src/shared/metrics.py
```diff
@@ -14,3 +48,43 @@ def write_metrics_textfile(name: str, registry: CollectorRegistry) -> None:
...
+def observe_matching(duration_seconds: float, candidates: int) -> None:
+    """Record matching wall time and candidate cardinality."""
+    _matching_duration.observe(duration_seconds)
+    _matching_candidates.observe(float(candidates))
+
+
+def observe_eval_batch_ingest(duration_seconds: float, attributes: int) -> None:
+    """Record eval batch ingest wall time and batch size."""
+    _eval_batch_duration.observe(duration_seconds)
+    _eval_batch_attributes.observe(float(attributes))
```
fricklerhandwerk, line 62: These and their subroutines are used exactly once, they may as well be done in place. Rule of thumb: if there's no more than 2 repetitions, extracting (especially into a different place) almost never makes sense. Uncle-Bob-ing a bit for readability is not wrong, but putting into a separate file and wrapping a constructor in a procedure brings no benefit here.

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,22 +96,84 @@ def handle(self, *args: Any, **options: Any) -> None:
...
         # `pghistory` events are never auto-deleted — each step explicitly clears relevant events first.
 
+        deleted: dict[str, int] = {kind: 0 for kind in DELETED_KINDS}
+        step_durations: dict[str, float] = {}
+        start_total = time.time()
+
         self.stdout.write("\n[1/6] Deleting stale matches")
-        self._delete_stale_matches(cutoff, batch_size, dry_run)
+        step_start = time.time()
+        for kind, count in self._delete_stale_matches(
```
fricklerhandwerk, line 105: Why not have the helper return its own duration? Then you get rid of the boilerplate at the call site.

And maybe you could also collect the returned dicts `kind: count` into a list and just sum them up at the end in one go? Then you don't need `DELETED_KINDS` at all, which confused me quite a bit until I understood what you're trying to do.

fricklerhandwerk (commented):
Metrics server should be extracted properly. Haven't looked at the dashboards and tests yet.

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: Oops, this one belongs to #1183

nix/configuration.nix
DarshanCode2005: we need the sidecar because evaluator workers use multiprocess mode: each process must serve `MultiProcessCollector`. Putting that in `wst-manage` keeps Django settings/`METRICS_HTTP_PORT` the same as the worker path. An inline Python one-lineer in nix is possible, but it would still need the same `prometheus_client` + multiproc wiring, and would drift from how the worker exposes metrics. Happy to extract a thinner entrypoint if you want it out of Django management commands , but "serve aggregated multiproc metrics" still has to live somewhere next to the app that writes those files.

src/shared/metrics.py
DarshanCode2005: Okay

nix/configuration.nix
fricklerhandwerk: I'm not 100% sure I understand the problem: do you mean you want to prevent config drift between the two? That is a non-issue, since that is exactly what NixOS is for. As with the suggestion in an earlier PR, you can grab that setting from `config.services.nix-security-tracker` under whichever attribute you've put it and assign that service the right directory and permissons. And also since it seems this is just serving a file, why not simply tack it under a nginx virtualhost?

PR #1195, opened 2026-07-14
feat: Add a three days cooldown on NPM packages

Erethon:
Following up on #1194, let's add a cooldown when updating node packages.


frontend/.npmrc
```diff
@@ -0,0 +1 @@
+min-release-age=3
```
fricklerhandwerk, line 4: Can we document in place why that's there?

fricklerhandwerk (approved):

PR #1198, opened 2026-07-15
feat: ignore/restore references in API and new UI

florentc:
- API endpoints to ignore/restore suggestion references
    - With model-level methods to ignore/restore
    - With tests
- Implementation in new UI
    - Fits legacy UI design
    - Snappy optimistic updates when clicking on the buttons
    - Future proofing
        - Auto update activity log
        - Live update of relative times in activity log after mutation
        - Popup errors mechanism for unexpected API or network errors (or other on-the-spot messages)
        - Automatic refresh of the suggestion when a stale-state is suspected from a 4xx API error (which is not possible when the suggestion does show the current state): this can happen if another user edits the suggestion at the same time as you, or you have several tabs open, or both legacy and new UI
    - Tests

### DIFF
--- a/frontend/src/components/suggestions/ActivityLog.tsx
+++ b/frontend/src/components/suggestions/ActivityLog.tsx
@@ -11,8 +11,16 @@ import {
 } from "lucide-preact";
 import { useGetSuggestionActivityLog } from "@/api/generated/endpoints";
 import { type ActivityLogEntry, StatusEnum } from "@/api/generated/models";
+import { Skeleton } from "@/components/ui/Skeleton";
+import { Spinner } from "@/components/ui/Spinner";
+import { useTick } from "@/hooks/useTick";
 import { formatTime } from "@/utils/date";
 import { SuggestionStatusIcon } from "./SuggestionStatusIcon";
+
+// Re-render frequency for relative-time displays updates
+const TICK_INTERVAL_MS = 10_000;
+// Time during which a recent update is shown as "just now"
+const JUST_NOW_MS = 5_000;
 
 type Props = {
   suggestionId: number;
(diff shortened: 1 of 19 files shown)


frontend/src/components/suggestions/ActivityLog.tsx
```diff
@@ -158,31 +166,44 @@ function entryDescription(entry: ActivityLogEntry) {
...
+  if (!data || data.length === 0) return null;
 
   const last = data[data.length - 1];
   const summaryVerb = last.action === "create" ? "created" : "updated";
 
   return (
-    <details className="column gap-small align-end">
+    <details
+      className="column gap-small align-end"
+      data-testid={`suggestion-${suggestionId}-activity-log`}
```
fricklerhandwerk, line 203: No strong opinion, but we may as well leave it as `id`... I know it's the idiomatic thing to do with Playwright, but we're not stripping those in production, so users may as well use them for anchors, right?

frontend/src/hooks/useReference.ts
```diff
@@ -0,0 +1,68 @@
...
+              [toKey]: [...refs[toKey], moving],
+            },
+          };
+        });
+
+        return { previous };
+      },
+      onError: (err: unknown, vars: MutationVars, context?: MutationContext) => {
+        if (err instanceof ApiError && err.status === 400) {
+          // Likely a stale-state conflict (e.g. another user already ignored/restored this reference).
```
fricklerhandwerk, line 49: As @florentc suggested in review, we *may* demote the popup to a warning here, because it's not a hard error but simply a state reset, but no strong opinion from either of us there.

frontend/src/components/suggestions/ActivityLog.tsx
```diff
@@ -11,9 +11,17 @@ import {
...
+import { Skeleton } from "@/components/ui/Skeleton";
+import { Spinner } from "@/components/ui/Spinner";
+import { useTick } from "@/hooks/useTick";
 import { formatTime } from "@/utils/date";
 import { SuggestionStatusIcon } from "./SuggestionStatusIcon";
 
+// Re-render frequency for relative-time displays updates
+const TICK_INTERVAL_MS = 1_000;
+// Time during which a recent update is shown as "just now"
+const JUST_NOW_MS = 5_000;
```
fricklerhandwerk:
```suggestion
const TICK_INTERVAL_MS = 60_000;
// Time during which a recent update is shown as "just now"
const JUST_NOW_MS = 60_000;
```

As discussed

fricklerhandwerk (approved):

fricklerhandwerk:

PR #1201, opened 2026-07-16
Add a pgbouncer to be used by the ASGI server only

### LINKED ISSUE #1019: Connection pool
We keep running into exhausting the number of database connections, which requires restarting the service manually once in a while. Set up a connection pool to prevent that from happening.

Erethon:
Adds a pgbouncer service (optionally) to the host configuration and configures it so that only the ASGI server goes through it. Worker processes and management commands will use the direct database connection.

I've split this in two commits, because the autoformatting created a big diff and I hope this makes it easier to review.

Closes #1019

### DIFF
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -240,14 +240,15 @@ in
           pgbouncer = {
             pool_mode = "transaction";
             auth_type = "trust";
-            auth_file = toString (pkgs.writeText "pgbouncer-userlist" ''
-              "nix-security-tracker" ""
-            '');
+            auth_file = toString (
+              pkgs.writeText "pgbouncer-userlist" ''
+                "nix-security-tracker" ""
+              ''
+            );
             ignore_startup_parameters = "extra_float_digits";
           };
           databases = {
-            nix-security-tracker =
-              "host=/run/postgresql dbname=nix-security-tracker";
+            nix-security-tracker = "host=/run/postgresql dbname=nix-security-tracker";
           };
         };
       };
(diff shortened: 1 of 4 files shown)


nix/configuration.nix
```diff
@@ -181,6 +197,33 @@ in
...
+        default = 500;
+      };
+      reservePoolSize = mkOption {
+        type = types.int;
+        default = 10;
+      };
+      maxPreparedStatements = mkOption {
+        type = types.int;
+        default = 1000;
+      };
```
fricklerhandwerk: Let's not add these; we can configure pgbouncer directly when needed

nix/configuration.nix
```diff
@@ -59,6 +61,20 @@ let
...
+  # The ASGI server should not hold Django-level persistent connections when
+  # PgBouncer multiplexes server connections. Setting CONN_MAX_AGE=0 lets
+  # PgBouncer do the pooling. Other services (workers and management commands)
+  # keep the value from `cfg.settings` so their direct connections persist.
+  serverDjangoSettingsJson = builtins.toJSON (
+    cfg.settings // { DATABASE_CONN_MAX_AGE = 0; }
+  );
+  pgbouncerAuthFile = pkgs.writeText "pgbouncer-userlist" ''
+    "nix-security-tracker" ""
+  '';
```
fricklerhandwerk: Let's move this inline to where it's set, otherwise it's hard to read out of context

nix/configuration.nix
```diff
@@ -300,6 +331,19 @@ in
             ''
               daphne ${networking} project.asgi:application
             '';
+        }
+        // optionalAttrs cfg.enablePgbouncer {
+          # When PgBouncer is enabled, the ASGI server connects through it over
+          # its unix socket on port 6432
+          environment.DATABASE_URL = "postgres:///nix-security-tracker?host=/run/pgbouncer&port=6432";
```
fricklerhandwerk:
```suggestion
          environment.DATABASE_URL = let
            port = config.services.pgbouncer.settings.pgbouncer.listen_port;
          in
          "postgres:///nix-security-tracker?host=/run/pgbouncer&port=${toString port}";
```

fricklerhandwerk (approved):
Small nit. Any reason not to run it in production directly? We'll deploy it to staging first anyway.

Erethon:
> Small nit. Any reason not to run it in production directly? We'll deploy it to staging first anyway.

Just wanted to make sure it doesn't get deployed by accident, but I've enabled it now for production too. We should just be careful once it's merged that it's tested in staging first.

Issue #1038, opened 2026-05-18
Endpoint for training data on matching


fricklerhandwerk:
Discussed the design in https://github.com/NixOS/nix-security-tracker/issues/722#issuecomment-4350963285

In order to check out of band whether a new matching algorithm will perform better than the existing one, we need a way to obtain existing data that's relevant for matching quality. This is essentially all data that was touched by users having something to do with matching. The out of band measurement would obtain (and possibly cache) the data from the endpoint to be implemented for this issue.

Eventually (probably not for this issue) we also want the measurement to run in CI (manually! it's expensive) as an optional guard to ensure that the matching quality is indeed improved by a change.

DarshanCode2005:
Design Blurb for the issue:
**Problem:** We can deploy new matching algorithm versions (#1018), but there is no systematic, out-of-band way to check whether a candidate performs better. Acceptance rate in Grafana is a lagging proxy, not a reproducible benchmark.

So , if production has CVE-derivation proposals that security team members have accepted, rejected, published, or corrected via overlays, so a maintainer runs a local benchmark command against production, following this they can obtain a bounded dataset (CVE inputs+ linked derivations and user labels) which I suppose is sufficient to re-run a candidate matcher offline and compute false-postive rate without copying full channel evals or writing to production

**Hypothesis:** Exporting only derivations already linked to user-curated proposals (plus CVE matching inputs and overlay labels) is enough to measure false-positive reduction. Full historical eval replay is infeasible; false negatives wait for #213 .

**Proposed v0 (this issue):**

1. Read-only API endpoint, maintainer-authenticated
2. Filter: user-curated proposals only
3. Payload per record: CVE matching inputs, proposal status/labels, linked derivations (attribute, name, system, provenance_flags, minimal metadata), overlays
4. Omit eval commit unless we find we need it during implementation
5. DRF default pagination

**Metric (offline tool, not the endpoint):** on accepted proposals, candidate should keep non-overlayed links; on rejected/overlay-filtered, should not re-propose removed derivations -> FP count, kept count, SNR.

**Out of scope for v0:** rate limiting, bulk downtime dumps, CI, in-prod A/B, false negatives (#213), full eval export.

**Open questions** (from [#722 discussion](https://github.com/NixOS/nix-security-tracker/issues/722#issuecomment-4350963285)):

1. After fetching pages from the API, how do we ingest locally to benchmark against a fork? Management command vs filtered postgres dump of specific tables?
[...]

PR #1203, opened 2026-07-20
feat: matching training data schema and roundtrip

DarshanCode2005:
### Summary

- Add a versioned shared schema (`schema_version=1`) for matching training data (#1038)
- Export user-curated proposals (`status != pending`, including auto-rejects) to that schema
- Import records onto a synthetic `benchmark` channel so the matcher can run locally
- Roundtrip tests: export -> import -> re-export equality + `resolve_linkage_candidates` test

### Context

Staging measurement showed ~4k curated proposals / ~12k links (~8-12 MB JSON) vs a 179 GB DB. Bounded JSON + shared export/import schema is the agreed approach; `pg_dump` is out of scope here.

### DIFF
--- a/src/shared/evaluation.py
+++ b/src/shared/evaluation.py
@@ -3,7 +3,7 @@ import logging
 import time
 from collections.abc import Callable, Generator
 from dataclasses import dataclass, field
-from typing import Any, TypeVar
+from typing import Any, NamedTuple, TypeVar
 
 from dataclass_wizard import JSONWizard, LoadMixin
 from django.db.models import Model
@@ -60,6 +60,13 @@ class MetadataAttribute(JSONWizard, LoadMixin):
     known_vulnerabilities: list[str] = field(default_factory=list)
 
 
+class DerivationKey(NamedTuple):
+    drv_path: str
+    attr: str
+    name: str
+    meta_name: str | None
+
+
 @dataclass
 class EvaluatedAttribute(JSONWizard):
     """
@@ -74,27 +81,44 @@ class EvaluatedAttribute(JSONWizard):
     outputs: dict[str, str]
     system: str
 
-    def as_key(self) -> tuple[str, str, str, str | None]:
-        """
-        Unique dictionary key for a derivation
-
-        These are the actual degrees of freedom for a derivation, judging from the data.
-        """
-        # FIXME(@fricklerhandwerk): We should only need the derivation path!
-        # Extract the extra fields to save more space.
-        # A `NixPackage` could indeed consist of just `pname` (parsed from `name`, validate against `attribute` and `drv_metadata.name`).
-        # Then we'd describe all occurrences of a `NixPackage` with
-        # - NixDerivation
-        # - attribute_name
-        # - metadata__name
-        # - parent_evaluation (also extracted since derivation paths of close-to-root packages can be the same across evaluations)
-        # - version (also parsed from `name`, for easier querying)
-        return (
-            self.drv_path,
-            self.attr,
-            self.name,
-            self.meta.name or None if self.meta else None,
+    def as_key(self) -> DerivationKey:
+        """Unique dictionary key for a derivation (see :func:`derivation_as_key`)."""
+        return derivation_as_key(
+            drv_path=self.drv_path,
+            attr=self.attr,
+            name=self.name,
+            meta_name=self.meta.name or None if self.meta else None,
         )
+
+
+def derivation_as_key(
+    *,
+    drv_path: str,
+    attr: str,
+    name: str,
+    meta_name: str | None,
+) -> DerivationKey:
+    """
+    Unique dictionary key for a derivation.
+
+    These are the actual degrees of freedom for a derivation, judging from the data.
+    Shared by evaluation ingestion and matching training-data import.
+    """
+    # FIXME(@fricklerhandwerk): We should only need the derivation path!
+    # Extract the extra fields to save more space.
+    # A `NixPackage` could indeed consist of just `pname` (parsed from `name`, validate against `attribute` and `drv_metadata.name`).
+    # Then we'd describe all occurrences of a `NixPackage` with
+    # - NixDerivation
+    # - attribute_name
+    # - metadata__name
+    # - parent_evaluation (also extracted since derivation paths of close-to-root packages can be the same across evaluations)
+    # - version (also parsed from `name`, for easier querying)
+    return DerivationKey(
+        drv_path=drv_path,
+        attr=attr,
+        name=name,
+        meta_name=meta_name,
+    )
 
 
 @dataclass
@@ -159,7 +183,7 @@ def parse_evaluation_result(line: str) -> PartialEvaluatedAttribute:
 
 def by_drv_key[T](
     gen: Generator[tuple[EvaluatedAttribute, list[T]]],
-) -> dict[tuple[str, str, str, str | None], list[T]]:
+) -> dict[DerivationKey, list[T]]:
     return dict((origin.as_key(), elements) for origin, elements in gen)
 
 
@@ -288,7 +312,7 @@ class SyncBatchAttributeIngester:
 
     def ingest(self) -> list[NixDerivation]:
         start = time.time()
-        bulk_derivations: dict[tuple[str, str, str, str | None], NixDerivation] = {}
+        bulk_derivations: dict[DerivationKey, NixDerivation] = {}
         bulk_maintainers: dict[int, NixMaintainer] = {}
         bulk_licenses: dict[str, NixLicense] = {}
         metadata = []
--- a/src/shared/matching_training_data/serializers.py
+++ b/src/shared/matching_training_data/serializers.py
@@ -1,0 +1,418 @@
+"""DRF ModelSerializers for matching training-data roundtrip.
+
+The exported derivation and CVE data is exactly what we assume is required for
+matching them reproducibly.
+Adapt that appropriately when the matching algorithm changes.
+
+Serializers are named after the models they read/write. Export with
+``CVEDerivationClusterProposal(instance).data``; import with
+``CVEDerivationClusterProposal(data=...).is_valid(); .save()``.
+"""
+
+from __future__ import annotations
+
+import uuid
+from typing import Any, cast
+
+from django.db import transaction
+from rest_framework import serializers
+
+from shared import models
+from shared.evaluation import DerivationKey, derivation_as_key
+
+SCHEMA_VERSION = 1
+
+# Synthetic channel used when materializing a local benchmark corpus
+BENCHMARK_CHANNEL_BRANCH = "benchmark"
+BENCHMARK_RELEASE_BRANCH = "benchmark-master"
+
+# Fixed dummy git SHA so get_or_create stays obviously idempotent.
+_BENCHMARK_DUMMY_SHA1 = "0" * 40
+
+# Stable org for imported training CVEs
+_TRAINING_ORG_UUID = uuid.UUID("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
+
+
+def ensure_benchmark_evaluation() -> models.NixEvaluation:
+    """Create or reuse the synthetic benchmark channel + completed evaluation."""
+    release_branch, _ = models.NixpkgsBranch.objects.get_or_create(
+        name=BENCHMARK_RELEASE_BRANCH,
+        defaults={"head_sha1_commit": _BENCHMARK_DUMMY_SHA1},
+    )
+    channel, _ = models.NixChannel.objects.get_or_create(
+        channel_branch=BENCHMARK_CHANNEL_BRANCH,
+        defaults={
+            "release_branch": release_branch,
+            "state": models.NixChannel.ChannelState.UNSTABLE,
+            "head_sha1_commit": _BENCHMARK_DUMMY_SHA1,
+            "variant": None,
+        },
+    )
+    evaluation = (
+        models.NixEvaluation.objects.filter(
+            channel=channel,
+            state=models.NixEvaluation.EvaluationState.COMPLETED,
+        )
+        .order_by("-updated_at")
+        .first()
+    )
+    if evaluation is None:
+        evaluation = models.NixEvaluation.objects.create(
+            channel=channel,
+            commit_sha1=_BENCHMARK_DUMMY_SHA1,
+            state=models.NixEvaluation.EvaluationState.COMPLETED,
+        )
+    return evaluation
+
+
+def pick_first_container(
+    proposal: models.CVEDerivationClusterProposal,
+) -> models.Container | None:
+    """Prefer a container with ``package_name`` (needed for matching); else any first container."""
+    return (
+        proposal.cve.container.filter(affected__package_name__isnull=False).first()
+        or proposal.cve.container.first()
+    )
+
+
+def _training_drv_path(*, attribute: str, name: str, system: str) -> str:
+    """Stable synthetic drv path so :func:`derivation_as_key` stays deterministic."""
+    return f"/nix/store/training-{attribute}-{name}-{system}.drv"
+
+
+class AffectedProduct(serializers.ModelSerializer):
+    cpes = serializers.ListField(child=serializers.CharField(), required=False)
+
+    class Meta:
+        model = models.AffectedProduct
+        fields = (
+            "vendor",
+            "product",
+            "package_name",
+            "cpes",
+        )
+
+    def to_representation(self, instance: models.AffectedProduct) -> dict[str, Any]:
+        return {
+            "vendor": instance.vendor,
+            "product": instance.product,
+            "package_name": instance.package_name,
+            "cpes": sorted(cpe.name for cpe in instance.cpes.all()),
+        }
+
+
+class Container(serializers.Serializer):
+    """
+    Subset of Container fields relevant for training data
+    """
+
+    tags = serializers.ListField(child=serializers.CharField(), required=False)
+    affected = AffectedProduct(many=True, required=False)
+
+    def to_representation(self, instance: models.Container) -> dict[str, Any]:
+        affected = [
+            AffectedProduct().to_representation(product)
+            for product in instance.affected.all()
+        ]
+        return {
+            "tags": sorted(tag.value for tag in instance.tags.all()),
+            "affected": sorted(
+                affected,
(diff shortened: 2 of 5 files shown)


src/shared/matching_training_data/schema.py
fricklerhandwerk, line 1: IMO all of this should be done in [DRF serializers](https://www.django-rest-framework.org/api-guide/serializers/), which should be round-tripping. We have a bunch in `/src/api`.

DarshanCode2005: Okay , I will push an update to this , and rebase #1204

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+    known_vulnerabilities = serializers.ListField(
+        child=serializers.CharField(), required=False
+    )
+    provenance_flags = serializers.IntegerField()
+    was_linked = serializers.BooleanField()
+
+
+class PackageOverlayDataSerializer(serializers.Serializer):
+    package_attribute = serializers.CharField()
+    overlay_type = serializers.CharField()
```
fricklerhandwerk: We can use the enum from the model here

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+class MaintainerOverlayDataSerializer(serializers.Serializer):
+    github_id = serializers.IntegerField()
+    github = serializers.CharField()
+    overlay_type = serializers.CharField()
+
+
+class ReferenceOverlayDataSerializer(serializers.Serializer):
+    reference_url = serializers.CharField()
+    overlay_type = serializers.CharField()
+    deduplicated_name = serializers.CharField(required=False, allow_blank=True)
```
DarshanCode2005: Doesn't affect the matching

src/shared/matching_training_data/schema.py
```diff
@@ -0,0 +1,159 @@
...
+    reference_overlays = [
+        {
+            "reference_url": item["reference_url"],
+            "overlay_type": item["overlay_type"],
+            "deduplicated_name": item.get("deduplicated_name") or "",
+        }
+        for item in (labels.get("reference_overlays") or [])
+    ]
+
+    derivations = [
```
DarshanCode2005: Remove the schema.py

src/shared/matching_training_data/importer.py
```diff
@@ -0,0 +1,231 @@
...
+        defaults={"short_name": "training-data"},
+    )
+    return org
+
+
+def _clear_existing_cve(cve_id: str) -> None:
+    CveRecord.objects.filter(cve_id=cve_id).delete()
+
+
+def _create_container(record: dict[str, Any], org: Organization) -> Container:
```
fricklerhandwerk: This should be done by the serializers themselves via declarative annotations. They should even be able to track relations:
- https://www.django-rest-framework.org/api-guide/relations/
- https://python.plainenglish.io/understanding-deserialization-in-django-rest-framework-drf-a-complete-guide-228e50972f49

The worst that could happen here is that we don't get bulk inserts out of the box, but that's not a big deal for the prototype.

src/shared/tests/test_matching_training_data.py
```diff
@@ -0,0 +1,279 @@
...
+                "was_linked": True,
+            },
+        ],
+        "algorithm_version": 1,
+    }
+
+
+def test_schema_dict_roundtrip_preserves_fields() -> None:
+    record = _sample_record()
+    restored = record_from_dict(record_to_dict(record))
```
fricklerhandwerk: Call `TrainingRecordSerializer` directly here, and remove the wrapper (which does nothing). Otherwise you're needlessly obscuring what's happening.

src/shared/tests/test_matching_training_data.py
```diff
@@ -0,0 +1,279 @@
...
+    serializer = TrainingRecordSerializer(data=record_to_dict(record))
+    assert serializer.is_valid(), serializer.errors
+
+
+def test_schema_version_rejected() -> None:
+    record = _sample_record()
+    record["schema_version"] = SCHEMA_VERSION + 1
+    with pytest.raises(ValueError, match="schema_version"):
+        record_from_dict(record)
+
```
fricklerhandwerk:
```suggestion
```

When you remov the dict conversion wrapper, this becomes unnecessary. At the management command, later, you can just let the `ValidationError` bubble up and show your message. It doesn't have to be super ergonomic, just in principle debuggable; it's an internal tool after all.

src/shared/matching_training_data/export.py
```diff
@@ -0,0 +1,168 @@
...
+from shared.matching_training_data.schema import normalize_record
+from shared.models.cve import Container
+from shared.models.linkage import (
+    CVEDerivationClusterProposal,
+    DerivationClusterProposalLink,
+    PackageOverlay,
+)
+
+
+def user_curated_proposals() -> QuerySet[CVEDerivationClusterProposal]:
```
fricklerhandwerk: This looks very similar to https://github.com/NixOS/nix-security-tracker/pull/1211#discussion_r3703271851 -- in any case this query should live on the model, both for general discoverability and because we're very likely going to reuse it for something.

src/shared/tests/test_matching_training_data.py
```diff
@@ -0,0 +1,279 @@
...
+        derivation=ignored,
+        provenance_flags=int(ProvenanceFlags.PACKAGE_NAME_MATCH),
+    )
+    PackageOverlay.objects.create(
+        suggestion=proposal,
+        package_attribute="foobar.tests",
+        type=PackageOverlay.Type.IGNORED,
+    )
+
+    original = export_proposal(proposal)
```
fricklerhandwerk: `export_proposal` doesn't need to exist. Just feed `proposal` directly to the serializer (annotated with the model appropriately).

src/shared/tests/test_matching_training_data.py
```diff
@@ -0,0 +1,279 @@
...
+    }
+    assert ("foobar", "foobar-1.2.3", "x86_64-linux") in kept
+    assert ("foobar.tests", "foobar-tests-1.2.3", "x86_64-linux") not in kept
+
+    CveRecord.objects.filter(cve_id="CVE-2026-4242").delete()
+    assert not CVEDerivationClusterProposal.objects.filter(
+        cve__cve_id="CVE-2026-4242"
+    ).exists()
+
+    imported = import_record(original)
```
fricklerhandwerk: Same as above, call the serializer on `original` and then do `.save()`.

src/shared/tests/test_matching_training_data.py
```diff
@@ -0,0 +1,279 @@
...
+    first = import_record(record)
+    second = import_record(record)
+    assert first.pk != second.pk
+    assert (
+        CVEDerivationClusterProposal.objects.filter(cve__cve_id="CVE-2026-1111").count()
+        == 1
+    )
+
+
+def test_ensure_benchmark_evaluation_reuses_channel(db: None) -> None:
```
fricklerhandwerk:
```suggestion
def test_ensure_benchmark_evaluation_idempotency(db: None) -> None:
```

No strong opinion... took me a second glance to be sure I understood what the test intended, but it may be me.

src/shared/matching_training_data/constants.py
fricklerhandwerk, line 1: Anything speaking against putting them into `serializers.py`?

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+
+class DerivationDataSerializer(serializers.Serializer):
+    attribute = serializers.CharField()
+    name = serializers.CharField()
+    system = serializers.CharField()
+    known_vulnerabilities = serializers.ListField(
+        child=serializers.CharField(), required=False
+    )
+    provenance_flags = serializers.IntegerField()
+    was_linked = serializers.BooleanField()
```
fricklerhandwerk: [Serializer relations](https://www.django-rest-framework.org/api-guide/relations) should allow transparently encoding the helper table where the "provenance flags" live. I think I understand why you wanted this shortcut, but we don't need the JSON to be extra compact (we're not saving much that way anyway), but letting the serializers do the work is better for keeping it maintainable. Also, noone will look at that JSON ever anyway.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+class ReferenceOverlayDataSerializer(serializers.Serializer):
+    reference_url = serializers.CharField()
+    overlay_type = serializers.CharField()
+    deduplicated_name = serializers.CharField(required=False, allow_blank=True)
+
+
+class LabelsSerializer(serializers.Serializer):
+    status = serializers.CharField()
```
fricklerhandwerk: I think if you annotate the serializer with `model = CVEDerviationClusterProposal`, all you have to do is list the "boring" (char, int, ...) fields by name, and only come up with (similar) serializers for the nested structured ones.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,89 @@
...
+    overlay_type = serializers.CharField()
+
+
+class ReferenceOverlayDataSerializer(serializers.Serializer):
+    reference_url = serializers.CharField()
+    overlay_type = serializers.CharField()
+    deduplicated_name = serializers.CharField(required=False, allow_blank=True)
+
+
+class LabelsSerializer(serializers.Serializer):
```
fricklerhandwerk:
```suggestion
class SuggestionSerializer(serializers.Serializer):
```

This is just dumping (most of) that model's fields to a dict, right? Let's use the names of the models they correspond to.

src/shared/matching_training_data/serializers.py
fricklerhandwerk, line 1: Let's call the serializers exactly by the model they're reading/writing. Then you can access them by qualified name at the call site, e.g.

```
from shared.matching_training_data import serializers

foo = serializers.AffectedProduct
```

Also very likely you won't use the database model name in the same file, in which case you can pre-import and access them directly.

src/shared/models/linkage.py
```diff
@@ -61,6 +61,10 @@ def target_proposals(self) -> "CVEDerivationClusterProposalQuerySet":
             | ~Q(status=CVEDerivationClusterProposal.Status.PENDING)
         )
 
+    def user_curated(self) -> "CVEDerivationClusterProposalQuerySet":
```
fricklerhandwerk, line 73: Side note: This seems to be the same query as here https://github.com/NixOS/nix-security-tracker/pull/1211/changes#r3703271851, and we may or may not pre-emptively facilitate reuse.

src/shared/tests/test_matching_training_data.py
fricklerhandwerk: If you generated it manually by calling the serializer on a Python object, this is essentially generated code checked in. How about we do it the other way round and take objects from fixtures that already exist? Then exercising the serializers will match exactly how they'll be used when calling the API.

fricklerhandwerk: This would also make the actual test case look a lot more like all the other tests, which is exactly what we want: remove noise and emphasize what's special here.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+)
+from shared.models.nix_evaluation import (
+    NixChannel,
+    NixDerivationMeta,
+    NixEvaluation,
+    NixpkgsBranch,
+)
+from shared.models.nix_evaluation import (
+    NixDerivation as NixDerivationModel,
+)
```
fricklerhandwerk:
```suggestion
from shared import models
```

And then use the model names qualified, e.g. `models.CVEDerivationClusterProposal`

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+
+# Stable org for imported training CVEs
+_TRAINING_ORG_UUID = uuid.UUID("aaaaaaaa-bbbb-cccc-dddd-eeeeeeeeeeee")
+
+
+def ensure_benchmark_evaluation() -> NixEvaluation:
+    """Create or reuse the synthetic benchmark channel + completed evaluation."""
+    release_branch, _ = NixpkgsBranch.objects.get_or_create(
+        name=BENCHMARK_RELEASE_BRANCH,
+        defaults={"head_sha1_commit": secrets.token_hex(20)},
```
fricklerhandwerk: Maybe (optional) make the dummy hash a constant here, so it doesn't raise questions next time somebody touches this, such as "is this actually idempotent"?

fricklerhandwerk: I think the element of surprise here is that the randomness seems entirely unwarranted.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+        """
+        evaluation = ensure_benchmark_evaluation()
+        validated_data.pop("schema_version")
+        cve_id = validated_data.pop("cve_id")
+        container_data = validated_data.pop("container")
+        links_data = validated_data.pop("derivationclusterproposallink_set", [])
+        overlays_data = validated_data.pop("package_overlays", [])
+
+        CveRecord.objects.filter(cve_id=cve_id).delete()
+        org = _ensure_organization()
```
fricklerhandwerk: Let's  inline that here, since it's used only once and really doesn't do a lot of work

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+        or proposal.cve.container.first()
+    )
+
+
+class AffectedProduct(serializers.ModelSerializer):
+    cpes = serializers.ListField(child=serializers.CharField(), required=False)
+
+    class Meta:
+        model = AffectedProductModel
+        fields = ("vendor", "product", "package_name", "cpes")
```
fricklerhandwerk:
```suggestion
        fields = (
            "vendor",
            "product",
            "package_name",
            "cpes",
        )
```

Generally easier to parse later in the game when eventually some of these entries will change (such as adding or removing fields). This applies to all sorts of non-trivial or not-guaranteed-to-be-static-forever lists.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+        return {
+            "vendor": instance.vendor,
+            "product": instance.product,
+            "package_name": instance.package_name,
+            "cpes": sorted(cpe.name for cpe in instance.cpes.all()),
+        }
+
+
+class Container(serializers.Serializer):
+    """Training subset of Container: tags + affected products."""
```
fricklerhandwerk:
```suggestion
    """
    Subset of Container fields relevant for training data
    """
```

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+    class Meta:
+        model = PackageOverlayModel
+        fields = ("package_attribute", "type")
+
+
+class CVEDerivationClusterProposal(serializers.ModelSerializer):
+    """Versioned matching training-data record (export/import wire format)."""
+
+    schema_version = serializers.IntegerField()
+    cve_id = serializers.CharField()
```
fricklerhandwerk, line 182: Not sure this is needed, these are primitives. Please check whether we can omit them.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
...
+                vendor=affected_data.get("vendor"),
+                product=affected_data.get("product"),
+                package_name=affected_data.get("package_name"),
+            )
+            for cpe_name in affected_data.get("cpes") or []:
+                cpe, _ = CpeModel.objects.get_or_create(name=cpe_name)
+                affected.cpes.add(cpe)
+            container.affected.add(affected)
+
+        by_fingerprint: dict[tuple[str, str, str], NixDerivationModel] = {}
```
fricklerhandwerk: I suggest to somehow re-use the `as_key()` helper from the ingestion code:

https://github.com/NixOS/nix-security-tracker/blob/858824f972b9839fdc5198d51192f8563c5d16cf/src/shared/evaluation.py#L77-L97

That's our source of truth for what makes a derivation unique in the database. Not sure how to keep the coupling to a minimum, since your serializer outputs are morally very different things than the hydra outputs.

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,410 @@
+"""DRF ModelSerializers for matching training-data roundtrip.
+
```
fricklerhandwerk, line 2: How you're doing all of the following is exactly right, and let's advertise that up front so readers don't have to learn it from the patterns in the code.

```suggestion

The exported derivation and CVE data is exactly what we assume is required for matching them reproducibly.
Adapt that appropriately when the matching algorithm changes.

```

fricklerhandwerk (approved):
Very good!

src/shared/matching_training_data/serializers.py
DarshanCode2005: Confirmed: omitting them raises ImproperlyConfigured, they're not model fields. Adding a short comment documenting why they must stay.

src/shared/evaluation.py
```diff
@@ -75,28 +75,40 @@ class EvaluatedAttribute(JSONWizard):
...
         )
 
 
+def derivation_as_key(
+    *,
+    drv_path: str,
+    attr: str,
+    name: str,
+    meta_name: str | None,
+) -> tuple[str, str, str, str | None]:
```
fricklerhandwerk: If we make this a NamedTuple, it will be obvious from the signature what all those fields are

src/shared/matching_training_data/serializers.py
```diff
@@ -0,0 +1,419 @@
...
+    if evaluation is None:
+        evaluation = models.NixEvaluation.objects.create(
+            channel=channel,
+            commit_sha1=_BENCHMARK_DUMMY_SHA1,
+            state=models.NixEvaluation.EvaluationState.COMPLETED,
+        )
+    return evaluation
+
+
+def _select_container(
```
fricklerhandwerk: Maybe we should put this on the model as `.first_container`? This is used in https://github.com/NixOS/nix-security-tracker/pull/1244/ again... It's not wrong *not* to pollute the model with it since it's not part of the core business logic. But maybe rename it to `pick_first_container()` or something so it's more evident what and why.

PR #1204, opened 2026-07-20
feat: API endpoint for matching training data

DarshanCode2005:
### Summary
- Add `GET /api/v1/matching/training-data` for maintainer export of curated matching proposals (#1038)
- Auth: Knox/session + admin (`isadmin`: staff or security team)
- Paginated (50 default, max 100), payload uses the shared #1203 export schema

### Related files to this PR
* `src/api/matching/*`
* `src/api/tests/test_matching_training_data.py`


src/api/urls.py
```diff
@@ -19,5 +20,10 @@
     path("v1/me", CurrentUserView.as_view(), name="current-user"),
     path("v1/server-info", ServerInfoView.as_view(), name="server-info"),
     path("v1/tokens/me", TokenManagementView.as_view(), name="token-management"),
+    path(
+        "v1/matching/training-data",
```
fricklerhandwerk: @florentc opinions on where to put this, since you've been touching a lot of this recently?

src/api/matching/views.py
```diff
@@ -0,0 +1,54 @@
...
+from shared.auth import isadmin
+from shared.matching_training_data import user_curated_proposals
+from shared.models.linkage import CVEDerivationClusterProposal
+
+
+class IsAdmin(BasePermission):
+    """Staff or security-team members (see shared.auth.isadmin)."""
+
+    def has_permission(self, request: Request, view: APIView) -> bool:  # pyright: ignore[reportIncompatibleMethodOverride]
+        return bool(request.user and isadmin(request.user))
```
fricklerhandwerk: Should probably be a separate role so contributors can obtain the samples

src/api/urls.py
florentc: Not really.

Seems to me this is not an API endpoint a regular sec tracker user would use. If it's an admin restricted thing I guess it should live under a `v1/admin` prefix.

I don't see what else we would put under `matching` beside `training-data`. Do we have future endpoints in mind? If not maybe let's flatten the route.

DarshanCode2005:
So should I change the route to `v1/matching-training-data`?

fricklerhandwerk:
Yeah let's do that

DarshanCode2005:
Clarifying against the plan for #1038 : this endpoint is read-only export of curated proposals. Auth was intentionally locked to `isadmin` , because in the next follow-up PR `fetch_matching_training_data` will use a Knox/session token against staging/prod. You suggested a separate role so contributors can obtain samples. What I need to know:
1. Who is "contributor" here? : Matching-algorithm contributors outside security/committers, or should we reuse an existing bar (`iscommitter`/`user_can_edit_suggestion`)?
2. Should we widen access in this PR , or keep `isadmin` for this current PR and add a dedicated role when the Github team/ group exists?
3. Is the sensitivity of the payload ( labels, comments, overlays, rejection, reasons) a reason to keep this elevated, or is "any authenticated user who can mint a token" fine for public-ish training corpus?

fricklerhandwerk:
1. Let's just make a new Django group and assign people manually via database for now

    I suggest to hard-code that into a migration so we can't forget to deploy it.

2. As above
3. The concern is not sensitive data but server load. We don't have a concept for implementing rate limiting yet, so authentication to a limited group is the stopgap.

fricklerhandwerk (approved):
Good! Just a few style nits in https://github.com/DarshanCode2005/nix-security-tracker/pull/3 (reviewed together)

PR #1211, opened 2026-07-20
perf: batch metadata deletion properly during garbage collection

adekoder:
Extended garbage_collect to purge derivations from non-latest completed evaluations.

### DIFF
--- a/src/shared/management/commands/garbage_collect.py
+++ b/src/shared/management/commands/garbage_collect.py
@@ -167,27 +167,26 @@ class Command(BaseCommand):
 
         candidates = failed_crashed | completed_unmatched
 
-        meta_ids = list(
-            candidates.filter(metadata__isnull=False).values_list(
-                "metadata_id", flat=True
-            )
-        )
-
+        # Deleting metadata cascades to derivations
         self._delete_in_batches(
-            qs=candidates,
-            model=NixDerivation,
+            qs=NixDerivationMeta.objects.filter(
+                pk__in=candidates.filter(metadata__isnull=False).values_list(
+                    "metadata_id", flat=True
+                )
+            ),
+            model=NixDerivationMeta,
             pk_field="id",
-            label="derivations",
+            label="derivations with metadata",
             batch_size=batch_size,
             dry_run=dry_run,
         )
 
-        meta_candidates = NixDerivationMeta.objects.filter(pk__in=meta_ids)
+        # Delete whatever derivations without metadata remain
         self._delete_in_batches(
-            qs=meta_candidates,
-            model=NixDerivationMeta,
+            qs=candidates,
+            model=NixDerivation,
             pk_field="id",
-            label="derivation metas",
+            label="derivations without metadata",
             batch_size=batch_size,
             dry_run=dry_run,
         )


src/shared/management/commands/garbage_collect.py
```diff
@@ -165,7 +206,17 @@ def _delete_unmatched_derivations(
...
+        superseded_unmatched = NixDerivation.objects.filter(
+            parent_evaluation__state=NixEvaluation.EvaluationState.COMPLETED,
+            cve_links_proposals__isnull=True,
+        ).exclude(
+            parent_evaluation__in=NixEvaluation.objects.latest_completed_per_channel(),
+        )
+
+        candidates = failed_crashed | completed_unmatched | superseded_unmatched
 
         meta_ids = list(
```
fricklerhandwerk: That runs out of memory, since that list is now has dozens of millions of entries. The query for the metadata rows needs to be reworked.

fricklerhandwerk (changes requested):
Logically correct but needs work to be deployable.

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,19 +82,57 @@ def handle(self, *args: Any, **options: Any) -> None:
...
+            pgh_obj_id__in=link_ids,
+            label="derivation link events (superseded)",
+        )
+
+        self._delete_in_batches(
+            qs=candidates,
+            model=DerivationClusterProposalLink,
+            pk_field="id",
+            label="superseded proposal links",
+            batch_size=batch_size,
```
Erethon: Since we already have `dry_run` and `_delete_in_batches` accept a dry run, should we pass it along in case the check in line 118 ever gets removed/changed?

Erethon: While this wasn't touched by this PR, maybe we should the same thing in the called `_delete_in_batches` in line 185 too?

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +206,63 @@ def _delete_unmatched_derivations(
...
+        while True:
+            batch = list(qs.values_list("id", "metadata_id")[:batch_size])
+            if not batch:
+                break
+
+            drv_ids = [pk for pk, _ in batch]
+            meta_ids = [meta_id for _, meta_id in batch if meta_id is not None]
+
+            deleted, _ = NixDerivation.objects.filter(id__in=drv_ids).delete()
+            deleted_drvs += deleted
```
Erethon: Django [docs](https://docs.djangoproject.com/en/dev/ref/models/instances/#django.db.models.Model.delete) specify that:
> The return value will count instances from related models if Django is emulating cascade behavior via Python [on_delete](https://docs.djangoproject.com/en/dev/ref/models/fields/#django.db.models.ForeignKey.on_delete) variants.

So I think in this case the counts will be wrong due to cascaded objects being deleted too. I haven't run the command to verify this however.

adekoder: Not sure i understand, it is just the drvs count i need here, not the cascade records.

src/shared/management/commands/garbage_collect.py
adekoder: Fixed

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +207,63 @@ def _delete_unmatched_derivations(
...
+
+        while True:
+            batch = list(qs.values_list("id", "metadata_id")[:batch_size])
+            if not batch:
+                break
+
+            drv_ids = [pk for pk, _ in batch]
+            meta_ids = [meta_id for _, meta_id in batch if meta_id is not None]
+
+            deleted, _ = NixDerivation.objects.filter(id__in=drv_ids).delete()
```
fricklerhandwerk: This may not be necessary at all, since the linked `NixDerivationMeta` has `on_delete=CASCADE`. At best we could delete orphans (if they even exist, in a follow-up).

adekoder: Fair, making a change now

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,19 +82,58 @@ def handle(self, *args: Any, **options: Any) -> None:
...
         self._delete_inactive_channels(batch_size, dry_run)
-        self.stdout.write("\n[5/5] Pruning stale package attrpaths")
+        self.stdout.write("\n[6/6] Pruning stale package attrpaths")
         self._prune_stale_package_attrpaths(batch_size, dry_run)
 
         self.stdout.write(self.style.SUCCESS("\nGarbage collection complete."))
 
+    def _delete_superseded_matches(self, batch_size: int, dry_run: bool) -> None:
+        candidates = (
+            DerivationClusterProposalLink.objects.filter(
```
fricklerhandwerk: Just as a note for later: This describes the property of not being touched by users, so we may abstract that into a QueryManager.

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,19 +82,58 @@ def handle(self, *args: Any, **options: Any) -> None:
...
+                derivation__parent_evaluation__state=NixEvaluation.EvaluationState.COMPLETED,
```
fricklerhandwerk:
```suggestion
```

I think this should be true by construction for every such link, and also not even relevant for this query. If we ever caught derivations in a link that came from evaluations *not* completed, this would be a bug anyway and they should be GC'd in any case.

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,19 +82,58 @@ def handle(self, *args: Any, **options: Any) -> None:
         # Each step satisfies the cascading constraints that gate the next step.
         # `pghistory` events are never auto-deleted — each step explicitly clears relevant events first.
 
-        self.stdout.write("\n[1/5] Deleting stale matches")
+        self.stdout.write("\n[1/6] Deleting superseded evaluation matches")
```
fricklerhandwerk:
```suggestion
        self.stdout.write("\n[1/6] Deleting matches from superseded evaluations")
```

src/shared/management/commands/garbage_collect.py
```diff
@@ -82,19 +82,58 @@ def handle(self, *args: Any, **options: Any) -> None:
...
-        self.stdout.write("\n[4/5] Deleting inactive channels")
+        self.stdout.write("\n[5/6] Deleting inactive channels")
         self._delete_inactive_channels(batch_size, dry_run)
-        self.stdout.write("\n[5/5] Pruning stale package attrpaths")
+        self.stdout.write("\n[6/6] Pruning stale package attrpaths")
         self._prune_stale_package_attrpaths(batch_size, dry_run)
 
         self.stdout.write(self.style.SUCCESS("\nGarbage collection complete."))
 
+    def _delete_superseded_matches(self, batch_size: int, dry_run: bool) -> None:
```
fricklerhandwerk: Since we reverted the change that introduced rematching on new evaluations, and since this doesn't respect cutoff, there's no guarantee this won't empty out a recently created suggestion. 

Example: A suggestion was made with an evaluation yesterday. Then a new evaluation ran today. The following code will delete the link, since it's not related to the latest completed evaluation but hasn't been triaged yet.

The fix would be to properly finish the rematching first. And then this code would still need tests for concurrency with that. Example: The below query runs in parallel with the evaluation post-update listener that does the rematching. If the last batch here finishes and then continues to delete the now-empty suggestion itself, the rematching that happens to work on the same suggestion would simply crash. Rematching that would produce new candidates but comes later, would not even find that suggestion because it's already gone. Both needs to be accounted for.

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +207,55 @@ def _delete_unmatched_derivations(
...
+        self.stdout.write(f"Found {total} eligible derivations.")
+
+        if total == 0 or dry_run:
+            return
+
+        deleted_metas = 0
+        batch_num = 1
+
+        while True:
+            batch = list(qs.values_list("id", "metadata_id")[:batch_size])
```
fricklerhandwerk: Do we strictly need the `list()` here? Can we do `in batch` directly on the ORM list thingy?

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +207,55 @@ def _delete_unmatched_derivations(
...
+            if not batch:
+                break
+
+            meta_ids = [meta_id for _, meta_id in batch if meta_id is not None]
+
+            meta_deleted, _ = NixDerivationMeta.objects.filter(pk__in=meta_ids).delete()
+            deleted_metas += meta_deleted
+
+            self.stdout.write(
+                f"Batch {batch_num}: Deletd ({deleted_metas} derivation metas)."
```
fricklerhandwerk:
```suggestion
                f"Batch {batch_num}: Deleted ({deleted_metas} derivation metas)."
```

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +207,55 @@ def _delete_unmatched_derivations(
...
+
+        deleted_metas = 0
+        batch_num = 1
+
+        while True:
+            batch = list(qs.values_list("id", "metadata_id")[:batch_size])
+            if not batch:
+                break
+
+            meta_ids = [meta_id for _, meta_id in batch if meta_id is not None]
```
fricklerhandwerk: We'll have an infinite loop if all `batch_size` candidates have no metadata. Very likely we have that many such drvs in production. We need some place where we also delete those. Looks like my suggestion to rely on the metadata cascade was wrong after all.

src/shared/management/commands/garbage_collect.py
```diff
@@ -165,31 +207,55 @@ def _delete_unmatched_derivations(
...
-                "metadata_id", flat=True
-            )
+        # No cutoff: a non-latest completed evaluation can never be matched again, so its
+        # unmatched derivations (including ones just orphaned by `_delete_superseded_matches`)
+        # are eligible immediately.
+        superseded_unmatched = NixDerivation.objects.filter(
+            parent_evaluation__state=NixEvaluation.EvaluationState.COMPLETED,
+            cve_links_proposals__isnull=True,
+        ).exclude(
+            parent_evaluation__in=NixEvaluation.objects.latest_completed_per_channel(),
```
fricklerhandwerk: There's a window for a concurrency issue here. Example where it could fail:
- Matcher sees evaluation A as latest and creates a suggestion with drv x
- Evaluation B finishes and is marked completed
- The GC loop is running and the query captures unlinked drv x which is now not from the latest evaluation any more
- Matcher commits its transaction, linking drv x to a suggestion
- GC tries to delete drv x which is now linked to a suggestion and gets protected, `IntegrityError` and the whole command crashes

What may work is wrapping the loop body in a `try` and simply ignore the `IntegrityError`... not sure.

fricklerhandwerk (dismissed):
We should move the first block into a separate PR, which would either go on its own after rematching is reinstated again, or go together with rematching. Otherwise we'll regularly destroy a bunch of (what currently counts as) normal matches before they can be triaged. And in any case it will require a bunch of work to make it concurrency-safe.

The second block is fine conceptually but needs more edge cases covered (+corresponding tests).

src/shared/management/commands/garbage_collect.py
adekoder: but i'm  only fetching suggestions that are not from latest completed eval.

fricklerhandwerk: Yes, and that's the case my example covers.

src/shared/management/commands/garbage_collect.py
fricklerhandwerk: btw `meta_id` will always be set since we're querying for `isnull=False`. This still needs a way to delete the derivations without metadata.

adekoder: i will include this in next iteration

fricklerhandwerk:
I'm a bit confused why we're deleting unrelated tests here. Let's put this to the side for a moment, we really need to make GC more effective since we're running out of space in production.

fricklerhandwerk (approved):

PR #1214, opened 2026-07-20
feat: package ignore/restore in API and new UI

florentc:
- methods in model for package ignore/restore
    - takes cache update into account
    - removed the logic from legacy webviews
- API endpoints for packages in suggestions
    - GET returns a categorized (original/active/ignored) triplet of packages out of consistency with references and maintainers, but the model isn't like that yet for packages so there is a FIXME to remember it. GET is not used by the UI, the packages are obtained from suggestion serializers
    - PATCH follows same interface as references
    - tests
- New UI
    - Exactly the same button based ignore/restore thing as in legacy UI
    - Optimistic updates like for references (to avoid waiting for API response)
    - Optimistic updates are more complex because we also edit the maintainers (auto ignore maintainers when ignoring packages)

### DIFF
--- a/frontend/src/components/suggestions/ActivityLog.tsx
+++ b/frontend/src/components/suggestions/ActivityLog.tsx
@@ -1,11 +1,11 @@
 import { FormatRelativeTime } from "@ark-ui/react";
 import {
-  FolderMinusIcon,
-  FolderPlusIcon,
   InboxIcon,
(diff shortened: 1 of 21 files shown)


fricklerhandwerk (approved):

fricklerhandwerk:

PR #1228, opened 2026-07-28
feat: paginated suggestion lists (unfiltered) in API and new UI

florentc:
This brings paginated (for now hardcoded page size) suggestion lists to the API and the new UI. Filtering by status or other is not in scope for this PR so everything shows up by most recently updated first (including published or dismissed suggestions). The route under ui-v2 to see that list is `/suggestions`.


fricklerhandwerk:

PR #1233, opened 2026-08-03
feat: fetch and import matching training data commands

DarshanCode2005:
## Description

* Add `manage fetch_matching_training_data` to download paginated curated training records from `/api/v1/matching-training-data` (Knox Bearer via `--token` or `MATCHING_TRAINING_DATA_TOKEN`) into `page-*.json` + `manifest.json`
* Add `manage import_matching_training_data` to load that dump onto the synthetic local `benchmark` channel
* Document the fetch -> import workflow, tests cover pagination/auth, roundtrip, idempotency, bad `schema_version`, and `--limit`

Part of [#1038](https://github.com/NixOS/nix-security-tracker/issues/1038). Depends on [#1204](https://github.com/NixOS/nix-security-tracker/pull/1204) (API) / [#1203](https://github.com/NixOS/nix-security-tracker/pull/1203) (schema).

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -13,6 +13,7 @@ Maps each labeled diagram component to its implementation.
 
 - [fetch_all_channels](../src/shared/management/commands/fetch_all_channels.py)
 - [ingest_delta_cve](../src/shared/management/commands/ingest_delta_cve.py)
+- [fetch_matching_training_data](../src/shared/management/commands/fetch_matching_training_data.py) / [import_matching_training_data](../src/shared/management/commands/import_matching_training_data.py) - see [Offline matching training data](./data_ingestion_and_matching.md#offline-matching-training-data)
 - [Evaluate Nixpkgs](../src/shared/listeners/nix_evaluation.py)
 - [CVE matching](../src/shared/listeners/automatic_linkage.py)
 - [WSGI Django](../src/project/asgi.py) (Daphne ASGI in production)
--- a/docs/data_ingestion_and_matching.md
+++ b/docs/data_ingestion_and_matching.md
@@ -89,3 +89,34 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
 ```
 
 This should produce untriaged matches.
+
+### Offline matching training data
+
+In order to measure matching algorithm accuracy locally when developing, you will need access to production data, which contains past automatic matches curated with corrections by users.
+You need an [API token](https://tracker.security.nixos.org/ui-v2/user/tokens), and permissions to access the endpoint with your user account which you can request by contacting maintainers.
+
+Fetch paginated pages (ca. 500k items total):
+
+```console
+export MATCHING_TRAINING_DATA_TOKEN=<api-token>
+manage fetch_matching_training_data \
+  --base-url https://tracker.security.nixos.org \
+  --output ./training-data/
+```
+
+`--output` must be an empty directory (the command refuses to overwrite existing files).
+Use `--limit N` to fetch only a sample.
+
+This will produce one JSON file per page fetched.
+
+Import into your local database:
+
+> [!WARNING]
+> Do _not_ run `manage listen` during import.
+> The command suppresses `pgpubsub` triggers so workers are not flooded with tasks.
+
+```console
+manage import_matching_training_data --input ./training-data/
+```
+
+Re-importing the same dump is idempotent per CVE ID.
--- a/src/shared/management/commands/fetch_matching_training_data.py
+++ b/src/shared/management/commands/fetch_matching_training_data.py
@@ -1,0 +1,200 @@
+"""Fetch curated matching training-data pages from a tracker API."""
+
+from __future__ import annotations
+
+import argparse
+import json
+import os
+from datetime import UTC, datetime
+from pathlib import Path
+from typing import Any
+from urllib.parse import urlparse
+
+import requests
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+from django.urls import reverse
+
+from shared.matching_training_data import SCHEMA_VERSION
+
+TOKEN_ENV = "MATCHING_TRAINING_DATA_TOKEN"
+
+
+def _resolve_token(cli_token: str | None) -> str:
+    token = cli_token or os.environ.get(TOKEN_ENV)
+    if not token:
+        raise CommandError(
+            f"Provide --token or set the {TOKEN_ENV} environment variable."
+        )
+    return token
+
+
+def _page_size(value: str) -> int:
+    page_size = int(value)
+    if page_size < 1 or page_size > 100:
+        raise argparse.ArgumentTypeError("--page-size must be between 1 and 100.")
+    return page_size
+
+
+def _positive_int(value: str) -> int:
+    parsed = int(value)
+    if parsed < 1:
+        raise argparse.ArgumentTypeError("must be a positive integer")
+    return parsed
+
+
+def _endpoint(base_url: str) -> str:
+    parsed = urlparse(base_url.rstrip("/") + "/")
+    if not parsed.scheme or not parsed.netloc:
+        raise CommandError(f"Invalid --base-url: {base_url!r}")
+    return parsed._replace(
+        path=reverse("matching-training-data"),
+        params="",
+        query="",
+        fragment="",
+    ).geturl()
+
+
+def _ensure_empty_output_dir(output_dir: Path) -> None:
+    """Create output_dir if needed; refuse to write into a nonempty directory."""
+    output_dir.mkdir(parents=True, exist_ok=True)
+    if any(output_dir.iterdir()):
+        raise CommandError(
+            f"Output directory is not empty: {output_dir}. "
+            "Choose an empty directory or remove existing files first."
+        )
+
+
+class Command(BaseCommand):
+    help = (
+        "Download paginated matching training-data records from a tracker "
+        "instance into a local directory."
+    )
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--base-url",
+            default="https://tracker.security.nixos.org",
+            help="Tracker base URL. Defaults to the official deployment.",
+        )
+        parser.add_argument(
+            "--output",
+            required=True,
+            type=Path,
+            help="Empty directory to write page-*.json and manifest.json into",
+        )
+        parser.add_argument(
+            "--token",
+            default=None,
+            help=(
+                "Knox API token for a matching_training_data group member. "
+                f"Defaults to ${TOKEN_ENV}."
+            ),
+        )
+        parser.add_argument(
+            "--page-size",
+            type=_page_size,
+            default=100,
+            help="Page size to request (API max is 100). Default: 100.",
+        )
+        parser.add_argument(
+            "--limit",
+            type=_positive_int,
+            default=None,
+            help="Stop after writing at most N records (for smoke tests).",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        token = _resolve_token(options["token"])
+        page_size: int = options["page_size"]
+        limit: int | None = options["limit"]
+        output_dir: Path = options["output"]
+        _ensure_empty_output_dir(output_dir)
+
+        url = _endpoint(options["base_url"])
+        params: dict[str, Any] | None = {"page": 1, "page_size": page_size}
+
+        total_count: int | None = None
+        pages_written = 0
+        records_written = 0
+
+        while True:
+            response = requests.get(
+                url,
+                headers={"Authorization": f"Bearer {token}"},
+                params=params,
+                timeout=settings.NETWORK_REQUEST_TIMEOUT,
+            )
+
+            if response.status_code != 200:
+                raise CommandError(
+                    f"HTTP {response.status_code} fetching training data: "
+                    f"{response.text}"
+                )
+
+            try:
+                payload = response.json()
+            except ValueError as exc:
+                raise CommandError("Response was not valid JSON.") from exc
+
+            results = payload.get("results")
+            if not isinstance(results, list):
+                raise CommandError("Unexpected response shape: missing results list.")
+
+            if total_count is None:
+                total_count = int(payload.get("count") or 0)
+
+            if limit is not None:
+                remaining = limit - records_written
+                if remaining <= 0:
+                    break
+                results = results[:remaining]
+
+            pages_written += 1
+            page_path = output_dir / f"page-{pages_written:05d}.json"
+            page_path.write_text(
+                json.dumps(results, indent=2, sort_keys=True) + "\n",
+                encoding="utf-8",
+            )
+            records_written += len(results)
+
+            self.stdout.write(
+                f"Wrote {page_path.name} "
+                f"({len(results)} records; {records_written}/{total_count})"
+            )
+
+            if limit is not None and records_written >= limit:
+                break
+
+            next_url = payload.get("next")
+            if next_url and results:
+                url = next_url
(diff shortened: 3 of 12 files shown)


docs/README.md
```diff
@@ -13,6 +13,7 @@ Maps each labeled diagram component to its implementation.
 
 - [fetch_all_channels](../src/shared/management/commands/fetch_all_channels.py)
 - [ingest_delta_cve](../src/shared/management/commands/ingest_delta_cve.py)
+- [fetch_matching_training_data](../src/shared/management/commands/fetch_matching_training_data.py) / [import_matching_training_data](../src/shared/management/commands/import_matching_training_data.py) - see [Working with data locally](./data_ingestion_and_matching.md#offline-matching-training-data)
```
fricklerhandwerk: Heading seems out of sync with the target? Better point to the section heading

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
 ```
 
 This should produce untriaged matches.
+
+### Offline matching training data
+
+To reproduce curated matching labels locally (for offline algorithm benchmarks), fetch the bounded JSON corpus from staging or production and import it onto a synthetic `benchmark` channel. You need a Knox API token for a user in the `matching_training_data` group.
```
fricklerhandwerk:
```suggestion
In order to measure matching algorithm accuracy locally when developing, you will need access to production data, which contains past automatic matches curated with corrections by users. 
You need a [API token](https://tracker.security.nixos.org/ui-v2/user/tokens), and permissions to access the endpoint with your user account which you can request by contacting maintainers.
```

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
...
+
+1. Fetch paginated pages (about 7–10 MiB / ~8k records on staging):
```
fricklerhandwerk:
```suggestion
1. Fetch paginated pages (ca. 500k items total):
```

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
...
+
+```console
+export MATCHING_TRAINING_DATA_TOKEN=<knox-token>
```
fricklerhandwerk:
```suggestion
export MATCHING_TRAINING_DATA_TOKEN=<api-token>
```

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
...
+manage fetch_matching_training_data \
+  --base-url https://tracker.security.nixos.org \
+  --output ./training-data/
+```
```
fricklerhandwerk, line 105:
````suggestion
```

This will produce one JSON file per page fetched.
````

File: src/shared/management/commands/fetch_matching_training_data.py [hunk 6]
```diff
@@ -0,0 +1,164 @@
...
+                "Knox API token for a matching_training_data group member. "
+                f"Defaults to ${TOKEN_ENV}."
+            ),
+        )
+        parser.add_argument(
+            "--page-size",
+            type=int,
+            default=100,
+            help="Page size to request (API max is 100). Default: 100.",
+        )
```
fricklerhandwerk, line 99: Maybe worth adding a `limit` parameter so there's a cleaner way than `Ctrl+C`...

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,164 @@
...
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        token = _resolve_token(options["token"])
+        page_size: int = options["page_size"]
+        if page_size < 1 or page_size > 100:
+            raise CommandError("--page-size must be between 1 and 100.")
+
+        output_dir: Path = options["output"]
+        output_dir.mkdir(parents=True, exist_ok=True)
+        _clear_previous_pages(output_dir)
```
fricklerhandwerk: Given that obtaining a full dataset is expensive for both sides, we should instead refuse to write to a nonempty directory and let the user decide what to do.

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
 ```
 
 This should produce untriaged matches.
+
+### Offline matching training data
+
+To reproduce curated matching labels locally (for offline algorithm benchmarks), fetch the bounded JSON corpus from staging or production and import it onto a synthetic `benchmark` channel. You need a Knox API token for a user in the `matching_training_data` group.
+
+1. Fetch paginated pages (about 7–10 MiB / ~8k records on staging):
+
```
fricklerhandwerk, line 99: Indent the list item body, otherwise you'll just have free-floating paragraphs that happen to start with numbers.

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
...
+```console
+export MATCHING_TRAINING_DATA_TOKEN=<knox-token>
+manage fetch_matching_training_data \
+  --base-url https://tracker.security.nixos.org \
+  --output ./training-data/
+```
+
+2. Import into your local database (do **not** run `manage listen` during import; the command suppresses pgpubsub triggers so cache/linkage workers are not flooded):
```
fricklerhandwerk:
```suggestion
2. Import into your local database

   > [!WARNING]: Do **not** run `manage listen` during import.
   > The command suppresses `pgpubsub` triggers so workers are not flooded with tasks.
```

docs/data_ingestion_and_matching.md
```diff
@@ -89,3 +89,24 @@ manage ingest_bulk_cve --from 2026-01-01 --to 2026-01-31
...
+
+```console
+manage import_matching_training_data --input ./training-data/
+```
+
+Re-importing the same dump is idempotent per CVE id. A follow-up `benchmark_matching` command will score candidate matchers against these labels.
```
fricklerhandwerk:
```suggestion
Re-importing the same dump is idempotent per CVE ID.
```

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+            "--limit",
+            type=int,
+            default=None,
+            help="Stop after writing at most N records (for smoke tests).",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        token = _resolve_token(options["token"])
+        page_size: int = options["page_size"]
+        if page_size < 1 or page_size > 100:
```
fricklerhandwerk: You can (and IMO should) move this type check into `type=` as [described in the docs](https://docs.python.org/3.8/library/argparse.html#type)

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+from urllib.parse import urljoin, urlparse, urlunparse
+
+import requests
+from django.conf import settings
+from django.core.management.base import BaseCommand, CommandError
+
+from shared.matching_training_data import SCHEMA_VERSION
+
+TOKEN_ENV = "MATCHING_TRAINING_DATA_TOKEN"
+API_PATH = "/api/v1/matching-training-data"
```
fricklerhandwerk: Hm, any reason we can't use `reverse()` here?

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+class Command(BaseCommand):
+    help = (
+        "Download paginated matching training-data records from a tracker "
+        "instance into a local directory."
+    )
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--base-url",
+            required=True,
```
fricklerhandwerk: Shouldn't we default this to our deployment?

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+        )
+    return token
+
+
+def _endpoint(base_url: str) -> str:
+    parsed = urlparse(base_url.rstrip("/") + "/")
+    if not parsed.scheme or not parsed.netloc:
+        raise CommandError(f"Invalid --base-url: {base_url!r}")
+    root = urlunparse((parsed.scheme, parsed.netloc, "", "", "", ""))
+    return urljoin(root + "/", API_PATH.lstrip("/"))
```
fricklerhandwerk: Clever! But IMO needlessly low-level. You can instead do `parsed._replace(path=API_PATH)`

https://docs.python.org/3/library/urllib.parse.html

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+        limit: int | None = options["limit"]
+        if limit is not None and limit < 1:
+            raise CommandError("--limit must be a positive integer.")
+
+        output_dir: Path = options["output"]
+        _ensure_empty_output_dir(output_dir)
+
+        url: str | None = _endpoint(options["base_url"])
+        headers = {"Authorization": f"Bearer {token}"}
+        params: dict[str, Any] | None = {"page": 1, "page_size": page_size}
```
fricklerhandwerk: Why not inline at least `headers` in `requests.get(`? Also  I don't see why `url` would ever be `None`. Either user supplies a valid one or crash.

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,192 @@
...
+                    url,
+                    headers=headers,
+                    params=params,
+                    timeout=settings.NETWORK_REQUEST_TIMEOUT,
+                )
+            except requests.RequestException as exc:
+                raise CommandError(f"Request failed: {exc}") from exc
+
+            if response.status_code != 200:
+                body = response.text[:500]
```
fricklerhandwerk: Why the cutoff?

src/shared/management/commands/import_matching_training_data.py
```diff
@@ -0,0 +1,96 @@
...
+    if not pages:
+        raise CommandError(f"No page-*.json files found in {input_dir}")
+    return pages
+
+
+def _load_records(page_path: Path) -> list[dict[str, Any]]:
+    try:
+        raw = json.loads(page_path.read_text(encoding="utf-8"))
+    except (OSError, json.JSONDecodeError) as exc:
+        raise CommandError(f"Failed to read {page_path}: {exc}") from exc
```
fricklerhandwerk:
```suggestion
        raise CommandError(f"Failed to read {page_path}") from exc
```

Won't `exc` be printed already in the stack trace like this?

src/shared/management/commands/import_matching_training_data.py
```diff
@@ -0,0 +1,96 @@
...
+                for raw in _load_records(page_path):
+                    if limit is not None and imported >= limit:
+                        self.stdout.write(
+                            self.style.SUCCESS(
+                                f"Imported {imported} record(s) (--limit reached)."
+                            )
+                        )
+                        return
+
+                    cve_id = raw.get("cve_id", "<unknown>")
```
fricklerhandwerk: Don't we guarantee this will be set and *something*? Crashing here would be actually good, we'd know something went very wrong.

src/shared/management/commands/import_matching_training_data.py
```diff
@@ -0,0 +1,96 @@
...
+                        serializer = mtd.CVEDerivationClusterProposal(data=raw)
+                        serializer.is_valid(raise_exception=True)
+                        serializer.save()
+                    except Exception as exc:
+                        raise CommandError(
+                            f"Failed importing {cve_id} from {page_path.name}: {exc}"
+                        ) from exc
+
+                    imported += 1
+                    if imported % 100 == 0:
```
fricklerhandwerk: Why not just print once every page is done?

src/shared/tests/test_matching_training_data_commands.py
```diff
@@ -0,0 +1,368 @@
...
+    ProvenanceFlags,
+)
+from shared.models.nix_evaluation import NixChannel, NixDerivation, NixEvaluation
+
+
+def _export(proposal: CVEDerivationClusterProposal) -> dict[str, Any]:
+    return cast(dict[str, Any], dict(mtd.CVEDerivationClusterProposal(proposal).data))
+
+
+def _sample_api_record(cve_id: str = "CVE-2026-9001") -> dict[str, Any]:
```
fricklerhandwerk, line 33: The comment may gone lost, but IMO we should just use a regular model fixture and dump it via serializer.

src/shared/management/commands/fetch_matching_training_data.py
DarshanCode2005: Changed this to build the endpoint with reverse

src/shared/management/commands/fetch_matching_training_data.py
DarshanCode2005: Done

src/shared/management/commands/fetch_matching_training_data.py
DarshanCode2005: Okay

src/shared/management/commands/fetch_matching_training_data.py
DarshanCode2005: Okay , just a preference for shorter error messages, although removed the truncation

src/shared/management/commands/import_matching_training_data.py
DarshanCode2005: Okay

src/shared/tests/test_matching_training_data_commands.py
DarshanCode2005: i left the fetch tests using explicit wire-format payloads because those tests are mainly about pagination/output behavior rather than serializer roundtrips, but I can refactor them to serializer-backed fixtures too if you'd prefer consistency there.

src/shared/management/commands/fetch_matching_training_data.py
```diff
@@ -0,0 +1,206 @@
...
+class Command(BaseCommand):
+    help = (
+        "Download paginated matching training-data records from a tracker "
+        "instance into a local directory."
+    )
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--base-url",
+            default=str(settings.BASE_URL),
```
fricklerhandwerk: Hah, that doesn't work because in the local environment this is `localhost`, and that's the only place from which this command is supposed to get called. We need to hardcode the URL here!

src/shared/tests/test_matching_training_data_commands.py
fricklerhandwerk: You're processing it anyway through the mock (whether you `json.dumps()` or pull it through the serializer doesn't matter), so we can as well use the default fixtures and avoid potential for mistyping or otherwise creating unexpectedly invalid test data.

fricklerhandwerk (commented):
I've run an import cycle, seems to work! Nice.

fricklerhandwerk:
Since the API is now in production, I also successfully fetched production data. Thanks to the recent huge GC, it doesn't even put a lot of load on the server, so we may in fact eventually remove the extra permission and just put it behind a rate limit (cc @florentc).

PR #1236, opened 2026-08-04
feat: garbage collect superseded proposal matches

adekoder:
Extended garbage_collect to suggestions from non-latest completed evaluations.


fricklerhandwerk:
Let's have this either wait for the rematching working correctly or subsume the rematching fix in this PR.

fricklerhandwerk:
#1245 is done, so this can be rebased.

PR #1241, opened 2026-08-06
feat: delete all unlinked derivations

fricklerhandwerk:
This gets rid of an enormous amount of garbage.

It's safe to do because we're excluding derivations from the latest
completed evaluation. That is, even if an evaluation just finished and no
derivations from it were linked yet, we will leave them intact. Derivations
from any other evaluation won't ever be matched if they weren't already.

Depends on #1240

### DIFF
--- a/src/shared/management/commands/garbage_collect.py
+++ b/src/shared/management/commands/garbage_collect.py
@@ -82,7 +82,7 @@ class Command(BaseCommand):
         self.stdout.write("\n[2/6] Purging obsolete channel links")
         self._purge_obsolete_channel_links(batch_size)
         self.stdout.write("\n[3/6] Deleting unmatched derivations")
-        self._delete_unmatched_derivations(cutoff, batch_size)
+        self._delete_unmatched_derivations(batch_size)
         self.stdout.write("\n[4/6] Deleting empty evaluations")
         self._delete_empty_evaluations(cutoff, batch_size)
         self.stdout.write("\n[5/6] Deleting inactive channels")
@@ -191,7 +191,7 @@ class Command(BaseCommand):
             event_model=DerivationClusterProposalLinkEvent,
         )
 
-    def _delete_unmatched_derivations(self, cutoff: Any, batch_size: int) -> None:
+    def _delete_unmatched_derivations(self, batch_size: int) -> None:
         failed_crashed = NixDerivation.objects.filter(
             parent_evaluation__state__in=[
                 NixEvaluation.EvaluationState.FAILED,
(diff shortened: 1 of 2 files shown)


src/shared/tests/test_garbage_collect.py
```diff
@@ -624,3 +624,81 @@ def test_stale_proposal_deletion_cascades_to_notifications(
...
+
+def test_gc_deletes_unlinked_derivation_from_stale_eval(
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+) -> None:
+    """Unlinked derivations from non-latest completed evaluations are deleted, along with the now-empty evaluation."""
+    eval_old = make_evaluation(age=timedelta(days=1))
+    drv = make_drv(evaluation=eval_old)
+
+    make_evaluation()
```
fricklerhandwerk: add a drv here and assert it persists

src/shared/management/commands/garbage_collect.py
```diff
@@ -328,7 +334,17 @@ def _delete_in_batches(
...
+                _, details = model.objects.filter(
+                    **{f"{pk_field}__in": batch_pks}
+                ).delete()
+            except models.ProtectedError:
+                self.stdout.write(
+                    self.style.WARNING(
+                        f"Batch {batch_num} skipped for {label}: protected by a concurrent write"
+                    )
+                )
+                continue
```
Erethon, line 352: The tests that currently hang for this PR is because of this. When we hit this exception, we go back to the beginning of the while loop, without updating the queryset, thus trying the same batch again and again in an infinite loop. We should update the queryset to exclude derivations that triggered the exception.

src/shared/management/commands/garbage_collect.py
```diff
@@ -313,9 +319,13 @@ def _delete_in_batches(
...
         totals: dict[str, int] = {}
         batch_num = 1
+        skipped_pks: set[Any] = set()
 
         while True:
-            batch_pks = list(qs.values_list(pk_field, flat=True)[:batch_size])
+            batch_qs = qs
+            if skipped_pks:
+                batch_qs = batch_qs.exclude(**{f"{pk_field}__in": skipped_pks})
+            batch_pks = list(batch_qs.values_list(pk_field, flat=True)[:batch_size])
```
fricklerhandwerk: I don't think we need to make it a list here, it's used only once regularly, and the set update on error can be expected to happen not that often.

```suggestion
            batch_pks = batch_qs.values_list(pk_field, flat=True)[:batch_size]
```

adekoder (approved):

PR #1244, opened 2026-08-10
feat: offline matching benchmark against training labels

DarshanCode2005:
## Summary

* Add `manage benchmark_matching` to score `resolve_linkage_candidates` against curated training-data labels on the synthetic `benchmark` channel (read-only; does not create proposals)
* Fingerprints are `(attribute, name, system)`; report TP, FP, SNR, and rejection-reason agreement vs `kept_derivations` / ignored package overlays
* Document as step 3 of the offline training-data workflow

Part of [#1038](https://github.com/NixOS/nix-security-tracker/issues/1038). Depends on [#1233](https://github.com/NixOS/nix-security-tracker/pull/1233) / [#1204](https://github.com/NixOS/nix-security-tracker/pull/1204) / [#1203](https://github.com/NixOS/nix-security-tracker/pull/1203).

### DIFF
--- a/src/shared/evaluation.py
+++ b/src/shared/evaluation.py
@@ -82,7 +82,9 @@ class EvaluatedAttribute(JSONWizard):
     system: str
 
     def as_key(self) -> DerivationKey:
-        """Unique dictionary key for a derivation (see :func:`derivation_as_key`)."""
+        """
+        Unique dictionary key for a derivation.
+        """
         return derivation_as_key(
             drv_path=self.drv_path,
             attr=self.attr,
--- a/src/shared/management/commands/benchmark_matching.py
+++ b/src/shared/management/commands/benchmark_matching.py
@@ -1,0 +1,76 @@
+"""
+Offline matching quality vs curated training-data labels.
+"""
+
+from __future__ import annotations
+
+import argparse
+from pprint import pformat
+from typing import Any
+
+from django.core.management.base import BaseCommand
+
+from shared.matching_training_data.benchmark import (
+    aggregate,
+    rematch,
+    score_proposal,
+    training_corpus_queryset,
+)
+
+
+def _positive_int(value: str) -> int:
+    parsed = int(value)
+    if parsed < 1:
+        raise argparse.ArgumentTypeError("must be a positive integer")
+    return parsed
+
+
+class Command(BaseCommand):
+    help = (
+        "Score current matching algorithm against curated training data. "
+        "Read-only: does not create proposals."
+    )
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--limit",
+            type=_positive_int,
+            default=None,
+            help="Score at most N training proposals (for smoke tests).",
+        )
+        parser.add_argument(
+            "--quiet",
+            action="store_true",
+            help="Suppress per-CVE ProposalScore lines (summary only).",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        limit: int | None = options["limit"]
+        quiet: bool = options["quiet"]
+
+        self.stdout.write("Querying curated matching data...")
+        qs = training_corpus_queryset()
+        if limit is not None:
+            qs = qs[:limit]
+
+        scores = []
+        for i, original in enumerate(qs, start=1):
+            new_match = rematch(original)
+            score = score_proposal(original, new_match)
+            scores.append(score)
+            if not quiet:
+                self.stdout.write(pformat(score))
+            if i % 100 == 0:
+                self.stdout.write(f"Scored {i} proposals…")
+
+        report = aggregate(scores)
+        if report.proposals == 0:
+            self.stdout.write(
+                self.style.WARNING(
+                    "No training-data proposals found on the benchmark corpus. "
+                    "Import with manage import_matching_training_data first."
+                )
+            )
+            return
+
+        self.stdout.write(self.style.SUCCESS(pformat(report)))
--- a/src/shared/matching_training_data/benchmark.py
+++ b/src/shared/matching_training_data/benchmark.py
@@ -1,0 +1,171 @@
+"""
+Score matcher output against curated training-data labels.
+
+Fingerprint identity is a derivation key shared with evaluation ingestion.
+Call the linkage resolver only - do not persist new proposals.
+"""
+
+from __future__ import annotations
+
+import math
+from collections.abc import Iterable
+from dataclasses import dataclass
+
+from django.db.models import Q, QuerySet
+
+from shared.evaluation import DerivationKey, derivation_as_key
+from shared.listeners.automatic_linkage import (
+    LinkageOutcome,
+    resolve_linkage_candidates,
+)
+from shared.matching_training_data.serializers import (
+    _TRAINING_ORG_UUID,
+    BENCHMARK_CHANNEL_BRANCH,
+    pick_first_container,
+)
+from shared.models.linkage import CVEDerivationClusterProposal, PackageOverlay
+
+
+def label_sets(
+    original_match: CVEDerivationClusterProposal,
+) -> tuple[set[DerivationKey], set[str]]:
+    """
+    Return kept fingerprints and ignored package attributes for a proposal.
+    """
+    ignored = {
+        overlay.package_attribute
+        for overlay in original_match.package_overlays.all()
+        if overlay.type == PackageOverlay.Type.IGNORED
+    }
+    if original_match.status == CVEDerivationClusterProposal.Status.REJECTED:
+        return set(), ignored
+
+    kept: set[DerivationKey] = set()
+    for link in original_match.derivationclusterproposallink_set.select_related(
+        "derivation__metadata"
+    ):
+        drv = link.derivation
+        if drv.attribute in ignored:
+            continue
+        metadata = drv.metadata
+        kept.add(
+            derivation_as_key(
+                drv_path=drv.derivation_path,
+                attr=drv.attribute,
+                name=drv.name,
+                meta_name=metadata.name if metadata is not None else None,
+            )
+        )
+    return kept, ignored
+
+
+def snr(true_positives: int, false_positives: int) -> float:
+    if true_positives < 0 or false_positives < 0:
+        raise ValueError(
+            f"true_positives and false_positives must be non-negative, "
+            f"got {true_positives} and {false_positives}"
+        )
+    if false_positives > 0:
+        return true_positives / false_positives
+    if true_positives > 0:
+        return math.inf
+    return math.nan
+
+
+@dataclass(frozen=True)
+class ProposalScore:
+    cve_id: str
+    true_positives: int
+    false_positives: int
+    rejection_agree: bool
+    snr: float
+
+
+@dataclass(frozen=True)
+class AggregateReport:
+    proposals: int
+    true_positives: int
+    false_positives: int
+    rejection_agree_count: int
+    snr: float
+
+
+def rematch(original: CVEDerivationClusterProposal) -> LinkageOutcome:
+    """
+    Re-run the matcher on the CVE container from a curated training match.
+    """
+    container = pick_first_container(original)
+    if container is None:
+        return LinkageOutcome()
+    return resolve_linkage_candidates(container)
+
+
+def score_proposal(
+    original: CVEDerivationClusterProposal,
+    new_match: LinkageOutcome,
+) -> ProposalScore:
+    """
+    Compare a rematch outcome to curated labels for one training proposal.
+    """
+    kept, _ignored = label_sets(original)
+    proposed: set[DerivationKey] = set()
+    if new_match.derivations:
+        proposed = {
+            derivation_as_key(
+                drv_path=drv.derivation_path,
+                attr=drv.attribute,
+                name=drv.name,
+                meta_name=drv.metadata.name if drv.metadata is not None else None,
+            )
+            for drv in new_match.derivations
+        }
+    outcome_rejection = (
+        new_match.rejection.reason if new_match.rejection is not None else None
+    )
+
+    true_positives = len(proposed & kept)
+    false_positives = len(proposed - kept)
+    return ProposalScore(
+        cve_id=original.cve.cve_id,
+        true_positives=true_positives,
+        false_positives=false_positives,
+        rejection_agree=original.rejection_reason == outcome_rejection,
+        snr=snr(true_positives, false_positives),
+    )
+
+
+def aggregate(scores: Iterable[ProposalScore]) -> AggregateReport:
+    scores_list = list(scores)
+    true_positives = sum(s.true_positives for s in scores_list)
+    false_positives = sum(s.false_positives for s in scores_list)
+    return AggregateReport(
+        proposals=len(scores_list),
+        true_positives=true_positives,
+        false_positives=false_positives,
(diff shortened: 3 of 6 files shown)


src/shared/tests/test_matching_training_data_benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+    aggregate,
+    format_snr,
+    label_sets,
+    score_proposal,
+    training_corpus_queryset,
+)
+from shared.models.linkage import CVEDerivationClusterProposal, ProvenanceFlags
+
+
+def _accepted_record(
```
fricklerhandwerk: Let's create objects here, it's less noisy and one won't be allowed to mistype field names.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+
+@dataclass(frozen=True)
+class AggregateReport:
+    proposals: int
+    tp: int
+    fp: int
+    rejection_agree_count: int
+
+    @property
+    def snr(self) -> str:
```
fricklerhandwerk: Let this return a float, parse the input values in the business logic to guarantee it will make sense.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+        outcome_rejection = None
+    else:
+        outcome = resolve_linkage_candidates(container)
+        proposed = proposed_from_outcome(outcome)
+        outcome_rejection = (
+            outcome.rejection.reason if outcome.rejection is not None else None
+        )
+
+    tp = len(proposed & kept)
+    fp = len(proposed - kept)
```
fricklerhandwerk: This should never happen I guess, but it's fine to be defensive and check that `tp/fp` makes sense and raise an exception otherwise, and only then bake them into the result value.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+        if drv.attribute in ignored:
+            continue
+        kept.add((drv.attribute, drv.name, drv.system))
+    return kept, ignored
+
+
+def proposed_from_outcome(outcome: LinkageOutcome) -> set[Fingerprint]:
+    if outcome.derivations is None:
+        return set()
+    return {(drv.attribute, drv.name, drv.system) for drv in outcome.derivations}
```
fricklerhandwerk: Why not call the `derivation_as_key` helper from the previous PR here?

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+def score_proposal(proposal: CVEDerivationClusterProposal) -> ProposalScore:
+    """Compare matcher output to labels for one curated training proposal."""
+    kept, _ignored = label_sets(proposal)
+    container = _select_container(proposal)
+    if container is None:
+        proposed: set[Fingerprint] = set()
+        outcome_rejection = None
+    else:
+        outcome = resolve_linkage_candidates(container)
+        proposed = proposed_from_outcome(outcome)
```
fricklerhandwerk: Better inline that, the wrapper is too thin to warrant the jumping back and forth when reading.

```suggestion
        proposed = set()
        if outcome.derivations:
            proposed = {derivation_as_key(drv) for drv in outcome.derivations}
```

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+    }
+    if proposal.status == CVEDerivationClusterProposal.Status.REJECTED:
+        return set(), ignored
+
+    kept: set[Fingerprint] = set()
+    for link in proposal.derivationclusterproposallink_set.select_related("derivation"):
+        drv = link.derivation
+        if drv.attribute in ignored:
+            continue
+        kept.add((drv.attribute, drv.name, drv.system))
```
fricklerhandwerk: Same here, use the helper so we're not obscuring the interesting bit behind a noisy syntactic surface

src/shared/matching_training_data/benchmark.py
fricklerhandwerk:
```suggestion
def score_proposal(original_match: CVEDerivationClusterProposal) -> ProposalScore:
    """Compare matcher output to labels for one curated training proposal."""
    kept, _ignored = label_sets(original_match)
    container = _select_container(original_match)
    if container is None:
        proposed: set[Fingerprint] = set()
        outcome_rejection = None
    else:
        new_match = resolve_linkage_candidates(container)
        proposed = proposed_from_outcome(new_match)
```

We need to somehow differentiate in the naming that we're comparing two different sources. It threw me off quite hard that both are sort of proposals. Just a suggestion, maybe you have a better idea for names.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+    return "n/a"
+
+
+@dataclass(frozen=True)
+class ProposalScore:
+    cve_id: str
+    tp: int
+    fp: int
+    kept_size: int
+    proposed_size: int
```
fricklerhandwerk:
```suggestion
```

We can probably remove these two, since they're not used anywhere except in the debug print currently. That would also circumvent the problem of confusing naming.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+    if tp > 0:
+        return "inf"
+    return "n/a"
+
+
+@dataclass(frozen=True)
+class ProposalScore:
+    cve_id: str
+    tp: int
+    fp: int
```
fricklerhandwerk:
```suggestion
    true_positives: int
    false_positives: int
```

src/shared/tests/test_matching_training_data_benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+            package_name=pkg,
+            links=[(pkg, f"{pkg}-1.0", "x86_64-linux")],
+        )
+    )
+    kept, ignored = label_sets(proposal)
+    assert ignored == set()
+    assert kept == {(pkg, f"{pkg}-1.0", "x86_64-linux")}
+
+    score = score_proposal(proposal)
+    assert score.tp >= 1
```
fricklerhandwerk:
```suggestion
    assert score.tp == 1
```

I'm not sure how the matcher would create more stuff, since the prerequisites are fixed.

src/shared/tests/test_matching_training_data_benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+    _import_record(
+        _accepted_record(
+            "CVE-2026-bench-corpus",
+            package_name="benchcorpusuniq",
+            links=[("benchcorpusuniq", "benchcorpusuniq-1.0", "x86_64-linux")],
+        )
+    )
+    assert (
+        training_corpus_queryset().filter(cve__cve_id="CVE-2026-bench-corpus").exists()
+    )
```
fricklerhandwerk:
```suggestion
```

This is testing your fixtures, and will be obsolete anyway if you switch to using objects directly.

src/shared/tests/test_matching_training_data_benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+            package_name=pkg,
+            links=[(pkg, f"{pkg}-1.0", "x86_64-linux")],
+        )
+    )
+    out = StringIO()
+    call_command("benchmark_matching", stdout=out)
+    text = out.getvalue()
+    assert "Scored 1 proposal" in text
+    assert "tp=" in text
+    assert "snr=" in text
```
fricklerhandwerk:
```suggestion
```

Having a smoke test is good, but let's not overdo the string matching as that stuff will change anyway.

src/shared/tests/test_matching_training_data_benchmark.py
```diff
@@ -0,0 +1,239 @@
...
+        _import_record(
+            _accepted_record(
+                f"CVE-2026-bench-lim-{i}",
+                package_name=pkg,
+                links=[(pkg, f"{pkg}-1.0", "x86_64-linux")],
+            )
+        )
+    out = StringIO()
+    call_command("benchmark_matching", "--limit", "1", stdout=out)
+    assert "Scored 1 proposal" in out.getvalue()
```
fricklerhandwerk:
```suggestion
```

Let's not overdo the testing, this is rather trivial stuff and may change more or less quickly, and also all this is developer-facing so we don't need to exhaust ourselves with ergonomics.

src/shared/management/commands/benchmark_matching.py
```diff
@@ -0,0 +1,77 @@
...
+        )
+        parser.add_argument(
+            "--verbose",
+            action="store_true",
+            help="Print one line per CVE (tp/fp/snr).",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        limit: int | None = options["limit"]
+        if limit is not None and limit < 1:
```
fricklerhandwerk: There's probably a declarative way to annotate the parser to expect a positive number

src/shared/management/commands/benchmark_matching.py
```diff
@@ -0,0 +1,77 @@
...
+        qs = training_corpus_queryset()
+        if limit is not None:
+            qs = qs[:limit]
+
+        scores = []
+        for i, proposal in enumerate(qs, start=1):
+            score = score_proposal(proposal)
+            scores.append(score)
+            if verbose:
+                self.stdout.write(
```
fricklerhandwerk: Maybe there's a way with `pprint` so we don't have to track the actual data structure here?

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,139 @@
...
+        rejection_agree_count=sum(1 for s in scores_list if s.rejection_agree),
+        snr=snr(true_positives, false_positives),
+    )
+
+
+def training_corpus_queryset() -> QuerySet[CVEDerivationClusterProposal]:
+    """User-curated proposals from the synthetic benchmark / training import graph."""
+    return (
+        CVEDerivationClusterProposal.objects.user_curated()
+        .filter(
```
fricklerhandwerk, line 156: Side note: Do we have a catch-all query for everything related to training data? I keep thinking how we can clean it up locally, or whether just swapping out the DB is the easier approach. The development VM I'm working on would make that somewhat easy: `mv` the image, restart the VM, do the metrics thingy, `mv` back when done. But one needs to do these steps manually and not forget them, otherwise cleaning up would require refetching a whole dump -- or writing that query. Thoughts?

src/shared/evaluation.py
```diff
@@ -91,12 +91,30 @@ def as_key(self) -> DerivationKey:
         )
 
 
+@overload
+def derivation_as_key(derivation: NixDerivation, /) -> DerivationKey:
```
fricklerhandwerk: I'm not sure why exactly we need to be that clever. What prevents us from passing the right parameters at the call site, where we have all the context for making the right decisions?

What makes me skeptical is that this is double-clever: You have this overloading at the interface, and the the sorting-through-the-input in the implementation. This puts a lot of load on a little helper procedure that should IMO solely focus on expressing what distinguishes a derivation.

DarshanCode2005: That overload was added because of earlier note to reuse `derivation_as_key` at the benchmark call sites, including the suggested form `{derivation_as_key(drv) for drv in ...}`. The helper from #1203 was kwargs-only, so I added the overload to make that call shape work. I'll drop the overload.

src/shared/matching_training_data/benchmark.py
DarshanCode2005: Right now, there is not catch-all cleanup. What there is rn:
* _TRAINING_ORG_UUID on imported CVEs
* the synthetic benchmark channel or evaluation for linked derivations
* `training_corpus_queryset()` for scoring (proposals only)

That is enough to find the training graph, but not way to purge, this could get wrong if the dump is too large.

For a full corpus I would also prefer swapping the DB ( or the VM image as u described) over a hand-rolled cleanup. Less to forget, and you don't need a second dump if something goes wrong.

Happy to add a small `purge_matching_training_data` in a follow-up , I wouldn't block this on it.

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,139 @@
...
+    ):
+        drv = link.derivation
+        if drv.attribute in ignored:
+            continue
+        kept.add(derivation_as_key(drv))
+    return kept, ignored
+
+
+def snr(true_positives: int, false_positives: int) -> float:
+    """Return TP/FP, ``inf`` when FP is 0 and TP > 0, or ``nan`` when both are 0."""
```
fricklerhandwerk:
```suggestion
```

That's quite literally what the code says already (taps [sign](https://github.com/NixOS/nix-security-tracker/blob/main/docs/styleguide.md#code-comments) ;))

src/shared/matching_training_data/benchmark.py
```diff
@@ -0,0 +1,139 @@
...
+@dataclass(frozen=True)
+class AggregateReport:
+    proposals: int
+    true_positives: int
+    false_positives: int
+    rejection_agree_count: int
+    snr: float
+
+
+def score_proposal(original_match: CVEDerivationClusterProposal) -> ProposalScore:
```
fricklerhandwerk: Taking another sober look, I now see why the initial review took me a while on this one: the name is misleading and concerns are lumped together. This procedure doesn't "score a proposal" but actually first rematches the container from the given match and then compares that. It would be easier to grok if `score_proposal` actually scored one proposal against another, and the rematching would be done externally (in another procedure, such as `rematch(original: CVEDerivationClusterProposal)`).

src/shared/matching_training_data/benchmark.py
fricklerhandwerk: Yeah not saying we should block on it. Just taking note that we need instructions for clean up, one way or another.

src/shared/management/commands/benchmark_matching.py
```diff
@@ -0,0 +1,72 @@
...
+    parsed = int(value)
+    if parsed < 1:
+        raise argparse.ArgumentTypeError("must be a positive integer")
+    return parsed
+
+
+class Command(BaseCommand):
+    help = (
+        "Score resolve_linkage_candidates against curated training-data labels "
+        "on the synthetic benchmark channel. Read-only: does not create proposals. "
```
fricklerhandwerk:
```suggestion
        "Score current matching algorithm against curated training data."
        "Read-only: does not create proposals. "
```

src/shared/management/commands/benchmark_matching.py
```diff
@@ -0,0 +1,72 @@
...
+        "Note: KNOWN_VULNERABILITY rejects but may still return derivations; if the "
+        "label stored REJECTED with empty kept_derivations, those count as FP (v0)."
```
fricklerhandwerk:
```suggestion
```

That seems like an implementation detail which doesn't belong at this level of abstraction, and also seems to verbatim repeat what's evident from the code. What would be much more interesting is knowing why that is in case it's not obvious, but I think it's clear enough for people working with the code so closely that they'd want to improve the matching algorithm.

fricklerhandwerk (approved):
Nice, it works. Given how long it takes until there's some output, let's maybe print verbose by default, and possibly even have one line initially like "Querying curated matching data...".

Also given how long it takes overall, for a possible future iteration we should consider storing the results somehow. Just thinking out loud...

PR #1245, opened 2026-08-14
Update suggestions on new evaluation

adekoder:


src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+
+
+@pgpubsub.listener(SuggestionRefreshChannel)
+def refresh_and_cache_suggestion(pk: int) -> None:
+    """
+    Refresh derivation links, then rebuild the cache, in that order, since
+    caching depends on the refreshed links.
+    """
+    try:
+        with transaction.atomic():
```
fricklerhandwerk: This is not necessary, the whole listener is called in a transaction by pgpubsub

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+        .values_list("pk", flat=True)
+    )
+
+    count = 0
+    for pk in suggestion_pks:
+        pgpubsub.notify(SuggestionRefreshChannel, pk=pk)
+        count += 1
+
+    if count:
+        logger.info(
```
fricklerhandwerk: Not sure we need to log this... if we want to meter this, we can also count the notifications processed and inspect their structured metadata.

src/shared/listeners/package_clustering.py
adekoder: i can't lock a row outside of a transaction block, check the next line in the query i have `select_for_update`

src/shared/listeners/package_clustering.py
adekoder: okay noted

src/shared/tests/test_package_clustering.py
```diff
@@ -412,12 +416,47 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
     assert PackageDerivation.objects.get(derivation=drv_b).package == pkg
 
 
-@pytest.mark.xfail(reason="Not implemented", strict=True)
+@pytest.fixture
+def dispatch_suggestion_refreshes() -> Callable[..., list[int]]:
```
fricklerhandwerk: This *could* be clarified a bit by a name that hints at that this recreates how the trigger is executed. E.g. `mock_trigger_...`

src/shared/tests/test_package_clustering.py
```diff
@@ -412,12 +416,47 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
...
+    """
+    Runs cluster_after_evaluation inside a committed transaction, as the real
+    pgpubsub listener would, and returns the suggestion pks it dispatched to
+    SuggestionRefreshChannel — without processing them. Matches this
+    codebase's convention of testing pgpubsub listeners by invoking their
+    registered callback directly rather than through real LISTEN/NOTIFY.
```
fricklerhandwerk: If changing the name is not enough, it would help a lot to phrase the comments at a much higher level such that it doesn't mention any code tokens (also a good way to prevent drift).

> Recreates how the notification is processed by `pgpubsub`

src/shared/tests/test_package_clustering.py
```diff
@@ -412,12 +416,47 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
...
+        return [call.kwargs["pk"] for call in mock_notify.call_args_list]
+
+    return run
+
+
+@pytest.fixture
+def cluster_and_process_refreshes(
+    dispatch_suggestion_refreshes: Callable[..., list[int]],
+) -> Callable[..., None]:
+    """Like dispatch_suggestion_refreshes, but also processes each dispatched pk, as a listener process would."""
```
fricklerhandwerk:
```suggestion
    """
    Recreates how the notifications following the package clustering are processed
    """
```

src/shared/tests/test_package_clustering.py
```diff
@@ -467,10 +508,151 @@ def test_only_pending_and_accepted_suggestions_updated(
...
+) -> None:
+    """All affected suggestions get their caches rebuilt, not just the first one."""
+    old_eval = make_evaluation()
+    new_eval = make_evaluation()
+
+    cached_befores = []
+    for i in range(3):
+        container = make_container(cve_id=f"CVE-2025-100{i}")
+        old_drv = make_drv(pname="foo", evaluation=old_eval)
+        make_drv(pname="foo", evaluation=new_eval, attribute=old_drv.attribute)
```
fricklerhandwerk: Maybe you also want to store the new drvs and later check that they're indeed linked?

src/shared/tests/test_package_clustering.py
```diff
@@ -467,10 +508,151 @@ def test_only_pending_and_accepted_suggestions_updated(
...
+        cache_new_suggestions(suggestion)
+        suggestions.append(suggestion)
+
+    failing_pk = suggestions[0].pk
+    original = refresh_suggestion_derivation_links
+
+    def maybe_fail(suggestion: CVEDerivationClusterProposal) -> None:
+        if suggestion.pk == failing_pk:
+            raise RuntimeError("simulated failure")
+        original(suggestion)
```
fricklerhandwerk:
```suggestion
    def maybe_fail(suggestion: CVEDerivationClusterProposal) -> None:
        if suggestion.pk == failing_pk:
            raise RuntimeError("simulated failure")
        refresh_suggestion_derivation_links(suggestion)
```

This temporary name is not required. I noticed we already use it in some existing test, I apparently failed to review that out. We can fix it on a different occasion, but not opposed to doing it now.

src/shared/tests/test_package_clustering.py
```diff
@@ -467,10 +508,151 @@ def test_only_pending_and_accepted_suggestions_updated(
...
+    make_evaluation: Callable[..., NixEvaluation],
+    make_drv: Callable[..., NixDerivation],
+    make_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    """
+    `on_commit` never fires if the enclosing transaction rolls back: the
+    on_commit-registered dispatch must never even be called, not just leave
+    no observable trace (which a synchronous, non-gated call would also do,
+    since its own writes would be rolled back along with everything else).
+    """
```
fricklerhandwerk:
```suggestion
```

Given the name I don't think we need that

nix/configuration.nix
```diff
@@ -164,6 +168,16 @@ in
...
     };
 
+    suggestionRefreshProcesses = mkOption {
+      description = ''
+        How many parallel pgpubsub listener processes to run for
+        SuggestionRefreshChannel, i.e. how many suggestion derivation-link
+        refreshes can run concurrently after an evaluation completes.
+      '';
+      type = types.int;
+      default = 8;
```
fricklerhandwerk:
```suggestion
      default = 2;
```

There's no real hurry here, and we only have 8 cores in prod.

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+                    CVEDerivationClusterProposal.Status.ACCEPTED,
+                ],
+            )
+            refresh_suggestion_derivation_links(suggestion)
+            cache_new_suggestions(suggestion)
+    except CVEDerivationClusterProposal.DoesNotExist:
+        logger.info("Suggestion %d no longer pending/accepted, skipping refresh", pk)
+
+
+def _dispatch_suggestion_refreshes(channel_id: int, evaluation_repr: str) -> None:
```
fricklerhandwerk: Let's move it to right below the call site, a bit confusing to have it down here

fricklerhandwerk: Also let's just inline it in the dispatch helper, it's not doing any significant work that mandates a separate procedure. In the corresponding test we can mock the pgpubsub notifier instead to let it crash.

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+    suggestion affected by this evaluation's clustering.
+
+    """
+    channel_id = evaluation.channel_id
+    evaluation_repr = str(evaluation)
+
+    def _run_after_commit() -> None:
+        try:
+            _dispatch_suggestion_refreshes(channel_id, evaluation_repr)
+        except Exception:
```
fricklerhandwerk: Not sure we need this level of defenisveness here. The procedure does a query and then a quick loop. If any of that fails, sure it may crash the listener, but it will just restart and pick up the queue again. And the stack trace will show up in the logs just as well.

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
         f"updated {result.packages_updated}, created {result.packages_created} packages, "
         f"updated {result.attrpaths_updated}, created {result.attrpaths_created} attrpaths."
     )
+
+    _update_suggestions_after_clustering(evaluation)
+
+
+def _update_suggestions_after_clustering(evaluation: NixEvaluation) -> None:
```
fricklerhandwerk: At that point we can inline the whole thing.

src/shared/listeners/package_clustering.py
```diff
@@ -26,3 +30,76 @@ def cluster_after_evaluation(old: NixEvaluation, new: NixEvaluation) -> None:
...
+            suggestion = CVEDerivationClusterProposal.objects.select_for_update().get(
+                pk=pk,
+                status__in=[
+                    CVEDerivationClusterProposal.Status.PENDING,
+                    CVEDerivationClusterProposal.Status.ACCEPTED,
+                ],
+            )
+            refresh_suggestion_derivation_links(suggestion)
+            cache_new_suggestions(suggestion)
+    except CVEDerivationClusterProposal.DoesNotExist:
```
fricklerhandwerk, line 69:
```suggestion
    except CVEDerivationClusterProposal.DoesNotExist:
        # We don't want to try again! If we crash here, the work is put back on the queue.
```

src/shared/tests/test_package_clustering.py
```diff
@@ -467,10 +508,151 @@ def test_only_pending_and_accepted_suggestions_updated(
...
+    cluster_and_process_refreshes(
         old=NixEvaluation(state=NixEvaluation.EvaluationState.IN_PROGRESS),
         new=new_eval,
     )
 
     link = DerivationClusterProposalLink.objects.get(proposal=suggestion)
     assert link.derivation == old_drv
+
+
+@pytest.mark.django_db(transaction=True)
```
fricklerhandwerk: Shouldn't this be in `test_linkage.py` since this is about linking and not clustering?

src/shared/tests/test_package_clustering.py
```diff
@@ -412,12 +416,45 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
     assert PackageDerivation.objects.get(derivation=drv_b).package == pkg
 
 
-@pytest.mark.xfail(reason="Not implemented", strict=True)
+@pytest.fixture
+def mock_trigger_pubsub_notify_and_cluster_after_evaluation() -> Callable[
+    ..., list[int]
+]:
+    """Triggers cluster_after_evaluation and captures the suggestion pks it dispatches to channel."""
```
fricklerhandwerk:
```suggestion
    """Run the post-evaluation handler and capture suggestions triggered for link refreshing."""
```

src/shared/tests/test_package_clustering.py
```diff
@@ -412,12 +416,45 @@ def run(pk: int, wait_in_critical_section: bool = False) -> None:
     assert PackageDerivation.objects.get(derivation=drv_b).package == pkg
 
 
-@pytest.mark.xfail(reason="Not implemented", strict=True)
+@pytest.fixture
+def mock_trigger_pubsub_notify_and_cluster_after_evaluation() -> Callable[
```
fricklerhandwerk: Maybe simpler...

```suggestion
def mock_cluster_after_evaluation_trigger() -> Callable[
```

src/shared/listeners/package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: fixed

src/shared/listeners/package_clustering.py
adekoder: fixed

src/shared/listeners/package_clustering.py
adekoder: fixed

src/shared/listeners/package_clustering.py
adekoder: i think it is best we leave it separated for readability and a bit of separation of concerns

src/shared/listeners/package_clustering.py
adekoder: fixed

src/shared/tests/test_package_clustering.py
adekoder: i think it make sense to leave this in the package clustering test as it is a followup task after clustering in done

fricklerhandwerk: yeah but we're not testing anything about how it clusters, only the (re-)matching

PR #1248, opened 2026-08-19
feat: issues in API and new UI

florentc:
This implements issues and issues lists in the API and new UI. It also simplifies some frontend code (in particular view mode management). It also introduces navigation links in the header bar to switch between suggestion list and issues list (with better support of narrow screens aka mobile mode).


src/shared/models/issue.py
```diff
@@ -46,6 +46,17 @@ class NixpkgsIssue(TimeStampMixin):
     def __str__(self) -> str:
         return self.code
 
+    @property
+    def github_issue_url(self) -> str | None:
+        """URL of the published GitHub issue, or None if not (yet) published.
+
+        Relies on `self.events` being prefetched and filtered to
```
fricklerhandwerk: The filter should be done here, otherwise this requires ambient authority. Prefetching at the call site for batches on the other hand is a routine thing, and there's no need to even mention it here.

src/api/issues/views.py
```diff
@@ -1,64 +1,83 @@
...
-        def get_cve(self, obj: NixpkgsIssue) -> str | None:
-            first = obj.suggestions.select_related("cve").first()
-            return first.cve.cve_id if first else None
+class IssueViewSet(ListModelMixin, RetrieveModelMixin, viewsets.GenericViewSet):
+    queryset = NixpkgsIssue.objects.all()
+    serializer_class = IssueSerializer
+    pagination_class = IssuePagination
+    permission_classes = [AllowAny]
+    lookup_field = "code"
+    lookup_value_regex = r"NIXPKGS-[0-9]{4}-[0-9]{4,19}"
```
fricklerhandwerk: Can we maybe reuse the regex from the model somehow?

florentc: done

PR #1256, opened 2026-08-21
feat: match CVEs by CPEs from Nixpkgs

### LINKED ISSUE #794: Match by CPE
With https://github.com/NixOS/nixpkgs/issues/354012 closed, we can now ingest `meta.identifiers` and use that to match derivations against CPEs in CVEs.

DarshanCode2005:
- Ingest nixpkgs `meta.identifiers` (`cpeParts` vendor/product and definitive `cpe` strings) into `NixDerivationMeta` during evaluation
- Match CVE CPEs against those fields (vendor + product), unioned with existing package-name/product matching
- Add `ProvenanceFlags.CPE_MATCH`, bump matching algorithm version to 2, and update linkage docs

Closes #794

### DIFF
--- a/src/shared/evaluation.py
+++ b/src/shared/evaluation.py
@@ -42,6 +42,30 @@ class LicenseAttribute(JSONWizard):
 
 
 @dataclass
+class CpePartsAttribute(JSONWizard):
+    """Structured CPE parts from nixpkgs meta.identifiers.cpeParts."""
+
+    part: str | None = None
+    vendor: str | None = None
+    product: str | None = None
+    version: str | None = None
+    update: str | None = None
+    edition: str | None = None
+    language: str | None = None
+    sw_edition: str | None = None
+    target_sw: str | None = None
+    target_hw: str | None = None
+    other: str | None = None
+
+
+@dataclass
+class IdentifiersAttribute(JSONWizard):
+    """Software identifiers from nixpkgs meta.identifiers (CPE parts only for now)."""
+
+    cpe_parts: CpePartsAttribute | None = None
+
+
+@dataclass
 class MetadataAttribute(JSONWizard, LoadMixin):
     outputs_to_install: list[str] = field(default_factory=list)
     available: bool = True
@@ -58,6 +82,7 @@ class MetadataAttribute(JSONWizard, LoadMixin):
     license: list[LicenseAttribute] = field(default_factory=list)
     platforms: list[str] = field(default_factory=list)
     known_vulnerabilities: list[str] = field(default_factory=list)
+    identifiers: IdentifiersAttribute | None = None
 
 
 class DerivationKey(NamedTuple):
@@ -280,6 +305,16 @@ class SyncBatchAttributeIngester:
         maintainers = self.parse_maintainers(metadata.maintainers)
         licenses = self.parse_licenses(metadata.license)
 
+        cpe_vendor: str | None = None
+        cpe_product: str | None = None
+        if metadata.identifiers is not None:
+            parts = metadata.identifiers.cpe_parts
+            if parts is not None:
+                if parts.vendor:
+                    cpe_vendor = parts.vendor
+                if parts.product:
+                    cpe_product = parts.product
+
         meta = NixDerivationMeta(
             name=metadata.name,
             insecure=metadata.insecure,
@@ -292,6 +327,8 @@ class SyncBatchAttributeIngester:
             main_program=metadata.main_program,
             position=metadata.position,
             known_vulnerabilities=metadata.known_vulnerabilities,
+            cpe_vendor=cpe_vendor,
+            cpe_product=cpe_product,
         )
 
         return meta, maintainers, licenses
--- a/src/shared/listeners/automatic_linkage.py
+++ b/src/shared/listeners/automatic_linkage.py
@@ -146,7 +146,9 @@ def build_derivation_links(
             proposal=proposal,
             derivation=drv,
             provenance_flags=ProvenanceFlags(
-                getattr(drv, "package_match", 0) | getattr(drv, "product_match", 0)
+                getattr(drv, "package_match", 0)
+                | getattr(drv, "product_match", 0)
+                | getattr(drv, "cpe_match", 0)
             ),
         )
         for drv in derivations
@@ -164,7 +166,11 @@ def build_package_links(
             pkg_id = drv.package_link.package_id
         except PackageDerivation.DoesNotExist:
             continue
-        flags = getattr(drv, "package_match", 0) | getattr(drv, "product_match", 0)
+        flags = (
+            getattr(drv, "package_match", 0)
+            | getattr(drv, "product_match", 0)
+            | getattr(drv, "cpe_match", 0)
+        )
         package_flags[pkg_id] = package_flags.get(pkg_id, 0) | flags
 
     return [
@@ -175,6 +181,24 @@ def build_package_links(
         )
         for pkg_id, flags in package_flags.items()
     ]
+
+
+def _cpe_vendor_product_pairs(
+    filtered_affected: models.QuerySet,
+) -> set[tuple[str, str]]:
+    """Collect (vendor, product) pairs from non-hardware CPE strings on affected products."""
+    pairs: set[tuple[str, str]] = set()
+    for affected in filtered_affected.prefetch_related("cpes"):
+        for cpe in affected.cpes.all():
+            parsed = cpe.parsed
+            if parsed.is_hardware():
+                continue
+            for vendor, product in zip(
+                parsed.get_vendor(), parsed.get_product(), strict=False
+            ):
+                if vendor and product and vendor != "*" and product != "*":
+                    pairs.add((vendor, product))
+    return pairs
 
 
 def produce_linkage_candidates(
@@ -194,6 +218,7 @@ def produce_linkage_candidates(
         .values_list("product", flat=True)
         .distinct()
     )
+    cpe_pairs = _cpe_vendor_product_pairs(filtered_affected)
 
     package_q = Q()
     for name in package_names:
@@ -203,8 +228,16 @@ def produce_linkage_candidates(
     for product in products:
         product_q |= Q(name__icontains=product)
 
+    cpe_q = Q()
+    for vendor, product in cpe_pairs:
+        cpe_q |= Q(
+            metadata__cpe_vendor__iexact=vendor,
+            metadata__cpe_product__iexact=product,
+        )
+
+    match_q = package_q | product_q | cpe_q
     # This does not seem to happen in practice though
-    if not package_q | product_q:
+    if not match_q:
         return NixDerivation.objects.none()
 
     annotations = {}
@@ -220,6 +253,12 @@ def produce_linkage_candidates(
             default=Value(0),
             output_field=IntegerField(),
         )
+    if cpe_q:
+        annotations["cpe_match"] = Case(
+            When(cpe_q, then=Value(ProvenanceFlags.CPE_MATCH)),
+            default=Value(0),
+            output_field=IntegerField(),
+        )
 
     # Methodology:
     # We start with a large list and we remove things as we sort out that list.
@@ -232,7 +271,7 @@ def produce_linkage_candidates(
             attribute__startswith="tests.",
         )
         .filter(
-            package_q | product_q,
+            match_q,
             parent_evaluation__in=list(latest_complete_channels),
         )
         .select_related("metadata", "package_link")
--- a/src/shared/models/linkage.py
+++ b/src/shared/models/linkage.py
@@ -119,7 +119,7 @@ class CVEDerivationClusterProposal(TimeStampMixin):
 
     @classproperty
     def CURRENT_ALGORITHM_VERSION(cls) -> int:  # noqa: N802, N805
-        return 1
+        return 2
 
     cached: "shared.models.cached.CachedSuggestions"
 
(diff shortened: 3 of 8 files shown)


src/shared/models/nix_evaluation.py
```diff
@@ -115,6 +115,17 @@ class NixDerivationMeta(models.Model):
 
     position = models.URLField(null=True)
 
+    # From nixpkgs meta.identifiers.cpeParts (intentful CPE mapping).
```
fricklerhandwerk: What do you mean by "intentful"?

src/shared/models/linkage.py
```diff
@@ -119,7 +119,7 @@ class RejectionReason(models.TextChoices):
 
     @classproperty
     def CURRENT_ALGORITHM_VERSION(cls) -> int:  # noqa: N802, N805
-        return 1
+        return 2
```
fricklerhandwerk, line 122: @adekoder We still haven't wired up rematching on new algo version... I guess this depends on #1245 plus some small extra steps.

This is, deploying it as is will skip all existing suggestions on recaching and make them invisible.

src/shared/models/linkage.py
```diff
@@ -671,6 +671,7 @@ class ProvenanceFlags(IntFlag, boundary=STRICT):
     # Whether the hardware constraint is matched for this derivation.
     HARDWARE_CONSTRAINT_INRANGE = 1 << 4
     KERNEL_CONSTRAINT_INRANGE = 1 << 5
+    CPE_MATCH = 1 << 7
```
fricklerhandwerk: Feel free to put it according to what's important to readers. We can watch a bit more closely in reviews when we're adding numbers. So let's put it to the front (can't make a syntactically correct suggestions thanks to some weird GitHub limitation).

src/shared/listeners/automatic_linkage.py
```diff
@@ -203,8 +231,16 @@ def produce_linkage_candidates(
     for product in products:
         product_q |= Q(name__icontains=product)
 
+    cpe_q = Q()
+    for vendor, product in cpe_pairs:
```
fricklerhandwerk, line 232: I don't understand why we need to form pairs here, we're not using that for anything. Can't we just merge all the vendors and products from all CPEs, and construct the query from those?

src/shared/models/nix_evaluation.py
```diff
@@ -115,6 +115,17 @@ class NixDerivationMeta(models.Model):
...
+    cpe_vendor = models.CharField(max_length=512, null=True)
+    cpe_product = models.CharField(max_length=2048, null=True)
+    # Definitive meta.identifiers.cpe strings when present.
+    cpes = models.ManyToManyField("Cpe", blank=True)
```
fricklerhandwerk:
```suggestion
```

We seem not to be using these

src/shared/evaluation.py
```diff
@@ -41,6 +42,31 @@ class LicenseAttribute(JSONWizard):
...
+    target_sw: str | None = None
+    target_hw: str | None = None
+    other: str | None = None
+
+
+@dataclass
+class IdentifiersAttribute(JSONWizard):
+    """Software identifiers from nixpkgs meta.identifiers (CPE only for now)."""
+
+    cpe: str | None = None
```
fricklerhandwerk: I think we should make up our minds about what's really needed here. Are we likely to ever parse the rendered CPE string ourselves? Surely we don't need it _now_. Let's not capture it in the first place.

src/shared/evaluation.py
```diff
@@ -276,10 +303,21 @@ def parse_meta(
...
         licenses = self.parse_licenses(metadata.license)
 
+        cpe_vendor: str | None = None
+        cpe_product: str | None = None
+        cpe_name: str | None = None
+        if metadata.identifiers is not None:
+            parts = metadata.identifiers.cpe_parts
+            if parts is not None and parts.vendor:
+                cpe_vendor = parts.vendor
+                cpe_product = parts.product
```
fricklerhandwerk: I don't see from the upstream source how `product` would be guaranteed to be coupled to `vendor`, so I think what you may mean is:

```suggestion
            if parts is not None:
                if parts.vendor:
                    cpe_vendor = parts.vendor
                if parts.product:
                    cpe_product = parts.product
```

Would be nice if we could have it all slightly more declarative, but let's not waste time on that right now..

docs/design/02_cpe-based_linkage.md
```diff
@@ -6,38 +6,32 @@ Refer to [the high-level overview](./01_linkage.md).
 
 We assume that we are looking at a `CveRecord` with a container containing at least one affected product which itself contain at least one CPE string.
 
+Nixpkgs packages may declare intentful CPE metadata under [`meta.identifiers`](https://ryantm.github.io/nixpkgs/stdenv/meta/#sec-meta-identifiers) (`cpeParts`, `cpe`). That data is ingested into `NixDerivationMeta` during evaluation (`cpe_vendor`, `cpe_product`, and optional `cpes`).
+
 ## CPE parsing
 
-Following spec XYZ, we parse CPE strings in a structured fashion into a CPE WFN, see [1] for terminology.
+CVE-side CPE strings are stored on `AffectedProduct` and parsed with the `cpe` Python library into a CPE WFN, see [1] for terminology.
```
fricklerhandwerk: Would be worth adding a link to that library.

docs/design/02_cpe-based_linkage.md
```diff
@@ -53,6 +47,8 @@ Every time, a delta CVE archive is ingested, new CveRecord models are inserted.
 
 As soon as one new CveRecord model is inserted, it's immediately run through the matching algorithm for the current channels and new matchings can occur or not.
 
+CPE identifiers on derivations only appear after an evaluation that includes `meta.identifiers`. Existing databases need a fresh channel evaluation to populate them.
```
fricklerhandwerk: I don't know what you mean by the second sentence.

Apart from being evident from how the change is done, this whole new paragraph sounds like a change log which shouldn't be in the design document. I think it can be removed altogether.

docs/design/02_cpe-based_linkage.md
```diff
@@ -6,38 +6,32 @@ Refer to [the high-level overview](./01_linkage.md).
...
-Given all product names in nixpkgs, we are reduced to a string matching problem:
+Linkage candidates are the union of:
 
-- exact matches are prioritized
-- in case of no exact matches, very low bigram similarity can be proposed as well (tuning is TBD).
+1. **Package name / product substring match** (legacy): derivation `name` contains the affected `package_name` or `product` (case-insensitive substring).
+2. **CPE vendor/product match**: for each non-hardware CPE on the affected product, match derivations whose ingested `cpe_vendor` / `cpe_product` equal the CPE vendor and product (case-insensitive).
 
-Once product names are obtained, we can look again at versions and try to match them over V.
+`possibleCPEs` from nixpkgs are **not** ingested or used (they are version guesses, not unambiguous identifiers).
```
fricklerhandwerk: This statement is, if at all, worth a note in the commit message.

docs/design/02_cpe-based_linkage.md
```diff
@@ -6,38 +6,32 @@ Refer to [the high-level overview](./01_linkage.md).
...
-Derivation names are usually: `$product-name-$version`.
-
-On our end, if we can first cut the version, which is possible if we can retrieve the `version` parameter so that we can trim it, we can obtain the `$product-name`.
-
-Given all product names in nixpkgs, we are reduced to a string matching problem:
+Linkage candidates are the union of:
 
-- exact matches are prioritized
-- in case of no exact matches, very low bigram similarity can be proposed as well (tuning is TBD).
+1. **Package name / product substring match** (legacy): derivation `name` contains the affected `package_name` or `product` (case-insensitive substring).
```
fricklerhandwerk: Hm, I'm very skeptical of whether we should touch this document at all. It's very old, and was never great, and merely served as a guideline for implementing the prototype (even if not everything it outlines is done yet). But this specific change details way too many things about the code that will definitely change in some way eventually, and this place won't get updated.

What's missing from this PR though is the main design decision it seems to be making: which precedence the CPE match has. Apparently we're unioning the matches, but why that is a good idea is not explained anywhere. Why wouldn't we e.g. have the CPE match take precedence, to eliminate some of the false positives?

fricklerhandwerk (commented):
From the volume of change it seems like this PR is doing a bit too much, and it took me quite a while to grok everything. But overall it looks good; removing the unused fields and the (IMO unnecessary) changes to the design doc will boil it down to the essential.

Now would be a good moment to give the measurement commands a ride! What's the impact of this change, in numbers?

src/shared/models/nix_evaluation.py
DarshanCode2005: I meant maintainer-provided CPE metadata from meta.identifiers.cpeParts, as opposed to guessed possibleCPEs.

i have rephrased the comment to "maintainer-provided CPE mapping".

src/shared/models/linkage.py
DarshanCode2005: Ahh ! Got it, okay , I will scoped to ingest + matching , and rest as a follow-up once #1245 lands. (plus the small steps you mentioned).

src/shared/models/linkage.py
DarshanCode2005: Done , moved it to the top for readability. Kept the bit value (1 << 7) so existing stored flags stay valid.

src/shared/listeners/automatic_linkage.py
DarshanCode2005: We keep (vendor, product) pairs so both fields come from the _same_ CPE.

If we merged vendors and products independently, a CVE with example (gnu, glibc) and (openssl, openssl) could match a derivation with vendor=gnu and product=openssl via the cross product. pairing avoids that.

src/shared/models/nix_evaluation.py
DarshanCode2005: Removed

src/shared/evaluation.py
DarshanCode2005: Agreed, dropped ingesting meta.identifiers.cpe. We only parse/store `vendor` and `product`.

src/shared/evaluation.py
DarshanCode2005: Good catch , applied your suggestion.

docs/design/02_cpe-based_linkage.md
DarshanCode2005: Agreed the rewrite was too implementation-specific for that old deisgn doc. Reverted the changes to this doc.

On union vs precedence : with CPE coverage still sparse , making CPE match exclusive would drop most suggestions that still only have name/product matches. Union keeps those, and CPE_Match is recorded in provenance so we can measure and later prefer CPE when present.  Happy to revisit precedence once we have numbers from the benchmark commands.

src/shared/listeners/automatic_linkage.py
fricklerhandwerk: Ah right, I was dumb when reading it. Any reason we only take the first such pair, except that the query will likely blow up quite a lot?

Then again, we're not in a terrible hurry when matching. And have you tried counting how many such pairs we even have in each CVE on average?

docs/design/02_cpe-based_linkage.md
fricklerhandwerk: No, by precedence I mean: if there's a product/name match and a CPE match, throw away the first and only use the second.

DarshanCode2005: Okay, thanks for clarification. I’ll keep union in this PR and add a follow-up for CPE-precedence so we can compare SNR on the training corpus

fricklerhandwerk (approved):
This looks good (last nit: https://github.com/NixOS/nix-security-tracker/pull/1256#discussion_r3842181583) but is blocked on rerunning matching on new algo https://github.com/NixOS/nix-security-tracker/pull/1256#discussion_r3839183939. We could skip bumping the algo version another time, simply to get it in... but then it would defeat the purpose of being able to easily chart how well each version performs (we'd have to manually reconstruct that by tracking down each deployment).

src/shared/listeners/automatic_linkage.py
DarshanCode2005: We already take every CPE on the CVE, pairs is a set over all of them. vendors[0] / products[0] is only within a single CPE string

On the full production training dump (21, 023 records):

1. CVEs with >= 1 usable CPE pair -> 1995 (9.5%)
2. Avg pairs among those -> 1.32
3. Median/Max -> 1/20
4. CVEs with > 1 pair -> 269

So the OR-of-pairs query stays small, so I think that there is no need to truncate to one pair per CVE.

I ran matching with current algo on the production training dump: (report)

```
AggregateReport(
  proposals=21032,
  true_positives=21730,
  false_positives=943727,
  rejection_agree_count=15703,
  snr ~ 0.023
)
```

DarshanCode2005: the dump has no derivation vendor / product yet , so this is essentially a name/product baseline, not a measure of CPE lift. Synthetic CPE fixtures locally score TP=1/FP=0. Real SNR needs identifiers on channel evals + in the training export, then we can also compare union vs precedence rule

fricklerhandwerk: I see, good and thanks a lot for running the numbers! Do I understand correctly we can then just iterate over `zip(vendors, products)`?

DarshanCode2005: > I see, good and thanks a lot for running the numbers! Do I understand correctly we can then just iterate over `zip(vendors, products)`?

Yes

src/shared/models/linkage.py
fricklerhandwerk: Once  https://github.com/NixOS/nix-security-tracker/pull/1277 is merged, so we can continue here

PR #1261, opened 2026-08-23
docs: add notes on code comments

fricklerhandwerk:
Depends on https://github.com/NixOS/nix-security-tracker/pull/1259

### DIFF
--- a/docs/styleguide.md
+++ b/docs/styleguide.md
@@ -47,6 +47,50 @@ To minimise turnaround time for getting your contribution merged:
 
   This allows for last-minute changes or resolving merge conflicts without your involvement.
 
+## Code comments
+
+- Comment sparingly.
+
+  Only add comments if names in the code aren't sufficient to explain to readers proficient with Python what's going on and why it's needed.
+
+- Don't mention code tokens in comments.
+  - Bad: mentions the token directly
+
+    ```
+    # Increment `RetryCounter` on each update to `RemoteProcess`.
+    ```
+
+  - Good: describes the concept
+
+    ```
+    # Track how often the remote process was restarted.
+    ```
+
+  Code among prose, all displayed in the same monospace font, is hard to distinguish from the prose itself.
+  And any markup would add noise that makes the prose harder to parse.
+
+  Comments are for prose conveying ideas about high-level concepts.
+  Call these concepts by their real-world names: CVE, package, suggestion, issue, ...
+
+- Write [docstrings](https://docs.python.org/3/glossary.html#term-docstring) with opening and closing `"""` on their own lines.
+
+  Example:
+
+  ```python
+  def foo():
+     """
+     This is a bit easier to read and work with.
+     """
+     pass
+  ```
+
+  Other than that, we're following [PEP 257](https://peps.python.org/pep-0257/).
+
+- Write comments one line per sentence, just like documentation.
+
+  Comment phrasing is changed once in a while, and rewrapping lines will make for noisy diffs.
+  If lines get too long, write shorter sentences.
+
 ## Tagged comments
 
 We use these tagged comments inspired by and loosely following [PEP 350](https://peps.python.org/pep-0350/#mnemonics):


docs/styleguide.md
```diff
@@ -0,0 +1,111 @@
...
+
+- Use [triple-quoted strings](https://docs.python.org/3/glossary.html#term-triple-quoted-string) for [docstrings](https://docs.python.org/3/glossary.html#term-docstring), with opening and closing `"""` on their own lines.
+
+  Example:
+
+  ```python
+  def foo():
+     """
+     This is a bit easier to read and work with.
+     """
```
DarshanCode2005, line 83: https://peps.python.org/pep-0257/

DarshanCode2005 (approved):

docs/styleguide.md
fricklerhandwerk: Ready to take the dirty looks! Though we should mention that otherwise it's PEP257.

docs/styleguide.md
```diff
@@ -47,6 +47,39 @@ To minimise turnaround time for getting your contribution merged:
...
+
+  Only add comments if names in the code aren't sufficient to explain to readers proficient with Python what's going on and why it's needed.
+
+- Don't mention code tokens in comments.
+
+  Code among prose, all displayed in the same monospace font, is hard to distinguish from the prose itself.
+  And any markup would add noise that makes the prose harder to parse.
+
+  Comments are for prose conveying ideas about high-level concepts.
+  Call these concepts by their real-world names: CVE, package, suggestion, issue, ...
```
DarshanCode2005, line 73: While reading this again, got confused about this sentence. You could either close the list with something like `... and similar` or add one more example so it doesn't read as an unfinished thought.

docs/styleguide.md
```diff
@@ -47,6 +47,39 @@ To minimise turnaround time for getting your contribution merged:
 
   This allows for last-minute changes or resolving merge conflicts without your involvement.
 
+## Code comments
+
+- Comment sparingly.
+
+  Only add comments if names in the code aren't sufficient to explain to readers proficient with Python what's going on and why it's needed.
+
+- Don't mention code tokens in comments.
```
DarshanCode2005, line 56: I think this will be read by some new contributor, then adding a before vs after will be helpful, because while re-reading, it took me a while to understand what exactly this scenario is about , so a quick contrast like; (just example) ->

**Bad :** mentions the token directly
_increament `counter` before stop_

**Good:** describes the concept
_track how many retries are required before escape_

docs/styleguide.md
```diff
@@ -47,6 +47,39 @@ To minimise turnaround time for getting your contribution merged:
...
+
+  Code among prose, all displayed in the same monospace font, is hard to distinguish from the prose itself.
+  And any markup would add noise that makes the prose harder to parse.
+
+  Comments are for prose conveying ideas about high-level concepts.
+  Call these concepts by their real-world names: CVE, package, suggestion, issue, ...
+
+- Write [docstrings](https://docs.python.org/3/glossary.html#term-docstring) with opening and closing `"""` on their own lines.
+
+  Otherwise we're following [PEP 257](https://peps.python.org/pep-0257/), and embrace the dirty looks.
```
DarshanCode2005: I don't know how to fix this, but this line to a new reader will be confusing. Someone who doesn't know what "dirty looks" refers to in regards of linter warnings , so please try be a little bit more descriptive.

DarshanCode2005 (approved):

docs/styleguide.md
fricklerhandwerk:
```suggestion
  Call these concepts by their real-world names: CVE, package, suggestion, issue, etc.
```

docs/styleguide.md
fricklerhandwerk: Yes, very good. I'll add that.

DarshanCode2005 (approved):

PR #1268, opened 2026-08-24
fix: inefficient activity log API requests from new UI

florentc:
Fetching the activity log was independent from fetching suggestions/issues.

In lists, this meant 1 request per suggestion shown.

This introduces an optional activity_log flag in API to inline it with the suggestions.

The new UI uses it to get rid of inefficient requests.

### DIFF
--- a/src/api/issues/views.py
+++ b/src/api/issues/views.py
@@ -1,3 +1,5 @@
+from typing import cast
+
 from django.db.models import Prefetch
 from django.db.models.query import QuerySet
 from drf_spectacular.utils import OpenApiParameter, extend_schema
@@ -9,7 +11,9 @@ from rest_framework.request import Request
 from rest_framework.response import Response
 
 from api.issues.serializers import IssueSerializer
+from api.params import ACTIVITY_LOG_PARAMETER, activity_log_requested
 from api.serializers import ErrorDetailSerializer
+from api.suggestions.serializers import build_activity_log_map
 from shared.models import (
     NIXPKGS_ISSUE_CODE_REGEX,
     EventType,
@@ -24,8 +28,7 @@ EXPAND_PARAMETER = OpenApiParameter(
     required=False,
     description=(
         "Comma-separated list of fields to inline instead of returning ids. "
-        "Currently only `suggestions` is supported: when included, the "
-        "`suggestions` field contains full suggestion objects instead of ids."
+        "Currently only `suggestions` is supported: when included, the `suggestions` field contains full suggestion objects instead of ids."
     ),
 )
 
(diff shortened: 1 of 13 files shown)


src/api/issues/views.py
```diff
@@ -62,27 +65,66 @@ def get_serializer_context(self) -> dict:
...
+        # Only meaningful when suggestions are inlined; otherwise there is nothing
+        # to attach the activity log to.
+        context["include_activity_log"] = expand_suggestions and activity_log_requested(
+            cast(Request, self.request)
+        )
         return context
 
+    def _serialize_with_activity_log(
+        self,
+        instance_or_page: NixpkgsIssue | list[NixpkgsIssue],
```
fricklerhandwerk:
```suggestion
        issues: list[NixpkgsIssue],
```

Why don't we put that responsibility to the caller, where we have the context to know what's what?

fricklerhandwerk: We may even require a queryset here, such that scraping out the `suggestion_ids` can be done in the database. (Would need replacing `get_object()` with `get_queryset()` in the detail view.)

src/api/suggestions/serializers.py
```diff
@@ -301,3 +318,20 @@ def folded_event_to_dict(event: FoldedEventType) -> dict:
     elif isinstance(event, FoldedReferenceEvent):
         base["references"] = [dict(r) for r in event.references]
     return base
+
+
+def build_activity_log_map(suggestion_ids: list[int]) -> dict[int, list[dict]]:
+    """Build the folded, serializable activity log for several suggestions at once."""
+    # FIXME(@florentc): Eventually we'll want to:
+    # - remove cancelling events: at the model level through a proper debouncing implementation
```
fricklerhandwerk, line 326: As in: not even store the duplicate events?

src/api/tests/test_suggestion_activity_log.py
```diff
@@ -0,0 +1,107 @@
...
+    response = client.get(detail_url(suggestion.pk))
+    assert response.status_code == 200
+    assert response.data["activity_log"] is None
+
+
+def test_detail_activity_log_inlined_when_requested(
+    make_cached_suggestion: Callable[..., CVEDerivationClusterProposal],
+) -> None:
+    client = APIClient()
+    suggestion = make_cached_suggestion()
```
fricklerhandwerk:
```suggestion
    cached_suggestion: CVEDerivationClusterProposal,
    client: APIClient,
) -> None:
```

We have these as fixtures already

src/api/tests/test_suggestion_activity_log.py
```diff
@@ -0,0 +1,107 @@
...
+
+from rest_framework.reverse import reverse
+from rest_framework.test import APIClient
+
+from shared.models.issue import NixpkgsIssue
+from shared.models.linkage import CVEDerivationClusterProposal
+
+
+def list_url() -> str:
+    return reverse("cvederivationclusterproposal-list")
```
fricklerhandwerk: This could be just variables, not sure why we would make them fixtures.

fricklerhandwerk: And then, we're not using it *that* often, why not just inline such a shallow wrapper?

src/api/suggestions/serializers.py
```diff
@@ -140,6 +143,22 @@ class SuggestionCategorizedMaintainersSerializer(serializers.Serializer):
     orphan = MaintainerSerializer(many=True)
 
 
+class ActivityLogReferenceSerializer(serializers.Serializer):
+    url = serializers.CharField()
+    name = serializers.CharField(allow_null=True)
+
+
+class ActivityLogEntrySerializer(serializers.Serializer):
```
fricklerhandwerk, line 151: As a follow up for some quiet day: Can we maybe strip all the `Serializer` suffixes and prefix the call sites with `serializers.`? That'd be great for readability I think.

fricklerhandwerk (approved):
Nice! Backend part looks good apart from some stylistic factoring concerns.

src/api/suggestions/serializers.py
florentc: This comment isn't new, it's just been moved. It was the result of a live discussion we had together some time ago, where debouncing would be a backend thing and grouping a frontend thing.

src/api/suggestions/serializers.py
florentc: Yes we agreed on doing this after we have our first iteration of the feature complete API

Issue #1045, opened 2026-05-21
Encapsulate the entire setup for development


devnchill:
I would like to share some conversation that happended today between me and Nobbz on unofficial discord server when I was looking for some help related to something else but stumbled upon this .. 

<img width="1217" height="397" alt="Image" src="https://github.com/user-attachments/assets/066a2a10-37da-4ac3-a0c0-f74402f50320" />

>  There you use the flake instead, or channel. The way it is proposed there is really something that I'd consider the least favored way to use the thing.
> Even worse, it reads as if it was setting up a dev-environment trhough system configuration, rather than doing a proper dev shell.
> 
>  Add an input for it, make it flake = false, and use the input in the module. Though it really screams…
> 
> To do it properly and remain pure, the best is actually to just add "${inputs.nix-security-tracker}/nix/dev-setup.nix" to your imports.
> 
> But still, my concerns remain, in my opinion a dev environment should not be set up through system config.
> 
> Looking at the implementation of the module, it just sets up a postgres user and DB. Instead they could suggest to do taht manually, or provide a docker-compose.yml that sets up a local dev DB that does not rely on the availability of a system wide installed postgres that then is "always on".

 So I was wondering if this idea could be implemented , maybe we add docker and docker compose to devshell and then pull up a postgres image using that docker ? this would then not rely on systemwide config/postgres db. 

I would love to hear some thoughts on it...

fricklerhandwerk:
Sorry for the long delay, in part I was procrastinating a reply because it's kinda tricky not to write a treatise. But thank you @devnchill for bringing it up. It was just a matter of time until someone rightfully got annoyed by the situation.

I agree with @NobbZ assessment that this setup with a system-wide service is very much not ideal, but it's that way for a few reasons:
- We don't have a straightforward way to do service management in devshells

  I know https://devenv.sh, and at the surface this goes in the right direction, but I would not take it on as a dependency for a number of reasons.

- I can't afford to maintain instructions for non-automatic setups.

  Things working as promised and not having to care is the whole point of using Nix. I just discussed this with our GSoC contributor @DarshanCode2005 again. One would have to actually know the other environments and keep up with their random drift. And we can't generally know which environment contributors will have, or what is easy or hard for them to do there.

  Ideally, contributors wouldn't have to deal with that at all, and just get into the shell to start working. If it breaks badly enough, delete one file or run some command, and start over. Which brings us back to the first problem.

[...]

PR #1274, opened 2026-08-26
feat: add restricted SSH access for obtaining a database dump

fricklerhandwerk:
The idea behind this change is to allow safely pulling a full database into the development VM (#1045).
Since code running in there is potentially untrusted (e.g. when testing PRs), we don't want that VM to have full access to remote machines.

It also allows us to better separate priviliges for developers and operators, where developers really only need a database and don't have any reason to touch a public instance on the inside.

@DarshanCode2005 @florentc do you have any other use case or reason to connect to staging via SSH?

@devnchill this would enable you to pull a copy from staging as well, when this is merged you can request adding your keys again.

### DIFF
--- a/infra/common.nix
+++ b/infra/common.nix
@@ -10,6 +10,7 @@ in
   imports = [
     "${sources.agenix}/modules/age.nix"
     ./keys.nix
+    ./dump-db.nix
   ];
 
   boot = {
--- a/infra/dump-db.nix
+++ b/infra/dump-db.nix
@@ -1,0 +1,39 @@
+{
+  config,
+  lib,
+  pkgs,
+  ...
+}:
+let
+  inherit (lib) mkOption types;
+  cfg = config.custom.dump-db;
+in
+{
+  options.custom.dump-db = {
+    authorizedKeyFiles = mkOption {
+      type = types.listOf types.path;
+      default = [ ];
+      description = ''
+        Public key files whose owners can fetch a compressed database dump over SSH.
+        Each key is installed with a forced command and the `restrict` option; no other action is possible with the key.
+      '';
+    };
+  };
+
+  config = {
+    users.users.dump-db = {
+      isSystemUser = true;
+      group = "dump-db";
+      shell = pkgs.bashInteractive;
+      openssh.authorizedKeys.keys =
+        let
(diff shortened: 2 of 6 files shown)


infra/dump-db.nix
```diff
@@ -0,0 +1,43 @@
...
+      openssh.authorizedKeys.keys =
+        let
+          command = "pg_dump -U postgres --create ${cfg.database} | zstd";
+          # Escaping rules: sshd(8) AUTHORIZED_KEYS FILE FORMAT command="command"
+          ssh-quote = lib.replaceStrings [ ''\'' ''"'' ] [ ''\\'' ''\"'' ];
+          # Each key file may contain multiple public keys, one per line.
+          # A forced command is prefixed to every line so authorization stays tied to the key, not the file.
+          restrict-key = key: ''restrict,command="${ssh-quote command}" ${key}'';
+        in
+        lib.concatMap (file: restrict-key (builtins.readFile file)) cfg.authorizedKeyFiles;
```
DarshanCode2005: restrict-key will return a string and map takes a list

fricklerhandwerk: yes `cfg.authorizedKeyFiles` is a list; the first argument to `map` is a lambda operating on one list element

fricklerhandwerk: Ah I see what you meant, the output was supposed to be a list. Turns out we don't need concatMap at all.

florentc (approved):
Currently I only log on staging to fetch db dumps indeed. So all fine with me :+1:

DarshanCode2005 (approved):

infra/common.nix
```diff
@@ -10,6 +10,7 @@ in
   imports = [
     "${sources.agenix}/modules/age.nix"
     ./keys.nix
+    ./dump-db.nix
```
Erethon, line 13: Is there a reason anyone should ever need access to the production database in this way? I see that we're not actually using this in production, but wouldn't it make more sense to include it only in infra/staging.nix?

fricklerhandwerk: Yes, I can imagine e.g. @florentc (or generally maintainers) getting a full dump to work with realistic data.

PR #1277, opened 2026-08-26
feat: refresh matches when new algorithm is available

adekoder:
This resolve the [issue](https://github.com/NixOS/nix-security-tracker/issues/1149), 
This allow us to regenerate suggestion rematch on new algorithm update.

### DIFF
--- a/nix/configuration.nix
+++ b/nix/configuration.nix
@@ -403,6 +403,7 @@ in
             };
             script = ''
               ${manage.name} backfill_package_clustering
+              ${manage.name} rematch_stale_suggestions
               ${manage.name} regenerate_cached_suggestions
             '';
           };
--- a/src/shared/management/commands/rematch_stale_suggestions.py
+++ b/src/shared/management/commands/rematch_stale_suggestions.py
@@ -1,0 +1,33 @@
+from typing import Any
+
+import pgpubsub
+from django.core.management.base import BaseCommand
+
+from shared.channels import SuggestionRefreshChannel
+from shared.models.linkage import CVEDerivationClusterProposal
+
+
+class Command(BaseCommand):
+    help = "Dispatch a re-match of untriaged suggestions when there's a newer matching algorithm."
+
+    def handle(self, *args: Any, **options: Any) -> None:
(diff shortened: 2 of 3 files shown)


src/shared/management/commands/rematch_stale_suggestions.py
```diff
@@ -0,0 +1,38 @@
...
+
+from shared.channels import SuggestionRefreshChannel
+from shared.models.linkage import CVEDerivationClusterProposal
+
+
+class Command(BaseCommand):
+    help = (
+        "Dispatch a re-match of PENDING/ACCEPTED suggestions still on an "
+        "outdated algorithm_version, e.g. after a change to the matching "
+        "algorithm."
```
fricklerhandwerk:
```suggestion
        "Dispatch a re-match of untriaged suggestions when there's a newer matching algorithm."
```

src/shared/tests/test_rematch_stale_suggestions.py
```diff
@@ -0,0 +1,69 @@
...
+    out = StringIO()
+    call_command("rematch_stale_suggestions", stdout=out)
+
+    expected_calls = [
+        mock.call(SuggestionRefreshChannel, pk=pending_suggestion.pk),
+        mock.call(SuggestionRefreshChannel, pk=accepted_suggestion.pk),
+    ]
+    mock_notify.assert_has_calls(expected_calls, any_order=True)
+    assert mock_notify.call_count == 2
+    assert "Dispatched 2 stale suggestion(s) for rematch." in out.getvalue()
```
fricklerhandwerk:
```suggestion
```

Nah, let's not overdo it with the string comparison

fricklerhandwerk (approved):
Very good

nix/configuration.nix
```diff
@@ -392,6 +392,7 @@ in
             script = ''
               wst-manage backfill_package_clustering
               wst-manage regenerate_cached_suggestions
+              wst-manage rematch_stale_suggestions
```
fricklerhandwerk: We may want to do the rematching first, since it also updates the cached values.

PR #1278, opened 2026-08-27
fix: double slashes in published tracker issue URLs

fricklerhandwerk:


Erethon (approved):
I ended up with the same diff when I looked at this (minus the test :P), LGTM

PR #1279, opened 2026-08-27
docs: clarify what determines size of training corpus

DarshanCode2005:
slight rewording in documentation


docs/data_ingestion_and_matching.md
```diff
@@ -95,7 +95,7 @@ This should produce untriaged matches.
 In order to measure matching algorithm accuracy locally when developing, you will need access to production data, which contains past automatic matches curated with corrections by users.
 You need an [API token](https://tracker.security.nixos.org/ui-v2/user/tokens), and permissions to access the endpoint with your user account which you can request by contacting maintainers.
 
-Fetch paginated pages (ca. 500k items total):
```
fricklerhandwerk, line 98: What I meant here are the derivations + link entries. Yes we do have 21k CVEs that went through the system, but that's doesn't say  much about the volume of data.

I'm fine with being more precise, but then let's actually be more precise and say what we're counting.

fricklerhandwerk (approved):

PR #1280, opened 2026-08-27
feat: purge matching training data

DarshanCode2005:
* Add manage purge_matching_training_data to remove locally imported training-org CVEs and synthetic benchmark-channel derivations

* Keeps the training organization and benchmark channel/evaluation for the next import

* Document the command, tests cover purge after import and empty corpus via DB pre/post conditions

Follow-up on https://github.com/NixOS/nix-security-tracker/pull/1244#discussion_r3869429160

### DIFF
--- a/docs/data_ingestion_and_matching.md
+++ b/docs/data_ingestion_and_matching.md
@@ -131,7 +131,8 @@ Reports true positives, false positives, and SNR vs kept derivations and ignored
 Prints one line per CVE by default.
 Use `--quiet` for the summary only, and `--limit N` for a sample.
 
-> [!NOTE]
-> There is no dedicated purge command yet.
-> Imported training data is marked by a fixed training-org UUID and the synthetic `benchmark` channel.
-> For a full corpus, swapping the local database (or development VM disk image) is usually simpler than hand-deleting the graph.
+Remove a previously imported corpus:
+
+```console
+manage purge_matching_training_data
+```
--- a/src/shared/management/commands/purge_matching_training_data.py
+++ b/src/shared/management/commands/purge_matching_training_data.py
@@ -1,0 +1,29 @@
+"""
+Remove imported matching training-data from the local database.
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+import pgtrigger
+from django.core.management.base import BaseCommand
+
+from shared.matching_training_data.purge import purge_training_corpus
+
+
+class Command(BaseCommand):
+    help = (
+        "Delete CVEs imported as matching training data and derivations on the "
+        "synthetic benchmark channel. Does not remove the training organization "
+        "or the benchmark channel/evaluation rows."
+    )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        # Quiet the purge: pgpubsub delete triggers would otherwise enqueue work.
+        with pgtrigger.ignore():
+            totals = purge_training_corpus()
+
+        self.stdout.write(
+            self.style.SUCCESS(f"Purged matching training data: {totals}")
+        )
--- a/src/shared/matching_training_data/purge.py
+++ b/src/shared/matching_training_data/purge.py
@@ -1,0 +1,39 @@
+"""
+Remove locally imported matching training-data from the database.
+"""
+
+from __future__ import annotations
+
+from django.db import transaction
+
+from shared.matching_training_data.serializers import (
+    _TRAINING_ORG_UUID,
+    BENCHMARK_CHANNEL_BRANCH,
+)
+from shared.models.cve import CveRecord
+from shared.models.nix_evaluation import NixDerivation
+
+
+@transaction.atomic
+def purge_training_corpus() -> dict[str, int]:
+    """
+    Delete training-org CVEs and synthetic benchmark-channel derivations.
+
+    CVEs are removed first so proposal links do not block derivation cleanup.
+    The training organization and benchmark channel/evaluation rows are kept.
+    """
+    totals: dict[str, int] = {}
+
+    _, cve_details = CveRecord.objects.filter(
+        assigner__uuid=_TRAINING_ORG_UUID
+    ).delete()
+    for label, count in cve_details.items():
+        totals[label] = totals.get(label, 0) + count
+
+    _, drv_details = NixDerivation.objects.filter(
+        parent_evaluation__channel__channel_branch=BENCHMARK_CHANNEL_BRANCH
+    ).delete()
+    for label, count in drv_details.items():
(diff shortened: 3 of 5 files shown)


src/shared/matching_training_data/purge.py
```diff
@@ -0,0 +1,47 @@
...
+    """
+    Delete training-org CVEs and synthetic benchmark-channel derivations.
+
+    CVEs are removed first so proposal links (PROTECT on derivation) do not block
+    derivation cleanup. The training organization and benchmark channel/evaluation
+    rows are kept for the next import.
+    """
+    cve_qs = CveRecord.objects.filter(assigner__uuid=_TRAINING_ORG_UUID)
+    cves_deleted = cve_qs.count()
+    cve_qs.delete()
```
fricklerhandwerk:
```suggestion
    cves_deleted = cve_qs.delete()
```

I recommend just storing the ORM report and pretty-printing that. Check how we're doing it with garbage collection.

Since this is developer tooling, we want maximum precision for minimum maintenance effort. We can expect contributors to know enough about the system to understand what those numbers with internal labels will mean.

src/shared/tests/test_matching_training_data_commands.py
```diff
@@ -366,3 +366,47 @@ def test_import_respects_limit(tmp_path: Path, db: None) -> None:
...
+
+    assert not CveRecord.objects.filter(cve_id="CVE-2026-cmd-purge").exists()
+    assert not CVEDerivationClusterProposal.objects.filter(
+        cve__cve_id="CVE-2026-cmd-purge"
+    ).exists()
+    assert not NixDerivation.objects.filter(
+        parent_evaluation__channel__channel_branch=BENCHMARK_CHANNEL_BRANCH
+    ).exists()
+    assert NixChannel.objects.filter(channel_branch=BENCHMARK_CHANNEL_BRANCH).exists()
+    assert "1 CVE" in out.getvalue()
```
fricklerhandwerk:
```suggestion
```

Screen scraping user-facing output is generally a code smell. Also in this specific case we don't care about UI stability at all, so why nail down something that we want to have freedom to change when needed?

src/shared/tests/test_matching_training_data_commands.py
```diff
@@ -366,3 +366,47 @@ def test_import_respects_limit(tmp_path: Path, db: None) -> None:
...
+
+
+def test_purge_empty_corpus_is_noop(db: None) -> None:
```
fricklerhandwerk: I don't think this is very critical code, so let's not overthink it. But if you want to test it for the sake of diligence, consider how to approach it through structured pre- and post-conditions, as you have in the first test. Also it's fine to test `purge_training_corpus()` in isolation; the command doesn't really do anything else.

src/shared/matching_training_data/purge.py
```diff
@@ -0,0 +1,47 @@
...
+@dataclass(frozen=True)
+class PurgeResult:
+    cves_deleted: int
+    derivations_deleted: int
+
+
+@transaction.atomic
+def purge_training_corpus() -> PurgeResult:
+    """
+    Delete training-org CVEs and synthetic benchmark-channel derivations.
```
fricklerhandwerk, line 20: One sentence per line please

docs/data_ingestion_and_matching.md
```diff
@@ -131,7 +131,11 @@ Reports true positives, false positives, and SNR vs kept derivations and ignored
...
-> Imported training data is marked by a fixed training-org UUID and the synthetic `benchmark` channel.
-> For a full corpus, swapping the local database (or development VM disk image) is usually simpler than hand-deleting the graph.
+Remove a previously imported corpus (training-org CVEs and synthetic `benchmark` channel derivations):
+
+```console
+manage purge_matching_training_data --yes
+```
+
+`--yes` is required.
+The training organization and benchmark channel/evaluation rows are kept for the next import.
```
fricklerhandwerk:
```suggestion
```

We can expect users to read help output, let's not repeat reference documentation.

docs/data_ingestion_and_matching.md
```diff
@@ -131,7 +131,11 @@ Reports true positives, false positives, and SNR vs kept derivations and ignored
 Prints one line per CVE by default.
 Use `--quiet` for the summary only, and `--limit N` for a sample.
 
-> [!NOTE]
-> There is no dedicated purge command yet.
-> Imported training data is marked by a fixed training-org UUID and the synthetic `benchmark` channel.
-> For a full corpus, swapping the local database (or development VM disk image) is usually simpler than hand-deleting the graph.
+Remove a previously imported corpus (training-org CVEs and synthetic `benchmark` channel derivations):
```
fricklerhandwerk:
```suggestion
Remove a previously imported corpus:
```

No need to bake in details that may change (e.g. somewhat likely what exactly we're storing will expand; do you want anyone to know this sentence exists and needs updating?)

src/shared/management/commands/purge_matching_training_data.py
```diff
@@ -0,0 +1,47 @@
...
+class Command(BaseCommand):
+    help = (
+        "Delete CVEs imported as matching training data and derivations on the "
+        "synthetic benchmark channel. Does not remove the training organization "
+        "or the benchmark channel/evaluation rows."
+    )
+
+    def add_arguments(self, parser: argparse.ArgumentParser) -> None:
+        parser.add_argument(
+            "--yes",
```
fricklerhandwerk:
```suggestion
            "--confirm",
```

More self-explanatory. Otherwise, what am I saying "yes" to?

But generally, while I think it's clever to add this extra step so one wouldn't accidentally drop stuff when just trying to figure out what the command will do, I think it's more ceremony than is warranted here. All this happens on local ephemeral data, and this data isn't even particularly valuable; very cheap to replace. For comparison, a full scoring run took hours, I'd put more energy in preserving what is hard to recreate.

src/shared/management/commands/purge_matching_training_data.py
```diff
@@ -0,0 +1,47 @@
...
+            action="store_true",
+            help="Required confirmation; refuse to purge without this flag.",
+        )
+
+    def handle(self, *args: Any, **options: Any) -> None:
+        if not options["yes"]:
+            raise CommandError(
+                "Refusing to purge without --yes. "
+                "This deletes all training-org CVEs and benchmark-channel "
+                "derivations from the local database."
```
fricklerhandwerk:
```suggestion
```

This repeats the help text.

fricklerhandwerk (approved):

PR #1283, opened 2026-08-30
chore: Remove stale info from overlay

### LINKED ISSUE #1193: Upstream into nixpkgs the django-rest-knox package
Similarly to what we did in #563, I think we should upstream the package definition of [django-rest-knox](https://github.com/NixOS/nix-security-tracker/blob/d2609982df1fc2a3978707ea69ff8199e66c68db/nix/overlay.nix#L15-L32) in nixpkgs

Erethon:
This updates overlay.nix to be a bit more up to date:

- https://github.com/NixOS/nixpkgs/pull/541733 is now merged and we've bumped nixpkgs since then.
- The python argparse and commitizen issue is fixed upstream now.

Closes #1193

### DIFF
--- a/nix/overlay.nix
+++ b/nix/overlay.nix
@@ -7,24 +7,6 @@ in
   python3 = prev.python3.override {
     packageOverrides = pyfinal: _pyprev: {
       psycopg2 = pyfinal.psycopg;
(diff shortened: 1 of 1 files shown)


fricklerhandwerk (approved):
Thanks!

Issue #1285, opened 2026-08-31
Add link to API spec on the token page


fricklerhandwerk:
Currently there's no way for users to know how to talk to the API when they generate a token.

PR #1294, opened 2026-09-02
feat: Link to OpenAPI specification in API settings page

florentc:
Implements #1285 and expose a Swagger UI

### DIFF
--- a/frontend/src/components/user-settings/ApiToken.tsx
+++ b/frontend/src/components/user-settings/ApiToken.tsx
@@ -128,6 +128,14 @@ export function ApiToken() {
           </div>
         </div>
       )}
+      <ul className="column gap">
+        <li>
+          <a href="/api/docs/">API documentation</a> via Swagger UI
+        </li>
+        <li>
+          <a href="/api/schema/">API specification</a> as OpenAPI YAML
+        </li>
+      </ul>
     </div>
   );
 }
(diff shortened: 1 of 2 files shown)


frontend/src/components/user-settings/ApiToken.tsx
```diff
@@ -128,6 +128,9 @@ export function ApiToken() {
           </div>
         </div>
       )}
+      <p>
+        Consult the <a href="/api/schema/">OpenAPI specification</a>.
```
fricklerhandwerk, line 132: I think we should emphasize the human-readable representation?

```suggestion
        <ul class="prose">
          <li><a href="/api/swagger-ui/">API documentation</a> via Swagger UI</li>.
          <li><a href="/api/schema/">API specification</a> as OpenAPI YAML</li>.
        </ul>
```

Though we may primarily want to serve the JSON schema, I guess? I always wondered what made the default for the internal spec be in YAML...

PS: Would be really nice to expose the symbolic URL names so we can do the same thing as in Python with `reverse()`...

florentc: The official documentation mentions that YAML is the preferred format https://learn.openapis.org/specification/structure.html because it's more compact and supports comments. So I guess this is why the django default is YAML.

As for swagger, I'm on it

For the named routes, I created an issue https://github.com/NixOS/nix-security-tracker/issues/1313

fricklerhandwerk (approved):
Excellent

PR #1300, opened 2026-09-04
fix: use Nix version ordering in Version.affects()

samiser:
This PR addresses the FIXME in `Version.affects()` about versions being compared as strings.

String comparison gets version ordering wrong, so for example "1.9" would be considered newer than "1.10". This adds `shared/version_compare.py`, a port of Nix's [`builtins.compareVersions`](https://github.com/NixOS/nix/blob/master/src/libstore/names.cc), and uses it for checking constraints.

The port reproduces the reference implementation's quirks exactly: `pre` sorts before every other component (so `2.3pre1` < `2.3`), and numeric components are parsed as C integers so digit runs above 2^31 − 1 compare as strings.

This also supersedes the version comparison part of #960, which left `affects()` unwired and inverted the `pre` ordering. That PR also makes some changes unrelated to version comparison.

## Testing

Besides known-answer tests for each comparison branch, a parity test validates the port against the reference implementation itself. It evaluates `builtins.compareVersions` on 1521 pairs of tricky versions in a single `nix-instantiate` call and requires identical results. Shelling out to Nix is confined to the tests.

The parity test skips gracefully where `nix-instantiate` is unavailable. Also the second commit puts `pkgs.nix` on the `manage` PATH, and `nix-build -A tests` confirms the test now runs and passes there. Running `manage test -- --pyargs shared` within the VM also runs the test and it passes.


fricklerhandwerk (changes requested):
I strongly recommend using a parsing library made by people who have dealt with the problem seriously, since it's not a small problem. For example https://github.com/aboutcode-org/univers

We won't figure out anything even remotely as sophisticated, and for sure we wouldn't have capacity to maintain it.

fricklerhandwerk (approved):
Nice, thank you very much!

I must admit, I had to stare at the original approach of using `builtins.compareVersions` for a while to understand why we would even want to do that in the first place. The primop is used only 90 times in Nixpkgs, most of it in update scripts (looking like a coarse approximation that works good enough), and otherwise simply re-exporting it in `lib`. Nix is not the authority on how Nixpkgs handles package versioning, and while Nixpkgs decided to be backward-compatible with `nix-env` version parsing, it merely takes whatever a software project has and encodes a bit of extra metadata into its syntax rules.

Therefore I'm not entirely convinced it would be a great idea to upstream Nix versioning to `univers`, because that would make it look like as if Nix is in any way privileged, which it is clearly not. From what I can tell, `builtins.compareVersions` was a (clever, but still) hack from the early days to make `nix-env --upgrade` work. On the other hand, since Nixpkgs versions always followed Nix ordering, it's not wrong either, but it may need appropriate documentation so people don't again get confused about the distinction between Nix, Nixpkgs, and NixOS (which has nothing to do with that issue whatsoever).

PR #1306, opened 2026-09-08
feat: Full notification features in API and new UI

florentc:
This completes the notifications API with necessary endpoints and implements a notification center on the new UI.

## Notification center

It is very similar in appearance to the legacy one but comes with quality of life features:

- Suggestions are displayed in the notifications. Not just a stub. By default it's in collapsed mode so it looks like it, but you can unfold as you need.
- The suggestions are mutable: you can dismiss/accept/publish or edit packages from within the notification itself
- These mutations live update the "Matching packages you follow" section
- When no package matches anymore, you get the "obsolete" notice

## Notification bell

The notification bell in the header bar used to be updated as part of full page reloads in the old UI. We no longer have page reloads in the new UI.

For the time being I went with the most pragmatic approach. I introduced an API endpoint that returns the unread counter. The frontend fetches the info on load, then polls every 2 minutes. It uses react query which is smart enough not to poll when the browser tab isn't in focus. 2 minutes is virtually nothing for the backend, the endpoint is tiny, and no notification is more urgent than a 2 minutes maximum delay.

A push notification / websocket approach seemed totally overkill to me. Unless in the future we introduce those globally on sec tracker for most interactions (concurrent suggestion editing, etc). In any case I think this will deserve its own PR.

The bell now animates a little to attract attention upon changes to the counter.

### Disabled buttons

At some point I needed to disable the "Mark all as read" and "Clear read" buttons (before realizing I should not). There never was visual feedback (apart from mouse cursor change) for disabled buttons.

Now there is:

It became unrelated to this PR though. I can remove it and put it in its own PR if needed.

### User settings in navigation bar

I have tried many variants to integrate the bell icon in the navbar. I tried to give it a label (otherwise the "User settings" label looked like text between 2 icons) but it was taking too much space. I tried to make everything a tab (like "Suggestions" and "Nixpkgs issues") but it looked out of place, didn't play well with the user avatar and notification counter, and it made the tabs shift upon loading the authentication state of the user. It also blurred the line between user matters and sectracker matters.

In the end I went with this:

The "User settings" label is removed. Instead, I added a settings icon next to the user avatar to convey that it's a settings button. It has the added benefit of reducing flashing when loading the user authentication status.

Furthermore, this pattern with bell icon + user avatar is very common and conventional (e.g. GitHub).


src/api/tests/test_notifications.py
```diff
@@ -158,3 +158,346 @@ def test_list_notifications_shows_both_types(
...
+
+    url = reverse("notifications-list")
+    response = client.get(url, {"expand": "suggestion"})
+
+    assert response.status_code == status.HTTP_200_OK
+    suggestion = response.data["results"][0]["suggestion"]
+    assert suggestion is not None
+    assert suggestion["id"] == db_notifications[0].suggestion_id
+    assert "packages" in suggestion
+    assert suggestion["activity_log"] is None
```
fricklerhandwerk:
```suggestion
```

Not sure we should be testing the negative space here?  If you want to assert that the activity log is toggled by a query param, why not make a test that parameterizes that flag and checks just for the presence/absence of the activity log?

src/api/tests/test_notifications.py
```diff
@@ -158,3 +158,346 @@ def test_list_notifications_shows_both_types(
...
+    db_notifications = make_maintainer_notification(user)
+
+    url = reverse("notifications-list")
+    response = client.get(url, {"expand": "suggestion"})
+
+    assert response.status_code == status.HTTP_200_OK
+    suggestion = response.data["results"][0]["suggestion"]
+    assert suggestion is not None
+    assert suggestion["id"] == db_notifications[0].suggestion_id
+    assert "packages" in suggestion
```
fricklerhandwerk:
```suggestion
```

IIUC we don't care about the package here

src/api/tests/test_notifications.py
```diff
@@ -158,3 +158,346 @@ def test_list_notifications_shows_both_types(
...
+
+def test_list_notifications_expand_suggestion_with_activity_log(
+    client: APIClient,
+    user: User,
+    make_maintainer_notification: Callable[..., list[Notification]],
+) -> None:
+    make_maintainer_notification(user)
+
+    url = reverse("notifications-list")
+    response = client.get(url, {"expand": "suggestion", "activity_log": "true"})
```
fricklerhandwerk: I don't have a preference, this could either be a parameter on the first test and we'll fold the assertion over activity log into the body there, or have a separate test here just for the activity log.

src/api/tests/test_notifications.py
```diff
@@ -158,3 +158,346 @@ def test_list_notifications_shows_both_types(
...
+
+    url = reverse("notifications-detail", kwargs={"pk": notification.pk})
+    response = client.patch(url, {"is_read": True}, format="json")
+
+    assert response.status_code == status.HTTP_200_OK
+    assert response.data["is_read"] is True
+    notification.refresh_from_db()
+    assert notification.is_read is True
+    user.profile.refresh_from_db()
+    assert user.profile.unread_notifications_count == 0
```
fricklerhandwerk, line 282: We may even want to do another request to make sure the data arrives at the client... IMO stylewise this is better because it checks the same boundary as in the precondition.

src/api/tests/test_notifications.py
```diff
@@ -158,3 +158,346 @@ def test_list_notifications_shows_both_types(
...
+
+
+def test_patch_notification_marks_unread(
+    client: APIClient,
+    user: User,
+    make_maintainer_notification: Callable[..., list[Notification]],
+) -> None:
```
fricklerhandwerk, line 293:
```suggestion
) -> None:
    """
    Check that marking already read notifications back as unread increases the counter and resets the is-read flag.
    """"
```

src/webview/tests/ui_v2/test_notification_center.py
```diff
@@ -0,0 +1,324 @@
...
+    expect(as_staff.get_by_test_id(f"notification-{notification.id}")).to_be_visible()
+
+
+def test_clear_read_resets_to_page_one(
+    live_server: LiveServer,
+    as_staff: Page,
+    staff: User,
+    make_maintainer_notification: Callable[..., list[Notification]],
+) -> None:
+    """Clearing read notifications always navigates back to page 1, even when the list shrinks and the current page would otherwise no longer exist."""
```
fricklerhandwerk:
```suggestion
    """
    Clearing read notifications always navigates back to page 1.
    This is important for when the list shrinks and the current page would no longer exist.
    """
```

src/api/issues/views.py
```diff
@@ -105,7 +101,10 @@ def list(self, request: Request) -> Response:
 
     @extend_schema(
         operation_id="getIssue",
-        description="Get full details of a Nixpkgs security issue.",
+        description=(
+            "Get full details of a Nixpkgs security issue. "
+            "Supports `expand=suggestions` to inline full suggestion objects instead of ids."
```
fricklerhandwerk:
```suggestion
```

This should be visible in the API docs, right?

fricklerhandwerk (approved):
Great stuff! I didn't review the frontend code at any depth, but it seems to follow the same patterns we already have, and I trust you will keep it from falling apart! :)

Worth noting from the review session:
- The long term plan is to replace notifications altogether with filters (e.g. show me suggestions containing packages I maintain)
- The frontend carries a lot of extra code to maintain cache consistency due to the optimistic updates. It is the way it is, having two interacting state machines.
  - We deem the additional complexity worth the improvement of user experience

Issue #1220, opened 2026-07-22
Rate limits


fricklerhandwerk:
Since we're exposing an API and grow the amount of interesting data, we need some rate limiting.

## Acceptance criteria

- anonymous users get less
- logged in users get more
- security team gets even more
- API-only requests (via token) are independent of requests via the UI (we can discriminate them using the authentication method)
  
  Rationale: If a security team member is overusing the API for custom scripting, they should still be able to work on regular triaging. It doesn't prevent "cheating" by using the API login for UI requests, but is meant to avoid mistakes such as using up your UI rate by a greedy script.

## Implementation notes

Since we don't have a lot of load at the moment and no good metrics (#1085 still in the pipeline and more follow-up work needed), we should have generous limits and narrow them down when we run into bottlenecks.

Django DRF has throttling built in: https://www.django-rest-framework.org/api-guide/throttling/

Use our beloved Pydantic settings for declaring the limits.

yash-dot-com:
hey @fricklerhandwerk 👋 
is this task still open for external contributions given that @florentc is currently assigned ?

florentc:
I'm ok with you giving it a shot.

yash-dot-com:
Thanks @florentc, I would really love to contribute & fix this issue! This is my first time contributing to open source, so I appreciate the opportunity. 😃 

Before I start drafting the PR, I had a few quick questions to align on expectations:

1. **Default Rate Limits**: Do you have specific default numbers in mind for each tier, or should I pick reasonable, generous defaults in the Pydantic settings for now?

2. **Security Team Identification**: What is the standard way to identify security team members in the codebase currently ?

3. **Auth Separation**: How are API tokens authenticated vs. UI sessions in the backend so I can separate their limits cleanly?

Once confirmed, I'll get working

devnchill:
hello @yash-dot-com 
> What is the standard way to identify security team members in the codebase currently ?

there is already a utility function [is_admin](https://github.com/NixOS/nix-security-tracker/blob/main/src/shared/auth/utils.py#L10)

> How are API tokens authenticated vs. UI sessions in the backend so I can separate their limits cleanly?

`knox` handles token based auth and `session` is from drf built in authenticator, you can look into `reqeust.auth` to differentiate between them

yash-dot-com:
@devnchill hey thanks, that clears things up perfectly. will draft a plan & share it here, get it verified before implementing & opening a PR.

devnchill:
[...]

PR #1312, opened 2026-09-10
feat: API rate limiting

florentc:
Implements #1220

This introduces throttling for API requests:

30/min for anonymous users
120/min for authenticated users
2000/min for security team (they are our trusted users and will likely need the API to script bulk operations)

There are 2 different cache keys for standalone API usage (authentication through bearer token) and UI API usage (cookie) as mentioned in the issue. Both have the same limits (so for authenticated users 120 requests per minute on the UI + 120 requests per minutes through using their API key)  but we could introduce different limits.

I have no idea if these values are realistic. I guess you'll have a better gut feeling @Erethon @fricklerhandwerk 

For the new UI, I'd say 30/min is the comfort zone, a.k.a. 1 click every 2 second. The UI is normally well optimized and a click usually generates 1 request only, at most 2. However I've see a bug occasionnally on staging (that I can't reproduce accurately) where the user info and server info are fetched at every click (so it'd be 3 requests per click). Even considering that, 30/min for anonymous users seems decent, especially considering query caching (a given info is considered fresh for 30 seconds see `staleTime` in the frontend).

### DIFF
--- a/src/project/settings.py
+++ b/src/project/settings.py
@@ -258,6 +258,9 @@ class Settings(BaseSettings):
             """,
             default=100 * 64 // 2,
         )
+        API_THROTTLE_ANONYMOUS: str = "30/min"
+        API_THROTTLE_AUTHENTICATED: str = "120/min"
+        API_THROTTLE_SECURITY_TEAM: str = "2000/min"
 
         @model_validator(mode="after")
         def default_server_email(self) -> Self:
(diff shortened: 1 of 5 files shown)


src/project/settings.py
```diff
@@ -258,6 +258,9 @@ class DjangoSettings(BaseModel):
             """,
             default=100 * 64 // 2,
         )
+        API_THROTTLE_ANONYMOUS: str = "30/min"
```
fricklerhandwerk, line 261: Side note: we may want to expose this as a configuration option if we want different deployments to have different values, but we can do that when concrete need arises.

fricklerhandwerk (commented):
Great!

fricklerhandwerk (approved):

PR #1320, opened 2026-09-14
feat: Access suggestions by CVE id in the new UI like in legacy UI

### LINKED ISSUE #1318: Access suggestions by CVE in new UI

florentc:
Closes #1318 

This mirrors what happens in legacy UI to the new UI. `by-cve` suggestion urls redirect to the canonical `by-id` url. Here, there is an optimization to avoid refetching the suggestion after the redirect.


fricklerhandwerk:
As we discussed previously, we will simplify this a lot in the future, where we'll have CVEs as first-class citizens. For now we're reimplementing existing behavior.

PR #1321, opened 2026-09-14
fix: Make development VM work with latest nixpkgs

Erethon:
Reasoning for this is in https://github.com/NixOS/nixpkgs/pull/552774#issuecomment-5670305160. Better to follow that thread for up to date information, as this PR might not be needed soon if that setting becomes the default.


fricklerhandwerk (approved):
That fixes it indeed. Thank you @Erethon!

PR #1328, opened 2026-09-15
fix: bring back dependencies to the devshell

fricklerhandwerk:
This allows editors to find symbols.

@florentc please test this locally, I didn't take the time to set up a JS LSP for myself yet.


florentc (commented):
Works like a charm for the python/Django code :+1:

No change unfortunately on the frontend side with `ts_ls`
Still getting `Cannot find module '@foo/bar' or its corresponding type declarations` on library imports in the typescript code.
Also getting `This JSX tag requires the module path 'preact/jsx-runtime' to exist, but none could be found. Make sure you have types for the appropriate package installed.` on JSX tags.

florentc (approved):
It worked after all but I needed to remove the pre-existing `node_modules` directory or the symlink would be inside it rather than being it

PR #1330, opened 2026-09-15
fix: Make sure dev VM ports don't conflict with ports for outgoing connections

Erethon:
The [default on Linux is for ports 32768-60999](https://github.com/torvalds/linux/blob/9b87fdc9af2fbfcdb5c24a64139685ef80f6573f/Documentation/networking/ip-sysctl.rst?plain=1#L1598) to be used as a client for outgoing connections. The old port-offset of 50000 allowed for conflicts to happen which could prevent the dev VM from starting. Setting this to 20000 ensures that the largest port we use (Prometheus 9090) falls outside this range (20000 + 9090 = 29090 < 32768).


fricklerhandwerk (approved):
Thanks! Please update the docs where `5\d{4}` appears.

Erethon:
> Thanks! Please update the docs where `5\d{4}` appears.

Done, I only found one such mention in the docs.

fricklerhandwerk (approved):

PR #1333, opened 2026-09-17
fix: node_modules overlay

fricklerhandwerk:
We need the JavaScript dependencies in the shell environment for editors, and also in the development VM for the system.
Layering a new directory in the bind mount within the VM requires a directory as a target.
Previously it had been a symlink, and that tripped up the upward resolution in search of a `node_modules` directory.


florentc (approved):
Works great 👍

PR #1337, opened 2026-09-18
feat: Autohide suggestion view mode selector when issues are collapsed

florentc:
By default on issue lists, the issues are collapsed and you can't see the suggestions inside. Yet, the suggestion view mode selector is visible. I found myself clicking on it rather than the issue view selector several times.

This PR auto hides the suggestion view mode selector.


fricklerhandwerk (approved):
Yeah I had the exact same experience, good catch of that paper cut.

PR #1340, opened 2026-09-21
feat: Package feed link (when package exists) in the package search UI

florentc:
This shows a link to the package feed in the package search when the package exists, even if there are no search results now.

This relies on a new API endpoint to check a package exists (which is independent from yielding empty or non-empty suggestion search results).

This new endpoint is the first package endpoint of our API. Eventually we want a package oriented API + UI.

For now, package still means attribute name. This will eventually have to be changed when we wire the API and UI to use the package data model.


src/api/packages/views.py
```diff
@@ -0,0 +1,30 @@
...
+
+    permission_classes = [AllowAny]
+
+    @extend_schema(
+        operation_id="getPackageExists",
+        description="Check whether a package with the given attribute name currently exists in Nixpkgs.",
+        responses={200: PackageExistsSerializer},
+    )
+    def get(self, request: Request, package_name: str) -> Response:
+        exists = NixDerivation.objects.filter(attribute=package_name).exists()
```
fricklerhandwerk, line 28: Not sure, we could also query `PackageAttribute` directly. It's a slightly smaller table (~100k vs ~400k) and gives us access to the other interesting things too when we want them.

fricklerhandwerk (commented):
Feedback on whether the package exists is good! Could be even more evident by making the background red or something.

 I'm not entirely convinced that folding the feed button into the search input field is taking a healthy direction though. Any detail about packages can be deferred to the dedicated view described in https://github.com/NixOS/nix-security-tracker/issues/1343.

florentc:
Of course the rightful place to get the feed link will be the package tab alongside the package views, but we don't have them yet. Showing the feed here for the time being is a way to expose the feature in a natural way. The workflow is efficient (you search, the feed icon appear, you click) and non invasive (the feeds are one of the more niche feature, users who look for them will immediately notice the icon, other users won't be distracted).

src/api/packages/views.py
florentc: The current implementation of this endpoint is not expected to live very long as one of our short term goals is to make use of the package data model. When we switch, we will look up in the rightful package data model tables. I chose not to in order to be clear that this is still written as part of the old "packages seen as attribute names" paradigm.

Do you prefer a FIXME to explain that or directly querying the other table?

fricklerhandwerk: I don't have any particular opinion here tbh, just wanted to point it out to make sure your choice here is deliberate.

PR #1341, opened 2026-09-21
feat: Post a compacted (no detail) issue on GitHub when body is too long

### LINKED ISSUE #1319: Publishing too many CVEs in an issue blows GitHub message limits

florentc:
Closes #1319 

This removes the body of GitHub issues that are too long instead of failing when posting them. Instead, a message invites the readers to consult the sec tracker issue via the provided link.

A deduplicated list of all the maintainers is still included to ping the relevant people.

This list is removed as well in the almost impossible case it alone breaks over the size limit.


src/shared/github.py
```diff
@@ -74,26 +78,51 @@ def cvss_details(suggestion: CVEDerivationClusterProposal) -> str:
...
+
+    def maintainers(suggestion: CVEDerivationClusterProposal) -> str:
+        maintainers_list = maintainer_usernames(suggestion)
         if maintainers_list:
             maintainers_joined = ", ".join(mention(m) for m in maintainers_list)
             return f"Affected package maintainers: cc {maintainers_joined}\n"
         else:
             return ""
 
+    def all_maintainers_line() -> str:
```
fricklerhandwerk, line 108: We can probably list the deduplicated maintainers unconditionally. The point of batching CVEs is to notify maintainers of all of them simultaneously, so if associated maintainers diverge across suggestions that's more likely an error in user data entry.

src/shared/github.py
```diff
@@ -74,26 +78,51 @@ def cvss_details(suggestion: CVEDerivationClusterProposal) -> str:
...
+        deduplicated maintainers of every suggestion in the bundle.
+        Used only for the compact body.
+        """
+        seen: set[str] = set()
+        deduped: list[str] = []
+        for cs in cached_suggestions:
+            for username in maintainer_usernames(cs.proposal):
+                if username not in seen:
+                    seen.add(username)
+                    deduped.append(username)
```
fricklerhandwerk, line 120: Why not...
```suggestion
        deduped: set[str] = set()
        for cs in cached_suggestions:
            for username in maintainer_usernames(cs.proposal):
                deduped.add(username)
```

src/shared/github.py
```diff
@@ -189,14 +218,39 @@ def help_text() -> str:
...
-[Nixpkgs security tracker issue]({tracker_issue_uri})
+{tracker_link}
 
 {sections}
 {help_text()}"""
 
+    # GitHub issue bodies are limited.
+    # Above that, only keeps the tracker link and the deduplicated maintainer mentions.
+    if len(body) > GH_ISSUE_BODY_MAX_LENGTH:
+        size_note = (
```
fricklerhandwerk, line 233: Most of the characters would be saved by reducing the CVE list to something like

```
- <severity> [<CVE ID>](<tracker link>)
```

which would have the nice side effect that the description is still useful at a glance. And then we can also keep the list of affected packages, deduplicated just like the maintainers are (which could also be unconditional).

florentc:
This was more annoying that I thought it would be. I reworked it all.

- maintainers and packages are combined and deduplicated globally and displayed at the end, no longer per CVE
- CVE-xxxx-xxxx links to the sec tracker suggestion, and "NVD" in parenthesis to the external CVE nvd link on sec tracker UI
- There are several levels of compacity depending on how much needs to be stripped
- The length are computed less stupidly 

Level 0: Everything visible
Level 1: Compact CVE list
Level 2: Just packages and maintainers
Level 3: Just maintainers
Level 4: Nothing (well, the link to sec tracker which is always there)

fricklerhandwerk (approved):
That sure is a busy-bee way to do that! I'd have been fine with just skipping the CVE details, which would likely have done the trick for all practical cases. But now that code will survive nuclear war. Why not.

PR #1342, opened 2026-09-21
fix: Remove links to non-existing logos

Erethon:
I noticed this via the nginx logs as we're getting 404s for these two paths. I don't think nixpkgs even has a logo, so let's remove these two broken links.


florentc:
It has a logo (the rainbow one) but @fricklerhandwerk didn't want to include them as assets in our codebase but rather get them from source. The nixpkgs one has builtin blank space around that need to be cropped first so it's a bit annoying and hasn't been done yet. See #1295

Erethon:
> It has a logo (the rainbow one) but @fricklerhandwerk didn't want to include them as assets in our codebase but rather get them from source. The nixpkgs one has builtin blank space around that need to be cropped first so it's a bit annoying and hasn't been done yet. See #1295

Whoops, I never saw #1295, closing this then.

PR #1344, opened 2026-09-22
feat: General improvements of user experience in new UI

florentc:
This PR improves many aspects of the UI, internally code-wise and externally UI wise.

## Suggestions and issues cards

This PR introduces a standardized card look for both and optimize space usage. Especially the placement of permalinks and external links would jump from one view to the other, the titles had to be truncated, and the view toggles were taking a lot of space.

Now the border integrates the id of the item (CVE-xxx-xxxx for suggestions and NIXPKGS-xxxx-xxxx for issues) with the same icons as those used in the main tabs. The id serve as permalink on sec tracker. The idea is to reuse this for packages when we introduce the package cards and the packages tab. The external links (NVD for CVEs and GitHub for issues) follow in parenthesis.

When the cards are nested (e.g. suggestions inside issues) it's immediately clear what's inside what.

The view toggles are small and now remind which global view mode is set (the current inherited view mode is shown in subtle black). This should improve view toggles and avoid misclicking on the current view mode and nothing happens.

---------------

---------------

---------------

---------------

## Mobile navigation

Users have reported issues with very narrow screens on mobile. Mobile navigation is not intended to be first class citizen in the Nixpkgs security tracker with this PR. Rather, it solves the low hanging fruits and avoid breaking the UI for the rest, using horizontal scroll as a fallback (which is a common pattern on mobile). It makes it possible to consult and use the tracker in decent conditions without glitches.

### Misc

There is a myriad of other such fixes shipped here, in terms of wrapping titles, activity log title, etc.


fricklerhandwerk (approved):
Great stuff!

